PR #34, opened 2022-08-17
Cleaning executed cell outputs

bsipocz:
A few of the cells have been executed, this PR cleans them up to have no output in the repo. (If we decide later to store executed notebooks we should make sure that the whole notebook has run not only a few cells).

I'm adding this as a separate PR, so it's not mixed with the actual content change of the coming PRs.


troyraen (approved):
LGTM

Issue #31, opened 2022-08-10
Necessary data files not in data folder


j-sarmiento:
Some files that the multiband_photometry.ipynb notebook needs are not in the data folder. For example, none of the 'infiles' files are in the data folder.

bsipocz:
This is something I'm working on ATM.

bsipocz:
This has been fixed, and the current version of the notebook should be able to run and acquire all the data necessary.

PR #35, opened 2022-08-17
Adding temporary requirements file

### LINKED ISSUE #32: Add environment file for package installation
Add an environment file that can be pulled up and installed in one step from the notebook than necessary, rather than do individual `try/except` and `pip install` lines in the import cell. Long-term packages should be added to the image and we should aim to minimize installing from the notebooks, but as an immediate workaround, we can indeed install things manually.

bsipocz:
To be used for notebook level installs until all those dependencies are added to the deployed container.

This PR also adds a bit of a cleanup to the import cell to remove unused imports, etc.

Builds on top of #34

To close #32


bsipocz:
This preference for installing stuff needs to be added to the internal confluence page about SPs.

forced_photometry/multiband_photometry.ipynb
```diff
@@ -132,44 +92,20 @@
     "\n",
     "### codes that may or may not already be installed in the platform\n",
     "try:\n",
```
troyraen, line 94: This is an empty `try/except` that prevents the cell from executing. Looks like it can be deleted.

bsipocz: thanks for spotting it, I was messing a lot with manual line commits in text mode and didn't spot this one. 

I may also rework the tractor imports, so to make most of the notebook runnable locally without a tractor install (which would e.g. help me a lot in hunting down the bug Marziye identified with the non-centralized cutouts)

bsipocz: I've ended not pushing the missing tractor try/excepts, but just keep them around locally. So this should be ready to go now.

troyraen (approved):
LGTM as far as I can test it. I ran into missing data (#31) and I get a `ConnectionError` when trying to create a firefly client. I stopped testing after that. Both errors also occur with the notebook in the main branch, so I don't think this PR needs to be held up for them.

PR #37, opened 2022-08-20
Adding IRSA S3 download to notebook

bsipocz:
The functionality requires https://github.com/fornax-navo/fornax-cloud-access-API/pull/4 to be merged, therefore temporary instructions for cloning the branch from the fork are included.

S3 data download works for me locally using my IRSA AWS token, and also works on daskhub. However there are some issues with falling back on the on-prem download for all other cases, it may be either a user bug in the notebook or a bug in the fornax API code.

@troyraen - Unfortunately atm we have download functionality, but not open, so kept the logic of dumping the data in the `../data` path.

### DIFF
--- a/forced_photometry/multiband_photometry.ipynb
+++ b/forced_photometry/multiband_photometry.ipynb
@@ -167,6 +167,115 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


troyraen (approved):
I'll need to obtain a token to properly test this, but otherwise LGTM.

PR #39, opened 2022-08-24
Minor workarounds for local development

bsipocz:
Temporary commit to (that will need to be rolled back when this gets deployed) that helps local development of the data access and cutout part without forcing a tractor installation.

(This is my attempt to cleanup the git stash and going back and forth with local temporary codes)

PR #41, opened 2022-08-25
MAINT: Cleanup galex data cells and naming of files

bsipocz:
This PR adds the final fixes so all cells work to the point of opening the data files. With this and the previous PR (#39) I'm passing the notebook now on to @troyraen to do code restructuring/refactoring as he sees fit.


bsipocz:
rebased to remove conflict with #43

troyraen:
Just tested this on dask. It runs for me through point of opening the data files, as expected.

bsipocz:
Thanks for checking it!

Issue #28, opened 2022-08-10
Make functions more modular


bsipocz:
Currently, I'm trying to reproduce some cutout bugs, but running into the issue that the cutout production is merged together with the tractor photometry in `calc_instrflux`. It can then live in the `code` directory, too.

I would suggest refactoring the tractor running function to be self-contained, e.g. it receives a `subimige` as a parameter.

PR #48, opened 2022-09-01
Extract background cutouts from `bkg_hdus` instead of `hdulists`

troyraen:
@bsipocz @jkrick (not sure who wants to be tagged here)

This PR includes a single change to fix what appears to be a bug in the cutout creation. The bug is that the calls to create the objects `subimage` and `bgsubimage` were identical. I'm guessing that `bgsubimage` is supposed be created from the sky background files in `bkg_hdus` -- this is the change made in this PR.

If this is not actually a bug, or is not the right fix, I will close this.

### DIFF
--- a/forced_photometry/multiband_photometry.ipynb
+++ b/forced_photometry/multiband_photometry.ipynb
@@ -640,7 +640,7 @@
     "    subimage, nodata_param, x1, y1, subimage_wcs = extract_cutout(ra, dec,cutout_width, mosaic_pix_scale, hdulists[g_band], wcs_infos[g_band])\n",
     "    #for the Galex images, also need to make a background cutout image\n",
     "    if background_method == 'skybg':\n",
-    "        bgsubimage, bgnodata_param, bgx1, bgy1, bgimage_wcs = extract_cutout(ra, dec,cutout_width, mosaic_pix_scale, hdulists[g_band], wcs_infos[g_band])\n",
+    "        bgsubimage, bgnodata_param, bgx1, bgy1, bgimage_wcs = extract_cutout(ra, dec,cutout_width, mosaic_pix_scale, bkg_hdus[g_band], wcs_infos[g_band])\n",
     "    \n",
     "    \n",
     "    #catch errors in making the cutouts\n",
--- a/forced_photometry/multiband_photometry.ipynb
+++ b/forced_photometry/multiband_photometry.ipynb
@@ -539,7 +539,8 @@
     "bkg_hdus = [fits.open(skybgfile)[0] for skybgfile in skybgfiles]\n",
     "\n",
     "#grab the WCS of the mosaics\n",
-    "wcs_infos = [WCS(hdulist) for hdulist in hdulists]"
+    "wcs_infos = [WCS(hdulist) for hdulist in hdulists]\n",
+    "bkg_wcs_infos = [WCS(bkg_hdu) for bkg_hdu in bkg_hdus]"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


troyraen:
I also don't understand why `g_band` is being used as the index for `hdulists` rather than `band`? Maybe that's a better question to be asked/answered offline.

forced_photometry/multiband_photometry.ipynb
```diff
@@ -640,7 +640,7 @@
     "    subimage, nodata_param, x1, y1, subimage_wcs = extract_cutout(ra, dec,cutout_width, mosaic_pix_scale, hdulists[g_band], wcs_infos[g_band])\n",
     "    #for the Galex images, also need to make a background cutout image\n",
     "    if background_method == 'skybg':\n",
-    "        bgsubimage, bgnodata_param, bgx1, bgy1, bgimage_wcs = extract_cutout(ra, dec,cutout_width, mosaic_pix_scale, hdulists[g_band], wcs_infos[g_band])\n",
+    "        bgsubimage, bgnodata_param, bgx1, bgy1, bgimage_wcs = extract_cutout(ra, dec,cutout_width, mosaic_pix_scale, bkg_hdus[g_band], wcs_infos[g_band])\n",
```
bsipocz, line 643: good job spotting this, maybe this is the bug we've been hunting for Marziye?

However, it would be great if you could also add a fix as part of this PR for making the function properly standalone rather than using local variables as if they were global ones. (e.g. `hdulist`/`bkg_hdus`/`wcs_infos` are all defined in a cell above, while they should really be provided as a kwarg to the function.)

troyraen: > add a fix as part of this PR for making the function properly standalone rather than using local variables as if they were global ones

I'm including that in a separate PR that addresses #28, which I hope to open tonight. I wanted to check whether this was actually a bug without mixing in that kind of refactoring. There are actually quite a few global variables being used in this function and I'd like to do something a little cleaner than just blindly passing everything as separate kwargs.

bsipocz: sounds great. Minor note, I would not suggest a blind passing of `**kwargs`, but to spell out all of the properly, as keyword arguments.

bsipocz: One other possible bug could be that this uses the WCS info from the header of the science images, blindly assuming that the WCS is the same for the background files, too. I would suggest to double-check whether this is true before merging this PR. (I suppose it can be a true assumption if the background images are clean-cut derivatives of the science images, but that may not be true for all cases. 
So to be sure which one is the case, I would suggest waiting for @jkrick's input here as she is the one who knows the Spitzer images and workflows.

jkrick: yes, thanks for catching this. I believe it is a bug.  I would want the background cutouts to come from the background mosaic.  

relevant to Brigitta's comment above, I would also suggest we pull in the wcs from the background images if possible, because that makes the code more useable for other datasets that might not have matching science and background images.

jkrick: ok, thinking for another full minute here, this is only relevant when the background image is different than the science image, which is not the case for the IRAC bands, but is the case for Galex.  Just wanted to be clear that this bug will only effect Galex frames.

troyraen: Thanks for the input. Should be no problem to pull in the WCS for background images. I'll make the changes.

troyraen: @bsipocz I've made these changes. Can I get an approval?

I was going to just put these changes in #49 (since the way the data gets loaded and passed around will be quite different) but in order for my test of that PR to work I need to get these bug fixes into `main`. So here there they are.

bsipocz (approved):

bsipocz:
Thanks @troyraen!

I've also added a `ready-for-review` label, that may help us track what PRs are still under work and which ones are ready for their final checks and merge.

PR #49, opened 2022-09-01
Make `calc_instrflux` function more modular

troyraen:
Completes #28 

### Overview: 

Refactor the function `calc_instrflux` so that it is more modular and does not rely on global variables.

### Testing:

See [test-pr49/results.ipynb](https://github.com/troyraen/fornax-demo-notebooks/tree/issue28/test-pr49/results.ipynb). 
There is significant refactoring in this PR including shuffling the way many parameters are defined, zipped together for input, and then handled on the other end. The tests demonstrate that all the parameters get zipped together and input properly, and that the function's results are unchanged.

### Highlights:

- Global variables: Most of the function's required variables that are currently global are associated with a specific passband. Create a simple class called `Band` which is a container to hold the the parameters for one band. Refactor the code that collects these parameters.

- Modular: Separate the function body into smaller functions and move them into a new module called `photometry`. Leave the main function `calc_instrflux` in the notebook so the user can see the steps/calls.

- Error handling: Remove the `nodata_param` and `fit_fail` variables that were being used to track the success/failure of the cutout extraction and tractor respectively. Instead, raise a custom error from each function (`CutoutError` or `TractorError`) at the spot that it fails. Then catch these errors in the main function and return `NaN`s for the flux and uncertainty. Note: I think we can probably just let the `CutoutError` raise instead of catching it. I don't think it's even being triggered (I have a test in progress right now). If it does get triggered in the future it probably means we actually did something wrong, so it would be better to let it raise.

- Update the `extract_cutout` module:

    - Rename the _module_: `extract_cutout` -> `cutout`.
    - Rename the _function_: `extract_cutout` -> `extract`. Update the input arguments and related internal code. (Name of the input file argument changed, and WCS is no longer an argument)
    - Add a wrapper called `extract_pair` that returns cutouts from a tuple of HDUs containing source and background images.

- Remove the parameter `background_method` that was being used to check which file to calculate the sky-background from. Instead, bind pairs of images (science and background) together in a tuple when the input files are first defined. If the science image should double as the background image, input it only once (so the tuple "pair" only has one element). In that case, have `cutout.extract_pair` do the extraction once and return it for both the science and background cutouts.

- ~Lazy-load the FITS files~. There are a small number of files and they get accessed many times each. Load the HDUs once, up front (as before), but leave all data access for things like WCS to the function that actually needs it. I'm guessing the I/O is significantly more expensive than access (we can time-test the data access if we still want to reduce runtime). Loading up front matches the cloud access pattern better as well.


code/photometry.py
```diff
@@ -0,0 +1,171 @@
...
+                warnings.filterwarnings("ignore", ".*divide by zero.*")
+                # warnings.simplefilter('ignore')
+                for tr in range(20):
+                    dlnp, X, alpha, flux_var = tractor.optimize(variance=True)
+                    # print('dlnp',dlnp)
+                    if dlnp < 1e-3:
+                        break
+
+    # catch exceptions and bad fits
+    except Exception:
```
bsipocz, line 115: A good rule of thumb is to be as precise as possible with excepting exceptions, e.g. only let the ones through that are in fact expected rather than the most generic `Exception`.

code/photometry.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+
+    # if cutout extraction failed, raise an error
+    if bgnodata_param is True:
+        raise CutoutError("Cutout could not be extracted")
+
+    return subimage, x1, y1, subimage_wcs, bgsubimage
+
+
+def run_tractor(subimage, prf, objsrc, skymean, skynoise):
```
bsipocz, line 80: These functions could definitely use a docstring

code/photometry.py
```diff
@@ -0,0 +1,171 @@
...
+    # suppress warnings about nans in the calculation
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=RuntimeWarning)
+        skymean, skymedian, skynoise = sigma_clipped_stats(
+            bgsubimage, sigma=3.0)
+
+    return subimage, objsrc, nconfsrcs, skymean, skynoise
+
+
+def make_cutouts(ra, dec, infiles, cutout_width, mosaic_pix_scale):
```
bsipocz, line 44: I feel it would be beneficial to enforce the arguments for all these functions in `code` to be keyword only, or at least the ones after ra/dec; overall that would make the notebook code more explanatory and less error-prone.

bsipocz (commented):
Some minor comments (I didn't run the notebook either), overall I very much like the direction this is going.

code/photometry.py
troyraen: Input from someone more familiar with tractor would be helpful. This is a minor improvement over the bare `except` that was here previously, and it's not clear to me what else to expect. I see the [tractor docs](http://thetractor.org/doc/index.html), but no specific documentation on `Tractor.optimize`. 

However, I did put in an `except Exception` in a different place that I know I can be more specific about, so I'll fix that.

bsipocz: Good point, I think we should then migrate this how-to to the confluence docs, something along the line:

- By default, avoid try/except, especially with generic `Exception`, add them only for well-understood and specific exceptions only (e.g. the fitter may raise a specific exception for non conversion, then catch that one only and handle the case, but don't catch everything else as that may make bugs more difficult to discover). 

So far, I think this is the first item that can be best addressed from the beginning, during the initial exploratory phase of the notebook/science use case. 

cc @jkrick

forced_photometry/multiband_photometry.ipynb
```diff
@@ -503,19 +458,75 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# prfs and other configs for each channel\n",
     "\n",
+    "# container for configs of a single band\n",
+    "class BandConfigs(NamedTuple):\n",
```
troyraen, line 473: I'm not sure that `BandConfigs` is the right name for this class (and the related `band_configs` variables below) -- specifically the "configs" part. This is something the user may interact with, so I'd like it to be well-named such that its meaning/contents is clear to the user. Seeking input @jkrick .

code/photometry.py
troyraen: I fixed this in all the files touched by this PR, and added the rest as a to-do in #51.

forced_photometry/multiband_photometry.ipynb
troyraen: Documenting a synchronous conversation: decided `BandParams` would be better.

troyraen:
@bsipocz I wasn't sure where to put the test code and results. For now they are just here, in a root-level directory. Is there a proper place for them? 

If not, I figured I would keep them in my own fork, but drop them from the PR before merging to `fornax-navo: main`.

code/photometry.py
```diff
@@ -0,0 +1,168 @@
...
+    skymean: float
+        Mean of sigma-clipped background
+    skynoise: float
+        Standard deviation of sigma-clipped background
+    """
+    # suppress warnings about nans in the calculation
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=RuntimeWarning)
+        skymean, skymedian, skynoise = sigma_clipped_stats(
+            bkgsubimage, sigma=3.0)
```
bsipocz, line 40: Nitpick: I think we can go with longer lines to avoid these line breaks, would e.g. 100 work for you?

code/photometry.py
```diff
@@ -0,0 +1,168 @@
...
+
+    return skymean, skynoise
+
+
+def run_tractor(*, subimage, prf, objsrc, skymean, skynoise):
+    """Make the tractor image and perform forced photometry.
+
+    Parameters:
+    -----------
+    subimage: <need type>
```
bsipocz, line 50: I suppose this expects an array at this point?

code/photometry.py
```diff
@@ -0,0 +1,168 @@
...
+                # warnings.simplefilter('ignore')
+                for tr in range(20):
+                    dlnp, X, alpha, flux_var = tractor.optimize(variance=True)
+                    # print('dlnp',dlnp)
+                    if dlnp < 1e-3:
+                        break
+
+    # catch exceptions and bad fits
+    except Exception as e:
+        raise TractorError("Tractor failed to converge") from e
```
bsipocz, line 105: again, how meaningful the tractor traceback, do we need to have this custom one here and the try/except?

code/photometry.py
```diff
@@ -0,0 +1,168 @@
...
+                warnings.filterwarnings("ignore", ".*divide by zero.*")
+                # warnings.simplefilter('ignore')
+                for tr in range(20):
+                    dlnp, X, alpha, flux_var = tractor.optimize(variance=True)
+                    # print('dlnp',dlnp)
+                    if dlnp < 1e-3:
+                        break
+
+    # catch exceptions and bad fits
+    except Exception as e:
```
bsipocz, line 104: I know that we can't necessarily change this now, but rule of thumb: only except specific exceptions rather than the very generic one. This current approach will catch all bugs and expected issues under one umbrella and calls it convergence error, while it may not be always the case.

bsipocz (commented):
Some superficial review of the code, without trying to run it.  I'll come back to the notebook and test stuff later.

bsipocz:
re test outputs:

 - I think it's OK to keep tests, and test running things in the repo. But not in the form of notebooks. So some removal will be needed in this PR before it goes in (including squashing them out from intermediate commits, too).
 
 - We can, and probably should also set up tests for the main notebook, if that is of any help, but that should probably be done once it's overall runtime cut back.

code/cutout.py
```diff
@@ -0,0 +1,115 @@
+# function to extract cutout image
+import math
+from astropy.nddata import Cutout2D
+import astropy.io.fits as fits
+from astropy.wcs import WCS
+
+from exceptions import CutoutError
+
+
+def extract_pair(ra, dec, *, img_pair, cutout_width, mosaic_pix_scale):
```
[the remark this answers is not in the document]
troyraen: We don't necessarily need the extra function. There is a bit of logic in there that figures out whether to pull the background cutout from the science image or a separate file, so I put it in a function thinking it would be easier to use that way. But the logic ended up being really simple, so in the end it's probably fine to go in the notebook if folks think that's better.

code/cutout.py
```diff
@@ -0,0 +1,115 @@
...
+    """Extract cutouts from the science and background images in `img_pair`.
+
+    If `img_pair` contains only one element, extract the cutout once and return it for both.
+
+    Parameters:
+    -----------
+    ra, dec: float or double
```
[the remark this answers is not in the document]
troyraen: I agree, `SkyCoords` and one arg rather than two would be great. I actually thought about doing that, but many functions would need both their signatures and internal code updated to handle it, so I didn't touch it here. Happy to make that change, but would prefer to do it in a separate PR.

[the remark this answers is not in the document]
troyraen: Ah, thanks

code/cutout.py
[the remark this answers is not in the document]
bsipocz: It's nitpicking for this use case (we can keep both), but in general I'm usually on the side of a simpler API. E.g having one function that two very similar ones. Here, in these use cases I think we can force the user to provide the "pairs", or to have the bkg input optional, rather than to provide both `extract` and `extract_pair`.

code/cutout.py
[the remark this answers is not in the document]
bsipocz: OK, a separate PR sounds good. I feel on the user level of the notebooks it's actually nice to force them to work with SkyCoords. (and one other option we do often in astroquery, but I don't necessarily say it's a good approach for this case, is to allow things that can a SkyCoord constructor consume. Here I would say that keeping it simple, and strict for skycoords is good)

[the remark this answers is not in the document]
troyraen: Thanks, is there some link you can send me to a style guide? For example, I was also wondering if these should these be wrapped in single backticks like:

```
`ra` : `float`
```

I've learned a couple of styles and still having trouble remembering one standard consistently. I had recently started referring to [LSST's guide](https://developer.lsst.io/v/DM-15183/python/numpydoc.html), but that is Numpydoc and I think we want something different here (but maybe I'm confused about that too).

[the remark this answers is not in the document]
bsipocz: I'm not familiar with the LSST one, but numpydoc is a usual one. (I use numpy as a reference myself, with the exception that in astropy land the single backticks are causing test failures if cannot be resolved to a sphinx reference, and we routinely double backtick those. And I recently discovered that that's kind of against the specification). 

Anyway, parameter name in the list doesn't do into backticks, sphinx boldfacing it by default. Single backticks, by default, are tried to be converted into sphinx link, assuming the intersphinx is set up properly. Double backticks the serve as monospace hightlights.

code/cutout.py
```diff
@@ -0,0 +1,115 @@
...
+        cutout = Cutout2D(hdu.data, position, size,
+                          copy=True, mode="trim", wcs=wcs_info)
+        subimage = cutout.data.copy()
+        subimage_wcs = cutout.wcs.copy()
+
+        # now need to set the values of x1, y1 at the location of the target *in the cutout*
+        x1, y1 = subimage_wcs.all_world2pix(ra, dec, 1)
+        #print('x1, y1', x1, y1)
+
+    except Exception as e:
```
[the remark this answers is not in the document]
troyraen: Short answer is: I think you're right, and I'm happy to do a separate PR that takes out this try-except altogether (plus the custom error).

I did it this way for now because I didn't want to change the guts of the try/except logic that was already here, since I'm already changing so many other things in this PR. I wanted to be able to test my changes against the input/output from the original notebook without having to sort out changes in whether a particular input raises an error.

So instead I opted to raise a custom error in the existing `except` blocks, just to make it easier to isolate those try-excepts and deal with changing them later. This way it's simple for everything else to decide whether to catch or let raise a `CutoutError` or `TractorError` without worrying about the details of those functions -- and we can separately work on those functions and narrow down the set of errors that triggers the custom errors. This is probably much more relevant for the tractor error than the cutout error.

For this cutout function, I think you're right that we probably don't need the try/except at all. I suspect that the `try` block never actually fails. (And I ran a preliminary test last night that confirmed this.) And if it does fail, that's probably an actual problem with our code, so better to just let the error raise.

code/photometry.py
troyraen: Yup that works. I'll change my IDE settings right now.

code/photometry.py
troyraen: Ya that looks right (`np.ndarray`), thanks!

code/photometry.py
troyraen: The custom errors were just a simple way to isolate these try-except blocks so that I could refactor and test the rest of the code more easily. They are mostly about making the development easier, and can certainly be taken out when we fix the try-except logic (but fixing that logic was outside the scope of this PR). I explained more of my thought process in my comment above about the `CutoutError`. That said, I'm certainly interested to know if you would prefer a different strategy here (short of fixing the try-except logic itself)? 

Bigger picture:
I suspect that tractor really is failing sometimes (unlike the cutouts), and that it's fine to just skip that source when it does fail. So I think it probably makes sense to have a try-except block here so that it doesn't crash the program entirely. IMO, the custom `TractorError` is useful because it lets the main function catch a specific error without being tied up in the details of what the error is -- so everything else can move on while we try to sort out those details. Once we do sort it out we can replace the custom error with whatever tractor is actually throwing.

code/photometry.py
troyraen: I understand and agree. To be clear, these try-except blocks are not introduced by this PR, they were already here. The only change I made to them directly was to replace the bare `except`s with `except Exception`. Otherwise, I tried to patch them up a bit around the edges and isolate them from the rest of the code, but I intentionally did not change how they fundamentally work because that would have interfered with my ability to test the changes that were the real focus of this PR.

troyraen:
re test outputs: 
Sounds good, and I'd love to have a more in-depth conversation about this. I really struggled trying to figure out how to set up and run my tests, given the notebook format/environment. You probably have some better ideas about how to do it.

code/cutout.py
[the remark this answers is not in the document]
troyraen: I like that approach. Both because it makes the code simpler and because it forces the user to be clear about what they're submitting (like units) so there's no room for confusion.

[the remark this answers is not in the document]
troyraen: Ah thank you that's very helpful. The single/double backtick thing has caused me time-sinks more than once 😖 .

code/photometry.py
bsipocz: > Bigger picture:

OK, I like your bigger picture explanation and don't mind to have the custom one if the user cannot really do anything to fix those sources. But I still do mind the generic exception, presumably tractor raises specific types, and only those (maybe even matching for the error messages rather than letting all e.g. ValueError through), should be captured. 

But I don't think at this point it's worth spending time to run tractor again waiting for these errors, so it's OK to leave this as is and keep this rule of thumb of capturing specific, and understood things only, in mind for the new use cases.

troyraen:
@bsipocz I believe I've made all the necessary updates here. Let me know if I can merge or if I've missed something.

I refactored the tests a little bit and removed results.ipynb. The test results are now written to a plain text file instead. All of the code that collects the output and runs the tests is in tests/pr49/collect_and_test.py. The README in that directory describes the tests, gives instructions for running the tests, and points to a results file from tests I actually ran today.

bsipocz (approved):
One last thing for a follow-up: I think we should drop the reference for the PR, the repo and it's content is better if self explanatory without references to issues and PRs. And maybe also rename `output` to expected_output, or just simply data?

bsipocz:
OK, let's merge this as a lot has been done here and no point in keeping adding more to it. 

Thank you so much @troyraen!

PR #53, opened 2022-09-09
Fix index error in flux_unc tractor result

troyraen:
Fixes an index error in the assignment of tractor results to the dataframe column for flux uncertainty.

@jkrick @bsipocz


bsipocz (approved):

bsipocz:
Thanks!

Issue #57, opened 2022-09-16
Remove the try-except from `cutout.extract`


troyraen:
I don't think there should be a try-except here. If the cutout fails let that exception and traceback come through, no need to wrap it up into our custom one. (Or if there is one, could you elaborate why we need it?)

_Originally posted by @bsipocz in https://github.com/fornax-navo/fornax-demo-notebooks/pull/49#discussion_r971186497_

PR #59, opened 2022-09-21
Restructure tests dir

troyraen:
Restructures the `tests` directory. Specifically:

- Renames `tests/pr49` -> `tests/test_calc_instrflux`
- Within the above directory, renames `output/RESULTS.out` -> ~`RESULTS_2022-09-20.out`~ `logs/test_calc_instrflux_2022-09-20.out` and updates the code that generates this filename. This removes the `output` directory from the repo and adds a `logs` directory.

These changes are in response to the following request:

> One last thing for a follow-up: I think we should drop the reference for the PR, the repo and it's content is better if self explanatory without references to issues and PRs. And maybe also rename output to expected_output, or just simply data?

_Originally posted by @bsipocz in_ https://github.com/fornax-navo/fornax-demo-notebooks/pull/49#pullrequestreview-1114206318


bsipocz:
> Within the above directory, renames output/RESULTS.out -> RESULTS_2022-09-20.out

In addition to version control, I don't think that we need to track the dates for files. E.g. we can keep this file around as "ground truth" and compare new runs to it if we want to formalize testing.

troyraen:
This file does not contain a ground truth, it's just a record of what was printed to `stdout` during the test:

```
Comparing output located in: /home/jovyan/raen/issue28/tests/pr49/output
Generated from branches:
    troyraenissue28 - HEAD SHA: 2e622ba40d018f154b8fc948038d123ec56abf2d
    fornaxnavomain - HEAD SHA: 67a2e624830f5957834688ab45e7232534444216
Running equality tests...
PASSED : Band params -- branches produced identical output
PASSED : Source params -- branches produced identical output
PASSED : calc_instrflux output -- branches produced identical output
```

I did not commit the intermediate files created by the test script (ie "ground truth"). Mostly because one of them is 50 MB. Seemed like we needed something more like a shared storage dir on daskhub (or somewhere) to keep that data around for testing. Then the test code could point to it but we wouldn't have to keep it in the repo. What do you think about that?

I figured I _should_ at least include a record of the test that I actually did run, so that's what this file is. Seemed like it needed some sort of identifier, and a SHA seemed too opaque so I went with the date instead. I agree that it's a little weird. Maybe there should be a separate results directory, or maybe this file doesn't need to be here at all?

bsipocz:
Ok, but then this is more like a log, rather than results.

troyraen:
> Seemed like we needed something more like a shared storage dir on daskhub (or somewhere) to keep that data around for testing. 

The more I think about this the more I like the idea. I'm about to do #57, and writing the test for that would become SO much easier if I can just point the test at some input parameters stored on disk. In #49, most of the code I wrote for the test was just getting the input parameters out of the notebook. Having to get the data (both the function's input parameters and its expected output) out of the notebook also makes the test workflow really clunky, and makes it require a human in the loop.

So actually, I think it might work a lot better to write one script/module that runs a notebook and dumps a bunch of the parameters and function output into files on disk somewhere. Then going forward, we can write tests that only have to deal with the functions themselves. Most of the functions we need to test have been pulled out of the notebook already, so they're easy to for a test script to access. And if/when the input or expected output changes, we can just re-run the script that generates the files on disk. But having some "official" location of shared storage for this would let us disconnect the notebook from the testing a lot more, which would make it so much easier to write and run a single test.

troyraen:
> this is more like a log, rather than results

Ah! Thank you 😄 

I renamed the file to "test_calc_instrflux_2022-09-20.out" and moved it to a "logs" dir. Let me know what you think about that.

bsipocz (approved):

bsipocz:
Thanks @troyraen!

PR #60, opened 2022-09-27
Remove CutoutError

troyraen:
Completes #57 

Removes the try-except from `cutout.extract`, the related try-except in the notebook, and the custom `CutoutError`. Although no formal testing is included here, I did perform some tests and the `CutoutError` was never triggered. I tried several times to run **_all_** of the sources from the notebook's `paramlist` through `cutout.extract` but was never able to get it to complete, mostly due to my daskhub server shutting down after some period of time. The longest I was able to run the test was about two hours.

I think that the cutouts are working just fine, and that it's not worth holding this PR back any longer for better testing. My sense is that @bsipocz agrees, so I'm marking this as "ready for review". We can always revisit this if we encounter an error in the future.


bsipocz (approved):
This all looks great to me, thank you!

PR #62, opened 2022-09-30
DOC: Sort out automated fornax-api cloning & add Troy to authors

### LINKED ISSUE #45: Add Troy as an author in the summary cell.

### LINKED ISSUE #61: ModuleNotFound error
running on IRSA SP, I am getting a ModuleNotFound error for the `import fornax` statement in the section on getting IRAC data from the cloud

bsipocz:
This closes #45 and closes #61


bsipocz:
@jkrick - this now checks whether one has a clone of the API repo or not yet, and clones it if not. It's really just a very basic workaround that addresses the actual issue you run into, but it would not work when a newer version is required, or etc, but I don't think it's worth going into a more robust solution with e.g. git submodules, or similar, but we should rather work on getting it into the daskhub image (and even then, it should eventually land in pyvo proper).

jkrick:
I was able to run it, and it works for me on the IRSA SP.  ready to merge.

PR #67, opened 2022-10-11
clean run through entire code now works on SMCE!

### LINKED ISSUE #66: nway (X-ray catalog cross matching tool) is not actually being run on the SMCE
Instead the catalog generated by a previous run of nway is getting used.  
I am getting the error "/bin/bash: nway.py: command not found"

jkrick:
fixes #66


forced_photometry/multiband_photometry.ipynb
```diff
@@ -1131,7 +1099,7 @@
    "outputs": [],
    "source": [
     "#nway calling sequence\n",
-    "!nway.py '../data/Chandra/COSMOS_chandra.fits' :ERROR_RADIUS '../data/multiband_phot.fits' 0.1 --out=../data/Chandra/chandra_multiband.fits --radius 15 --prior-completeness 0.9"
+    "!/home/jovyan/.local/bin/nway.py '../data/Chandra/COSMOS_chandra.fits' :ERROR_RADIUS '../data/multiband_phot.fits' 0.1 --out=../data/Chandra/chandra_multiband.fits --radius 15 --prior-completeness 0.9"
```
bsipocz, line 1102: not a big deal, but it would be much nicer if the python software could be used as a library rather than a cmdline tool. I looked at it's repo, not obvious whether it could be done easily.

forced_photometry/multiband_photometry.ipynb
```diff
@@ -872,14 +872,32 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "df"
+    "#When doing a large run of a large area, save the dataframe with the forced photometry \n",
+    "# so we don't have to do the forced photometry every time\n",
+    "\n",
+    "df.to_pickle('../data/COSMOS_15arcmin.pkl')\n",
```
bsipocz, line 878: minor organizational comment: I think these result files should be in a subdirectory. Maybe in `../data/results/forced_photometry`, or even just in the same directory as the notebook? I don't the a right answer, just the feeling that we shouldn't mix the input and outputs (especially that the latter is here mostly as a workaround to produce a real demo notebook that doesn't have to run for a long time).

bsipocz (commented):
Two minor comments, in addition to the previous one to change a format that is not pickle, for good practice.

None of these has to be done in this PR through.

forced_photometry/multiband_photometry.ipynb
jkrick: I agree, it is a bit strange to make a command line call. I don't remember the details because it has been a long time since I worked with nway, but I think for now, since it does work, we should leave it as is.

PR #71, opened 2022-12-09
Update Gaia part

afaisst:
I added to the Gaia photometry part of this usecase:
- Added extraction of epoch photometry from Gaia using Datalink
- Added possibility to convert epoch photometry to MultiBandTimeSeries using Brigitta's code.


afaisst:
reverted the changes, because it didn't merge correctly.

PR #78, opened 2023-02-01
Gaia update, change of units

afaisst:
Changes:
- Added Gaia sub-case, including (1) matching to Gaia, (2) extracting mean photometry, and (3) extracting epoch photometry
- Converted the epoch photometry (and mean photometry if epoch photometry does not exist) into a Pandas MultiIndex object. Can be added to other tables by using "pd.concat()"
- Changed units from Jy to mJy


jkrick:
looks good, I didn't try running it, but I see the output plots.  The only thing I see is that you don't need np.repeats for the band or the object id.  A single value is ok for those, and the multi index data frame knows what to do.

afaisst:
@jkrick : i implemented your changes. In addition:
- created nice function to convert Gaia epoch photometry in Pandas MultiIndex data frame
- created MultiIndex class `MultiIndexDFObject()', which creates a Pandas MultiIndex DF. You can use *.data to display the data and *.append() to append another data frame (that way the try: etc test to see if DF exists can be hidden).

PR #82, opened 2023-02-15
Added IceCube matching

afaisst:
Changes:
- Added IceCube matching
- Added data directory "data/icecube/..." with IceCube ancillary data
- modified figure to plot IceCube Neutrino detections.

More information (also posted in Slack):

It's a bit hard to plot the results (or add them to the light curve object). Problems are:
- If there is a neutrino detection to an event, it would only be 1 (maybe two?) neutrinos, i.e., detections in icecube.
- The energy is in GeV, therefore not comparable to photons on the final plot
- What I did is to select the closest 3 neutrino events for each source. The positional uncertainty is gigantic (> 1 deg). The light curve object does not include the distance to the source coordinates, therefore it's hard to judge and to link the neutrino to a given source. But... maybe that's the next step for a scientist using these data.
- For plotting (see attached), I added an "elsif" so a point with downward arrow is plotted at the times of the neutrino event. The y-axis is arbitrary (set to 70% of the plot size for now). We might want to have a better visualization, for example change the point size with angular distance between neutrino and source or indicate that as a number. For that, however, we have to add another column to the df_lc where we can store the angular distance.


afaisst:
We decided to merge this branch, however, we will improve on the IceCube neutrino detection:
- create a function, for example getIceCubeNeutrinos(ra, dec, time), to have a more straight forward way to query neutrinos
- eventually determine the time of flares in UV/optical/IR data and then query also in time space for neutrino detections.
- further along, we could also think of identifying flares automatically in UV/optical/IR data

PR #86, opened 2023-03-07
Updates faisst feb23

afaisst:
Changes:
- out-sourced functions for HCV, Gaia, icecube, mast. Added imports to some of the out-sourced functions (because imports in main Jupyter Notebook are not global and cannot be seen by other imports: maybe something to fix later??)
- Modified figure (legend out of plot, separate subplot for ZTF zoom-in)
- Figure output format is now PDF (but can easily be changed back to PNG)
- Added option to load the saved pickle file containing the lightcurve object
- Added in data directory pickle file from @jkrick from February 23, 2023

To do:
- next step is to add a running mean smoothing on ZTF data to highlight variations. This can be added to zoom-in plot. Will do this during next iteration.
- Find a smarter way to modularize the notebook. Basically create class or function F(ra,dec) that automatically runs everything. Modularize so it is easy to add sub-modules in the sense of Mast(ra,dec) to add lightcurve data to main lightcurve object. This is more a long-term goal and needs a couple of days of work. But it is straightforward to do.


bsipocz:
Few big picture things:
 - Don't include the ``__pychace__`` directory and its content.
 - we need to stop using pickle files to share data if it's being used outside of a single session and beyond very internal code.

afaisst:
- Ah! I thought I put __pycache__ in .gitignore!
- Yes, the pickle file is meant for debugging only. Eventually should be removed.

afaisst:
added `__pycache__` to .gitignore

bsipocz:
> Yes, the pickle file is meant for debugging only. Eventually should be removed. 

We really need to find a way for shared data storage, as a repo is not really ideal for that. Even if files are removed later, they stay in the history, and there is a limit to how big a repo can be.

afaisst:
Removed ./light_curves/code/__pycache__/ directory.

.gitignore
```diff
@@ -15,6 +15,7 @@
 
 #pycache
 _pycache_
```
bsipocz, line 17: There is a typo here, should have double underscores rather than single, thus it didn't work as expected.

afaisst:
> > Yes, the pickle file is meant for debugging only. Eventually should be removed.
> 
> We really need to find a way for shared data storage, as a repo is not really ideal for that. Even if files are removed later, they stay in the history, and there is a limit to how big a repo can be.

We should definitely put the icecube data somewhere else. That's 390MB. The pickle file is <1MB currently... but can grow to infinity.

bsipocz:
> We should definitely put the icecube data somewhere else.

If there is a programmatic way to access it, e.g. it's reproducible how you created it, I would say it should be part of the notebook, or even better, be a separate, small notebook on its own.

.gitignore
afaisst: changed that

afaisst:
> > We should definitely put the icecube data somewhere else.
> 
> If there is a programmatic way to access it, e.g. it's reproducible how you created it, I would say it should be part of the notebook, or even better, be a separate, small notebook on its own.

It's a zip file downloaded on a webpage. We can probably use wget or something similar to get it.

jkrick:
looks good to merge.  IceCube re-work can happen in a later version

PR #88, opened 2023-03-10
ztf functioned, labels added

xoubish:
Changed ZTF to function, added labels as one index of data array in all lightcurves, added tqdm to most cells for progress and time monitoring. Changed plotting to take extra index into account.


jkrick:
Looks good.  My only comment is that we might want to rename the ZTF function to something slightly descriptive, so maybe 'get_ZTF_data' or 'get_ZTF_lightcurves' or ???

xoubish:
Good suggestion. Do I change it after the merge or before and request again?

jkrick:
you can change it now, and push the new commit which will automatically go into this PR.

PR #89, opened 2023-03-15
Add documentation

jkrick:
added doc strings to functions
cleaned up markdown documenting cells in the main notebook
added references to the codes used to the notebook


light_curves/code/fluxconversions.py
```diff
@@ -1,6 +1,24 @@
...
+    mag : array-like
+        array of WISE magnitudes
+    magerr : array-like
+        array of WISE uncertainties on the magnitudes
+    band : str {'w1', 'w2'}
+        name of the WISE band to be converted
+        
+    Returns
+    -------
+    flux: array
```
bsipocz, line 17: sphinx rendering nitpick, spaces are needed before and after the `:`, otherwise it won't boldface correctly. You did it right for the Parameters, as I see it only affects the Returns

bsipocz: OTOH, we don't do sphinx API docs generation, so this is not an issue whatsoever

light_curves/code/fluxconversions.py
```diff
@@ -19,7 +37,28 @@ def convert_WISEtoJanskies(mag, magerr, band):
     return flux_Jy*1E3, fluxerr_Jy*1E3  #now in mJy
 
 def convertACSmagtoflux(date, filterstring, mag, magerr):
+    """converts HST ACS magnitudes into flux units of Jansies
```
bsipocz, line 40:
```suggestion
    """converts HST ACS magnitudes into flux units of Janskies 
```

bsipocz (approved):
Looks great, thanks @jkrick!

PR #94, opened 2023-04-19
Make `code` modules safe for parallelization

troyraen:
Included in this PR:

1. Change all `*_get_lightcurves()` functions so they no longer accept `df_lc` as an argument. Instead, each function creates its own `MultiIndexDFObject`, appends to that, and returns it. The `MultiIndexDFObject`s from each archive call can then be collected and concatenated into a single object. This makes the `*_get_lightcurves()` functions more self contained and safer for parallelization because it prevents processes from trying (and failing) to share the `df_lc` object.
2. Change `code.sample_selection.remove_duplicate_coords()` so that it returns `list`s. (Plus minor changes to some `*_get_lightcurves()` functions to accommodate.) Something in this function creates a weak reference instead of a real object, and a weak reference cannot be sent into a background process (because it cannot be pickled). Casting to a `list` creates a real object, and seems to be what was originally intended so that the output type matches the input.
3. Add import statements to individual `code` modules so that they don't rely on global imports.

This PR does not make changes to the main notebook, multiband_lc.ipynb, but the following is an example of what's envisioned.

### 1. Setup (same as in multiband_lc.ipynb)

```python
import astropy.units as u
import pandas as pd
import sys
from math import ceil
from multiprocessing import Pool

sys.path.append("./unTimely_Catalog_explorer")
sys.path.append("code/")

from sample_selection import get_yang_sample, remove_duplicate_coords
from panstarrs import panstarrs_get_lightcurves
from gaia_functions import Gaia_get_lightcurve
from HCV_functions import HCV_get_lightcurves
from icecube_functions import icecube_get_lightcurve
from ztf_functions import ZTF_get_lightcurve
from heasarc_functions import HEASARC_get_lightcurves
from TESS_Kepler_functions import TESS_Kepler_get_lightcurves
from WISE_functions import WISE_get_lightcurves


# define some variables
mission_list = ["FERMIGTRIG", "SAXGRBMGRB"]
heasarc_radius = 0.1 * u.degree
bandlist = ["w1", "w2"]
wise_radius = 1.0
panstarrs_radius = 1.0 / 3600.0  # search radius = 1 arcsec
lk_radius = 1.0  # arcseconds
hcv_radius = 1.0 / 3600.0  # radius = 1 arcsec

# build up the sample
coords, labels = [], []
get_yang_sample(coords, labels)
coords_list, labels_list = remove_duplicate_coords(coords, labels)
object_names = [["CLAGN" + str(i)] for i in range(len(coords_list))]
```

### 2. Query archives in parallel

This uses [multiprocessing.Pool.apply_async](https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.Pool.apply_async). We want one worker process per call to `apply_async`. Total number of workers can be larger than the number of CPUs available because most calls use a relatively small amount of CPU and memory resources. Of note:

- WISE call uses more CPU than most. Parallelization will likely be limited by number of CPUs available. Plan on 1 CPU per call to `WISE_get_lightcurves`.
- ZTF call spends most of the time waiting for a response and using almost no CPU/memory resources. Parallelization will likely be limited by the ZTF service itself.

```python
# calculate the number of workers needed.
# these two must sum to total number of archives called
n_fast_archives, n_slow_archives = 6, 2
# coords_list will be split into this many chunks. we'll make one api call per chunk per slow archive
n_chunks_per_slow_archive = 15
# number of workers
n_workers = n_fast_archives + n_slow_archives * n_chunks_per_slow_archive

# MultiIndexDFObject to collect dataframes returned by each archive call
# only the main process will append to this, via the `append_df_lc` callback
df_lc = MultiIndexDFObject()

def append_df_lc(archive_df_lc):
    """Append `archive_df_lc` to `df_lc`. When a `pool.apply_async` call completes, 
    this function will be called on the result (`archive_df_lc`).
    """
    try:
        df_lc.append(archive_df_lc.data)
    except AttributeError:
        # this archive call returned no data
        pass

# start a multiprocessing pool and run all the archive queries
# without ztf this takes less than 2 min
# ztf can take significantly longer (maybe needs fewer simultaneous calls?)
with Pool(processes=n_workers) as pool:

    # start the processes that call the fast archives
    pool.apply_async(
        Gaia_get_lightcurve, (coords_list, labels_list, object_names, 1), callback=append_df_lc
    )
    pool.apply_async(
        HEASARC_get_lightcurves, (coords_list, labels_list, heasarc_radius, mission_list), callback=append_df_lc
    )
    pool.apply_async(
        HCV_get_lightcurves, (coords_list, labels_list, hcv_radius), callback=append_df_lc
    )
    pool.apply_async(
        icecube_get_lightcurve, (coords_list, labels_list, object_names, 3, "./data/", 1), callback=append_df_lc
    )
    pool.apply_async(
        panstarrs_get_lightcurves, (coords_list, labels_list, panstarrs_radius), callback=append_df_lc
    )
    pool.apply_async(
        TESS_Kepler_get_lightcurves, (coords_list, labels_list, lk_radius), callback=append_df_lc
    )

    # split coords_list into smaller chunks and call slow archives
    chunksize = ceil(len(coords_list) / n_chunks_per_slow_archive)  # num coords per api call
    for n in range(0, len(coords_list), chunksize):
        coords = coords_list[n : n + chunksize]
        labels = labels_list[n : n + chunksize]

        # start the processes that call the slow archives
        pool.apply_async(
            WISE_get_lightcurves, (coords, labels, wise_radius, bandlist), callback=append_df_lc
        )
        pool.apply_async(
            ZTF_get_lightcurve, (coords, labels, 0), callback=append_df_lc
        )

    pool.close()  # signal that no more jobs will be submitted to the pool
    pool.join()  # wait for all jobs to complete, including the callback (append_df_lc)

# look at the result
df_lc.data.sample(10)
```

### DIFF
--- a/light_curves/code/HCV_functions.py
+++ b/light_curves/code/HCV_functions.py
@@ -186,8 +186,8 @@ def HCV_get_lightcurves(coords_list, labels_list, radius):
     for ccount, coord in enumerate(coords_list):
 
         #doesn't take SkyCoord, convert to floats
-        ra = coords_list.ra.deg[ccount]
-        dec = coords_list.dec.deg[ccount]
+        ra = coord.ra.deg
+        dec = coord.dec.deg
         lab = labels_list[ccount]
 
         #IC 1613 from the demo for testing
--- a/light_curves/code/HCV_functions.py
+++ b/light_curves/code/HCV_functions.py
@@ -4,6 +4,10 @@ from tqdm import tqdm
 import pandas as pd
 from astropy.coordinates import SkyCoord
 
+from .fluxconversions import convertACSmagtoflux
+from .data_structures import MultiIndexDFObject
+
+
 ## Functions related to the HCV.
 def get_hscapiurl():
     """
@@ -160,13 +164,11 @@ def checklegal_hcv(table,release,magtype):
             raise ValueError("Bad value for magtype (must be one of {})".format(
                 ", ".join(magtypelist)))
             
-def HCV_get_lightcurves(df_lc, coords_list, labels_list, radius):
+def HCV_get_lightcurves(coords_list, labels_list, radius):
     """Searches Hubble Catalog of variables for light curves from a list of input coordinates
     
     Parameters
     ----------
-    df_lc : pandas multiindex dataframe
-        the main data structure to store all light curves
     coords_list : list of astropy skycoords
         the coordinates of the targets for which a user wants light curves
     labels_list: list of strings
@@ -180,6 +182,7 @@ def HCV_get_lightcurves(df_lc, coords_list, labels_list, radius):
         the main data structure to store all light curves
     """
 
+    df_lc = MultiIndexDFObject()
     for ccount, coord in enumerate(coords_list):
 
         #doesn't take SkyCoord, convert to floats
--- a/light_curves/code/HCV_functions.py
+++ b/light_curves/code/HCV_functions.py
@@ -4,8 +4,8 @@ from tqdm import tqdm
 import pandas as pd
 from astropy.coordinates import SkyCoord
 
-from .fluxconversions import convertACSmagtoflux
-from .data_structures import MultiIndexDFObject
+from fluxconversions import convertACSmagtoflux
+from data_structures import MultiIndexDFObject
 
 
 ## Functions related to the HCV.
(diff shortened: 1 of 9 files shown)


jkrick:
Awesome, Im starting to go through this. First question, we had talked about not sending in coords and labels lists as input  to the sample_selection functions, but instead having them generate those lists themselves.  But I see that didn't go into the PR.  Is that because it causes problems, or because it wasn't necessary for the multiprocessing, so we can leave it alone for now?

PR #95, opened 2023-04-24
Multiprocessing & cleanup

jkrick:
This PR does the following things:
- cleanup the code a bit, removing things not relevant, move ML description to it's own notebook, light editing of text
- moves the plot settings to a separate file to get them out of the clunky first cell
- add's a multiprocessing cell which calls all of the archive functions, and benchmarking times to compare to serial calls to the archives.
- I ran autopep8 on it, so it should be formatted prettily, but I can't honestly tell, so please comment if you think this didn't work.

@troyraen The multiprocessing cell only works properly if the slow archives are not split into 'chunks'.  When split into chunks, something happens to the resulting data frame that makes the plots all jumbled, so it is not currently a viable option. Of course, the real speedup comes from splitting the slow archives into chunks.


troyraen (commented):
Generally looking good. I'm still looking into some things, but wanted to capture these thoughts/suggestions before I quit for the day:

1. Remove the cell outputs from the checked-in notebook, multiband_lc.ipynb
2. I noticed some output indicating that at least one archive call actually downloads data and noticed that, when running the notebook start to finish in a clean environment, the download time would be counted against the serial calls but not the parallel calls (which don't have to re-download). If the timing info is not a main point of the notebook, you could maybe just add a comment stating that this is true (assuming it is), and suggesting the user run that particular archive at least once before doing the time trial so that the download is already done for both.
3. I would move this line out of the cell doing the multiprocessing and into its own cell: `print('parallel processing took', time.time() - starttime, 's')`. The multiprocessing cell produces a whole lot of output and can easily bury this output at the end of a long scroll window. Maybe also capture `time.time()` in a variable so that the second cell doesn't have to be called right after the first to give a correct answer.
4. The ZTF call gives me a key error (pre-existing issue), so I have to take it out. I'll look closer into that error.
5. I can't get the figure to render at all, even if I remove the "slow" archives altogether (in the parallel call). I get: `KeyError: 1`. Might be user-error. I'm still looking into it.

afaisst:
Merging (in coordination with Jessica) to get Notebook running again. Working on some plotting issues.

PR #97, opened 2023-04-26
update-faisst-apr25

afaisst:
Notebook worked, but I didn't try the multiprocessing part (only serial part).

Changes:
- IceCube function is now downloading data from the IceCube webpage (https://icecube.wisc.edu/data-releases/2021/01/all-sky-point-source-icecube-data-years-2008-2018/) directly (instead of from my Caltech Box)
- Now plotting IceCube data at the end, that way get_ylim() is used properly. Before, the y-normalization of the IceCube events was messed up because the get_ylim() wasn't taking the values of the final plot. This is fixed now.
- Jessica noticed that some Panstarrs data had outliers, which messed up the plot. I moved the sigma-clipping selection out of the Kepler/Tess plotting section. Had to do some other changes in the code to make it work properly. One fix I had to do is to limit the sigma-clipping to bands for which >10 data points are available (else sigma-clipping doesn't make sense). Other issues were related to indexing (since with sigma-clipping, some data points are removed)
- Modified Gaia lightcurve function slightly so that it now also shows the progress bars (looks better!). 
- Added the temporary directory where IceCube data will be downloaded to the .gitignore.
- Fixed a small bug in the plotting of the ZTF zoom-in. The colors of the curves were not the same as the ones in the main plot. Used *.get_color() to get the correct color.


jkrick:
looks good to me, thanks for cleaning up lots of loose ends!

PR #98, opened 2023-04-26
Multiprocessing cleanup

troyraen:
Included changes:

- Combine the multiprocessing callback (`append_df_lc`) with `df_lc.append` for simplicity
- Create an empty dataframe when a `MultiIndexDFObject` is instantiated so that we don't have to check for one before doing things like `append`.
- Split the multiprocessing cell into a few separate ones for readability


jkrick:
I never would have come up with that solution, so thanks for finding a clean way to have a single function to 'append'.

troyraen:
@afaisst Is it ok if we download the IceCube data to the `data` directory instead of a `data/tmp` directory? The latter doesn't exist in the repo, so if the notebook's going to use it it needs to create it first. Easier to just use `data` unless there's a reason not to.

troyraen:
@jkrick @xoubish two issues with the index to be aware of: 

1. Splitting the slow archives into chunks for parallel processing messes up the `objectid` because it's assigned based on the object's index in `coords_list`. That is likely why the plots are wrong in this case. 
2. I'm guessing that the `KeyError` that gets raised by the ZTF call sometimes is a floating point comparison issue with the `time` index. I think it would be best to use a single, integer-type index when doing `remove` (which raises the error), but there isn't a unique integer index here.

I'll think more about how to handle both of these.

jkrick:
> @afaisst Is it ok if we download the IceCube data to the `data` directory instead of a `data/tmp` directory? The latter doesn't exist in the repo, so if the notebook's going to use it it needs to create it first. Easier to just use `data` unless there's a reason not to.

I would also like to put it in data instead of tmp.

afaisst:
> > @afaisst Is it ok if we download the IceCube data to the `data` directory instead of a `data/tmp` directory? The latter doesn't exist in the repo, so if the notebook's going to use it it needs to create it first. Easier to just use `data` unless there's a reason not to.
> 
> I would also like to put it in data instead of tmp.

Yes, I guess that's fine since the unzipping should create a new directory. I made the tmp directory because the stuff in there can be deleted.

jkrick (changes requested):
I don't know how this double PR works, but the untimely_explorer directory is still in this PR even though it got taken out of the cleanup-imports one, so I don't know if that will put it back into the repo or not.  

Other than that, I think the PR looks good.

jkrick:
I have also run into issues with the object id later on in the ML work.  When I remove some of the sample from the df_lc, then I get holes in the object ids, which causes problems.  As you are thinking of a solution, would it make sense to have a function in the MultiIndexDFObject class which re-assaigns object ID's as a continuous range of ints, or have that be part of the .remove?  The actual values of these object IDs are not tied to anything scientifically, it's just a  convenient way to access the DF.

afaisst:
> I have also run into issues with the object id later on in the ML work. When I remove some of the sample from the df_lc, then I get holes in the object ids, which causes problems. As you are thinking of a solution, would it make sense to have a function in the MultiIndexDFObject class which re-assaigns object ID's as a continuous range of ints, or have that be part of the .remove? The actual values of these object IDs are not tied to anything scientifically, it's just a convenient way to access the DF.

This is a problem I mentioned very early on... (and in fact that was a bug I had to fix in the plotting code when I did the sigma clipping). I think we have to get the object IDs out of the df_lc and the loop over them instead of looping over the "coords" list. That way we can uniquely map from coords (or labels) to the df_lc....

bsipocz:
> I don't know how this double PR works, but the untimely_explorer directory is still in this PR even though it got taken out of the cleanup-imports one, so I don't know if that will put it back into the repo or not.

Do a rebase (or I'm happy to do it if you like), that should fix things up.

troyraen:
> I don't know how this double PR works, but the untimely_explorer directory is still in this PR even though it got taken out of the cleanup-imports one, so I don't know if that will put it back into the repo or not.

@jkrick It should be gone by the time everything gets merged, but I'll double check. Apologies, I know it can get confusing when PRs are chained together, and also making everyone be careful about when and how things get merged. I oscillate between creating one big PR that includes different sets of changes (and then everything can get bogged down by discussion about one of them) or creating separate PRs that depend on each other because I'm working on them at the same time and need them to line up properly. Lately preferring the latter, but happy to hear feedback about this if it causes headaches in the future.

I see that you requested a change; was that just this question about the directory or did I miss something else?

jkrick (approved):

jkrick:
yes, the question about the directory was the change. Good to merge from my perspective.  Fine by me to work the PRs however it is easiest for you.  I just need to learn to use github better, and seeing all the different options gives me more ideas.

troyraen:
@jkrick @afaisst Thank you for your comments about the object IDs, they're both very helpful.

>  When I remove some of the sample from the df_lc, then I get holes in the object ids, which causes problems. As you are thinking of a solution, would it make sense to have a function in the MultiIndexDFObject class which re-assaigns object ID's as a continuous range of ints, or have that be part of the .remove?

@jkrick that might be a really good solution, especially given the problem you're describing. @afaisst do I'll look more into what you're referring to and how the ID is used in the plotting to make sure I understand. Hope to put all these pieces of info together into a possible solution later this afternoon/evening. I'll put it in a separate PR and tag you both.

Issue #96, opened 2023-04-25
Fix imports in light curve notebook


troyraen:
Documenting some install/import issues that came up when trying to run the first cell of the multiband_lc.ipynb notebook in a clean environment (conda env + repo dir).

## Meisner et al.

Current code:

```python
try:
    from unTimely_Catalog_tools import unTimelyCatalogExplorer
except ImportError:
    if not os.path.exists('./unTimely_Catalog_explorer'):
        !git clone https://github.com/fkiwy/unTimely_Catalog_explorer.git
    sys.path.append('./unTimely_Catalog_explorer')
    from unTimely_Catalog_tools import unTimelyCatalogExplorer
```

Suggestions:

- append to the system path before trying to import the first time. otherwise I think the first call will (almost) never succeed.
- remove the directory `light_curves/unTimely_Catalog_explorer` from the repo (its presence causes the 4th line to always return false, so `git clone` never gets executed).
- this seems to require the package `reproject`. maybe should just `pip install` it right after `git clone`.

## Alerce

Current code:

```python
!pip install alerce
```

Suggestion:

- treat this like the others using a `try/except`. no need to run the install every time

troyraen:
@bsipocz was working on refactoring a lot of that cell, though I'm not sure whether her scope will cover these issues. If it doesn't, we should revisit some of these after her changes go through because they might affect the best way to handle these.

@jkrick, not thinking about Brigitta's changes for a minute, I think I'd do Meisner like this:

```python
sys.path.append('./unTimely_Catalog_explorer')
try:
    from unTimely_Catalog_tools import unTimelyCatalogExplorer
except ImportError:
    !git clone https://github.com/fkiwy/unTimely_Catalog_explorer.git
    !pip install reproject  # could go in a requirements.txt file
    from unTimely_Catalog_tools import unTimelyCatalogExplorer
```

Generally, a `try` statement should be as short as possible, only containing the line(s) of code that you actually want to try -- i.e., the one(s) that can trigger the error(s) caught by `except`. I think we can skip the check for an existing directory as well. If the repo has already been cloned the first import statement should work, but if it doesn't the `except` block will fail as well (with or without the check). The user will have to manually delete the directory and try again (re-clone it). Checking for the directory doesn't really gain anything and has the potential to hide the problem from the user by suppressing helpful error messages.

[...]

PR #99, opened 2023-04-28
Basic cleanup of imports

troyraen:
This addresses the basic cleanup items in #96. Other items involving more structural changes are left for a future PR.

Changes to the notebook:

- remove all imports that are not actually used in the notebook, except those that need to be installed (currently in try/except blocks)
- switch unTimely to use `pip install`
- do a try/except to install `alerce`, so that it matches the others
- sort the imports

Changes to the code files/modules:

- add imports that were missing
- remove imports not currently used in the file
- sort the imports

Other changes:

- remove the empty unTimely directory


light_curves/multiband_lc.ipynb
```diff
@@ -110,47 +91,39 @@
...
     "\n",
     "try:\n",
     "    from unTimely_Catalog_tools import unTimelyCatalogExplorer\n",
     "except ImportError:\n",
-    "    if not os.path.exists('./unTimely_Catalog_explorer'):\n",
-    "        !git clone https://github.com/fkiwy/unTimely_Catalog_explorer.git\n",
-    "    sys.path.append('./unTimely_Catalog_explorer')\n",
+    "    !pip install git+https://github.com/fkiwy/unTimely_Catalog_explorer\n",
+    "    !pip install reproject\n",
     "    from unTimely_Catalog_tools import unTimelyCatalogExplorer\n",
```
bsipocz, line 105: I suggest chopping up the imports and moving them to the beginning of the different big sections of the notebook. E.g. make it possible to obtain all the data without having any of the exploration tools installed/be ready.

light_curves/multiband_lc.ipynb
```diff
@@ -110,47 +91,39 @@
...
-    "from plot_settings import setup_text_plots\n",
+    "\n",
     "\n",
     "## NEW! LOAD/SAVE PICKLE FILE! ##\n",
     "SAVEDF = False # if set to True, pickle file will be saved. If False, pickle file will be loaded.\n",
     "#pickle_file_name = \"data/dflc_Shooby_March102023.pkl\" # name of pickle file (to be loaded or saved)\n",
     "pickle_file_name = \"data/dflc_Krick_Feb232023.pkl\" # name of pickle file (to be loaded or saved)\n",
-    "###\n",
-    "\n"
+    "###"
```
bsipocz, line 126: This should go when the notebook is ready to be shipped, but until then, please move to a separate cell.

bsipocz:
The notebook is very nicely modular regarding the datasets, so I wonder whether it would be more useful to users if those cells would be stand-alone, too? E.g. move the imports into the cells, so it's copy pasteable as is. OTOH, people would need to have the custom functions, too, so maybe this is not as trivial.

light_curves/multiband_lc.ipynb
troyraen: thanks, done

troyraen:
> move the imports into the cells, so it's copy pasteable as is

Happy to switch that but I'll let @jkrick weigh in first.

jkrick:
@bsipocz @troyraen on the issue of whether or not we should move the more specialized imports further down into their respective cells: I think this brings up another issue.  The code currently is nicely modular with each archive, but assuming we get the multiprocessing working, we could theoretically just have two working cells.  One to get the data and a second to plot it.  However, it is nice for someone coming to the notebook to learn something, to have each archive call be a separate cell.  So, do we keep the serial calls to each archive for the point of it being a tutorial, or do we get rid of them?  And based on that answer, we could keep all imports at the top, or move them down.

troyraen:
> it is nice for someone coming to the notebook to learn something, to have each archive call be a separate cell

To me, separate cells for the serial calls makes sense. And actually, either way, maybe those imports should stay at the top of the notebook because otherwise they should probably be duplicated in the serial and parallel sections, in case they are not executed in order.

jkrick (approved):
I just found this 'review changes' button that lets me officially say I have done a review.  Sorry for all the other times I missed this.  

Thanks for cleaning this all up, I think it looks good as is, and we can push the discussion of moving the imports down further into the cells to a later PR.

jkrick:
I guess my question was if we should remove the serial calls altogether, or keep them.

troyraen:
> I guess my question was if we should remove the serial calls altogether, or keep them.

Oh I see. In that case I think I still agree with this:

> it is nice for someone coming to the notebook to learn something, to have each archive call be a separate cell

It probably makes it easier to understand, and also easier because the user doesn't have to construct the code themselves if they want to try out a regular call without multiprocessing.

PR #100, opened 2023-04-29
Bug fixes for object IDs and key errors

### LINKED ISSUE #101: Bugs in the light curve notebook
- the ZTF call raises a `KeyError` sometimes 
- chunking the "slow" archives for parallel processing causes a bug in the `objectid` index
- the plot raises a `KeyError` every time

_Originally posted by @troyraen in https://github.com/fornax-navo/fornax-demo-notebooks/issues/98#issuecomment-1527475826_

1. Splitting the slow archives into chunks for parallel processing messes up the `objectid` because it's assigned based on the object's index in `coords_list`. That is likely why the plots are wrong in this case. 
2. I'm guessing that the `KeyError` that gets raised by the ZTF call sometimes is a floating point comparison issue with the `time` index. I think it would be best to use a single, integer-type index when doing `remove` (which raises the error), but there isn't a unique integer index here.

_Originally posted by @troyraen in https://github.com/fornax-navo/fornax-demo-notebooks/issues/98#issuecomment-1527435808_

troyraen:
Closes #101 

The notebook now runs end-to-end for me with no errors (nice work everyone 🎉 ). The plots look like reasonable light curves to me, but others should definitely inspect them.

## Included changes

Attach a persistent `objectid` to each `SkyCoord` object. The `objectid` is the index in the original, de-duplicated list. I put this change at the end of the `remove_duplicate_coords` function, and renamed that function to `clean_sample`. `coords_list` is now a list of tuples containing (objectid, coord). The archive calls are no longer responsible for constructing the `objectid`.

Code blocks which previously started like this: 
    `for ccount, coord in enumerate(coords_list):` 
now start like this: 
    `for objectid, coord in coords_list:`
In the rest of the code block, a direct `ccount` -> `objectid` substitution gives the desired, bug-free behavior.

The `objectid`s in `df_lc` now match the original index (i.e., they are no longer incremented by 1). If there was a particular need for the 1-based indexing, can someone let me know what that was? I changed it because Python is 0-based, so (a) I can't imagine a function/method having trouble with it, and (b) switching to 1-based makes it very easy for off-by-one indexing errors to creep in to the code.

Bug-fix the `KeyError` in the plotting cell. This was caused by syntax errors when accessing the multi-index in `.loc`. Note that if you're asking for a single value (e.g., a single `objectid`) that index level is completely removed from the dataframe that is returned.

Bug-fix the `KeyError` in the ZTF archive call. I'm fairly certain this was caused by a floating point comparison issue with the time. The fix was to not use the time here, but instead to drop everything with the same objectid, label, and band.

## Potential problems not addressed by this PR

The two lists `labels_list` and `object_names` are still implicitly assumed to have ordered, one-to-one relationships with the _original_ `coords_list` that is returned by `sample_selection.clean_sample`. This does not currently cause problems (that I am aware of), and it allows this PR to fix the current bugs without changing a lot of extra code. However this could easily cause bugs in the future because the relationships are not robustly enforced, thus anyone changing code without fully understanding the implicit relationships risks getting unexpected results.

It would be better to tie all of the input arguments for one astronomical/literature object together into a single python object. Implementing this would require significant changes to the code, so I'm guessing that we don't want to do it for this notebook. However, similar issues related to the organization of input arguments also came up in the forced photometry notebook. So I'll open an issue that describes my suggestions and can be a place for general discussion/ideas for future notebooks.

Finally, the changes here will not affect this issue, which will need to be addressed separately:

> I have also run into issues with the object id later on in the ML work.  When I remove some of the sample from the df_lc, then I get holes in the object ids, which causes problems.
_Originally posted by @jkrick in https://github.com/fornax-navo/fornax-demo-notebooks/issues/98#issuecomment-1528108764_


light_curves/code/TESS_Kepler_functions.py
```diff
@@ -26,19 +26,19 @@ def TESS_Kepler_get_lightcurves(coords_list, labels_list, radius):
     
     df_lc = MultiIndexDFObject()
     #for all objects
-    for ccount, coord in enumerate(tqdm(coords_list)):
+    for objectid, coord in tqdm(coords_list):
```
bsipocz, line 29: It's rather unfortunate that one has to loop through one by one, I would open a feature request to lightkurve to accept vector queries.

bsipocz (commented):
Some archives do support vectorized queries, and the ones who do tend to prefer it, too, so we may want to look into whether it's possible in any of the cases of not looping through objects one by one (generating a lot of queries), but sending fewer, vectorized ones.

troyraen:
@jkrick @afaisst @xoubish Will you all take a look at this, at your convenience? I think it touches stuff that everyone worked on.

jkrick:
Letting all of these changes settle into my brain.....
What/which change(s) do you think is/are responsible for the multiprocessing chunking now working properly?  ie., the plots look good to me now with chunking.

troyraen:
> What/which change(s) do you think is/are responsible for the multiprocessing chunking now working properly? ie., the plots look good to me now with chunking.

@jkrick Attaching a persistent objectid to each SkyCoord object up front so that the archive calls are no longer responsible for constructing it. (It used to be that each archive call would assign an objectid based on the object's index in the `coords_list` that get's passed in as an argument. So when the list got sliced into chunks for parallel processing, multiple different objects would end up with the same objectid, since the indexes for each chunk start over at 0.)

jkrick:
I'm going to add the 'vectorize archive calls' as an issue outside of this PR.  It's a good comment that we can ask the archives to do this for the benefit of all....

jkrick:
I think this looks great, and am very excited to see this notebook be fast *and* functional. Merge when you are ready!

afaisst:
This looks good!

light_curves/code/TESS_Kepler_functions.py
troyraen: moved to issue #104

PR #110, opened 2023-05-11
Cleanup

### LINKED ISSUE #103: Update documentation in code modules
I noticed docstrings in multiple code modules that have fallen out of sync with the actual code. The most common issue I saw was docstrings with incorrect types (e.g., the `MultiIndexDFObject` docstrings say that `x` is a dict, but it is not). I suggest that everyone review/update the docstrings for the functions they wrote. 

It would be good to also remove comments that are not intended for the general public, and commented-out lines of code that are left over from the development process.

jkrick:
Cleans up the code including:
- placing the pip installs into a requirements.txt file instead of cluttering up the first cell in the notebook
- cleaning up all docstrings in all functions to be uniform and correct to the best of my ability (I have no illusions of perfection). 
- changing the save file type from pickle to parquet
- removing any lines in the code not meant for public use

I did not touch the plotting cell with this cleanup effort because @afaisst is going to turn that into a function once we are sure it works well on all platforms.

This closes #103 and half of #96


light_curves/requirements.txt
```diff
@@ -0,0 +1,14 @@
...
+matplotlib
+pandas
+astropy
+astroquery
+tqdm
+lightkurve
+acstools
+alerce
+wget
+
```
troyraen, line 14:
```suggestion
git+https://github.com/fkiwy/unTimely_Catalog_explorer
reproject  # needed for untimely

```

light_curves/multiband_lc.ipynb
```diff
@@ -68,35 +78,11 @@
...
-    "try:\n",
-    "    from alerce.core import Alerce\n",
-    "except ImportError:\n",
-    "    !pip install alerce\n",
-    "    from alerce.core import Alerce\n",
-    "\n",
+    "#get tools for accessing Meisner et al., 2023 WISE light curves\n",
     "try:\n",
     "    from unTimely_Catalog_tools import unTimelyCatalogExplorer\n",
     "except ImportError:\n",
```
troyraen, line 88: I think these installs can go in the requirements file as well, and this whole try/except block can be deleted. (GitHub won't let me make the specific code suggestion since this block isn't changed by this PR.)

light_curves/multiband_lc.ipynb
```diff
@@ -487,9 +454,9 @@
    "outputs": [],
    "source": [
     "# calculate the number of workers needed\n",
-    "n_fast_archives, n_slow_archives = 6, 2  # must sum to total number of archives called\n",
-    "n_chunks_per_slow_archive = 5  # will make one api call per chunk per slow archive\n",
-    "n_workers = n_fast_archives + n_slow_archives * n_chunks_per_slow_archive"
+    "n_single_archives, n_multiple_archives = 6, 2  # must sum to total number of archives called\n",
+    "n_chunks_per_archive = 5  # will make one api call per chunk per slow archive\n",
```
troyraen, line 458:
```suggestion
    "n_chunks_per_archive = 5  # will make one api call per chunk per 'multiple' archive\n",
```

Clarifying the comment, since "slow" was removed from variable names.

light_curves/multiband_lc.ipynb
```diff
@@ -528,7 +495,7 @@
     "    )\n",
     "\n",
     "    # split coords_list into smaller chunks and call slow archives\n",
```
troyraen, line 497:
```suggestion
    "    # split coords_list into smaller chunks and call remaining archives\n",
```

troyraen (approved):
Looks good. I added a few suggestions where I could.

light_curves/multiband_lc.ipynb
jkrick: ok, thanks, pushed as a separate commit

bsipocz (approved):
All looks good to me, thanks @jkrick!

PR #111, opened 2023-05-25
Switch to the new parquet catalog for WISE

### LINKED ISSUE #117: Speedup WISE_get_lightcurves
Suggestions for how to speedup this section:

1. Work with the parquet call:
  - is there a way to vectorize finding which partition an object is in?
  - potentially make a dict or pandas df with hp.query_circle
  - goal is to be able to do a sort of groupby on the coords of the objects
  - This is a general question of how to efficiently search the whole sky for a large list of coords that will benefit other use cases

2. Take all the other things out of the coordinate search
 - flux conversion 
    - make a list of result_table and concatenate that into one object either astropy table or a data frame.

3. Think about weather or not we have a dfsingle.append or collect all the dfsingles and append later 
 - This might already be changed as a ripple effect of the above suggestions.

troyraen:
Closes #117 

The Meisner et al., 2023 catalog has been converted to parquet and uploaded to an IRSA AWS bucket for testing. This PR updates the `WISE_get_lightcurves` call to use it. 

In initial tests, this reduces the time it takes to load 30 light curves (Yang sample) from 8min 45sec using the original catalog down to 1min 45sec using the parquet catalog.

To Do:

- [x] time precision: mjdmean is slightly different between parquet and original. why? does it matter?
- [x] flux units:  see code and comment below.
- [x] num rows: new call returns more rows than original (using unTimelyCatalogExplorer). need to understand why. original returns only one coadd_id per objectid. is [this](https://github.com/fkiwy/unTimely_Catalog_explorer/blob/954b06b5dc91c207c335e652083327c4ec136dd9/unTimely_Catalog_tools.py#L535) intentionally restricting to just the first item?
- [x] test how this scales with the number of sources. If it scales well, remove WISE from the "slow" archives in the parallel processing cell.


troyraen:
The parquet files need additional verification before they're ready to publish. This PR will be on hold until that's done.

light_curves/code/fluxconversions.py
```diff
@@ -23,22 +24,55 @@ def convert_WISEtoJanskies(mag, magerr, band):
...
-            
-    flux_Jy = zpt*(10**(-mag/2.5))
-    
-    #calculate the error
-    magupper = mag + magerr
-    maglower = mag - magerr
-    flux_upper = abs(flux_Jy - (zpt*(10**(-magupper/2.5))))
-    flux_lower = abs(flux_Jy - (zpt*(10**(-maglower/2.5))))
+    # wise_df["flux"] = convert_wise_flux_to_millijansky(nanomaggy_flux=wise_df["flux"])
+    # wise_df["dflux"] = convert_wise_flux_to_millijansky(nanomaggy_flux=wise_df["dflux"])
```
troyraen, line 28: Photometry is stored in the catalog as fluxes "in Vega nanomaggies". So, it seems like we should be able to do a direct unit conversion like these commented-out lines of code (function defined below). But, this is not returning the same fluxes as the original code. I suspect this is related to zero points.

The original code received magnitudes from unTimelyCatalogExplorer and converted them flux, so the code below does the same transformations (flux -> magnitude -> flux) in order to produce the same results as the original.

troyraen: @jkrick will you take a look at these flux conversions? If this is confusing, I'm happy to talk through it on a call.

To try to be a little more clear, right now this code is returning the same fluxes as the original code. If that is correct, there is no need to change anything. It seems unnecessary to go flux -> magnitude -> flux, but it doesn't cause any problems (compute time is negligible). I just want someone else to look at it because I'm getting confused about why the direct unit conversion doesn't work, and whether/why different zero points are used in the flux -> magnitude vs the magnitude -> flux.

troyraen:
I've convinced myself that the mjdmean values we're getting here from the Parquet files match the ones in the original FITS files. The values are altered slightly somewhere along the trip between the FITS files and what we were getting back from unTimelyCatalogExplorer in the original call. The difference is a couple of minutes.

troyraen:
This loads 500,000 sources in about 4.5 hours. 

The load time is scales very well with the number of sources because the number of partitions/pixels it has to dig through starts to saturate at around 25,000 sources.

wise-loadtime-vs-nobjects
wise-loadtime-vs-npixels

troyraen:
@jkrick In a sample of 500,000 objects, I found 3 negative fluxes. These end up as NaN in the final dataframe because they cannot be converted to magnitudes. Should we drop those from the dataframe? Or do something else with them?

jkrick:
> @jkrick In a sample of 500,000 objects, I found 3 negative fluxes. These end up as NaN in the final dataframe because they cannot be converted to magnitudes. Should we drop those from the dataframe? Or do something else with them?

Thanks for finding those.  We can drop these from the data frame.  I think I drop nans later in my ML code, but can't hurt to do it here.  Clearly a problem if the fluxes are negative, so not worth keeping those around.

jkrick:
It runs for me on the SMCE.  My only comment is that I would like to see more comments inside the functions in WISE_functions.py.  I don't know enough about what you are doing, so some helpful words could go a long way here.

troyraen:
Re: problem with number of rows
I spoke with Meisner and learned that we should not select sources based on the coadd_id, but we should select only sources within the coadd's primary region to avoid duplicates when an object is near the coadd boundary. I've implemented the change.

PR #112, opened 2023-05-31
Progress outputs

jkrick:
Two quick changes:
1) code now outputs the number of sources for a given radius.  The time that the code takes to run should scale roughly linearly with the number of sources.  This should make it easier to calculate the time it will take to run at full scale of radius = 40.
2) added a progress bar to the multiprocessing cell by wrapping it in tqdm.  The total number in the progress bar is not really the number of sources but is the number of parameters which go into the forced photometry function which is six times the number of sources.  For example, the progress bar will show n/600 for 100 sources.


jkrick:
Only for forced_photometry notebook.

PR #116, opened 2023-06-07
Updates to enable full run

zoghbi-a:
The main updates are:
- Switched the first `cosmos2015` query to use TAP service with pyvo instead of astroquery. The latter was using a lot of memory when requesting the full table.
- Switched from `concurrent` to `multiprocessing` when running the tractor in parallel. The main dataframe is now updated with the results after the whole run is complete instead of doing it inside the loop. These appear to prevent the staling. I don't know exactly why. 
- I added the draft `pyvo` version that deals with the cloud columns. It is not fully working yet, but it should be added for the final demo

### DIFF
--- a/forced_photometry/multiband_photometry.ipynb
+++ b/forced_photometry/multiband_photometry.ipynb
@@ -158,6 +158,7 @@
     "from astroquery.ipac.irsa import Irsa\n",
     "from astroquery.heasarc import Heasarc\n",
     "from astroquery.mast import Observations\n",
+    "import pyvo\n",
     "\n",
     "# Local code imports\n",
     "sys.path.append('../code/')\n",
@@ -214,6 +215,7 @@
     "Irsa.ROW_LIMIT = 3E6  \n",
     "Irsa.TIMEOUT = 600\n",
     "\n",
+    "#pull a COSMOS catalog from IRSA using pyvo\n",
     "\n",
     "#what is the central RA and DEC of the desired catalog\n",
     "coords = SkyCoord('150.01d 2.2d', frame='icrs')  #COSMOS center acording to Simbad\n",
@@ -232,7 +234,16 @@
     "    'r_MAGERR_AUTO', 'FLUX_24', 'FLUXERR_24', 'MAG_GALEX_NUV', 'MAGERR_GALEX_NUV',\n",
     "    'MAG_GALEX_FUV', 'MAGERR_GALEX_FUV'\n",
     "]\n",
-    "cosmos_table = Irsa.query_region(coords, catalog=\"cosmos2015\", radius=radius, selcols=','.join(cols))\n",
+    "#cosmos_table = Irsa.query_region(coords, catalog=\"cosmos2015\", radius=radius, selcols=','.join(cols))\n",
+    "\n",
+    "# The above line works for radius<=40arcmin; for the full 48arcmin; do it with pyvo\n",
+    "tap = pyvo.dal.TAPService('https://irsa.ipac.caltech.edu/TAP')\n",
+    "result = tap.run_sync(\"\"\"\n",
+    "           SELECT {}\n",
+    "           FROM cosmos2015\n",
+    "           WHERE CONTAINS(POINT('ICRS',ra, dec), CIRCLE('ICRS',{}, {}, {}))=1\n",
+    "    \"\"\".format(','.join(cols), coords.ra.value, coords.dec.value, radius.value/60.))\n",
+    "cosmos_table = result.to_table()\n",
     "\n",
     "\n",
     "print(\"Number of objects: \", len(cosmos_table))\n"
@@ -1211,7 +1222,8 @@
     "        img_pair = lookup_img_pair(sci_bkg_pairs, band.idx, row.galex_image)  # file paths only\n",
     "        paramlist.append(\n",
     "            [row.Index, band, row.ra, row.dec, row.type, row.ks_flux_aper2, img_pair, df]\n",
-    "        )            "
+    "        )\n",
+    "print('paramlist: ', len(paramlist))"
    ]
   },
   {
@@ -1256,8 +1268,10 @@
     "#wrapper to measure the photometry on a single object, single band\n",
     "def calculate_flux(args):\n",
     "    \"\"\"Calculate flux.\"\"\"\n",
-    "    f = calc_instrflux\n",
-    "    val = f(*args[1:])\n",
+    "    val = calc_instrflux(*args[1:])\n",
+    "    # add simple logging\n",
+    "    if (args[0] % 100) == 0 and val[0] == 0:\n",
+    "        with open('output.log', 'a') as fp: fp.write(f'{args[0]}\\n')\n",
     "    return(args[0], val)"
    ]
   },
@@ -1267,21 +1281,18 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%time\n",
-    "#Here is where the multiprocessing work gets done\n",
-    "t2 = time.time()\n",
-    "outputs = []\n",
-    "# number of processes can be controlled with a max_workers arg in ProcessPoolExecutor\n",
-    "# defaults to the number of available processors, beyond which no performance gains are seen\n",
-    "with concurrent.futures.ProcessPoolExecutor() as executor:\n",
-    "    for result in list(tqdm(executor.map(calculate_flux, paramlist), total = len(paramlist))):\n",
-    "        # print(result)\n",
-    "        df.loc[result[0],\n",
-    "                  'ch{:d}flux'.format(result[1][0] + 1)] = result[1][1]\n",
-    "        df.loc[result[0],\n",
-    "                  'ch{:d}flux_unc'.format(result[1][0] + 1)] = result[1][2]\n",
-    "        outputs.append(result)\n",
-    "t3 = time.time()"
+    "fname = f'results_{radius.value}.npz'\n",
+    "if os.path.exists(fname):\n",
+    "    results = np.load(fname, allow_pickle=True)['results']\n",
+    "else:\n",
+    "    from  multiprocessing import Pool\n",
+    "    t0 = time.time()\n",
+    "    with open('output.log', 'w') as fp: fp.write('')\n",
+    "    with Pool() as pool:\n",
+    "        results = pool.map(calculate_flux, paramlist)\n",
+    "    dtime = time.time() - t0\n",
+    "    np.savez(fname, results=results)\n",
+    "    print(f'Parallel calculation took {dtime} seconds')"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


forced_photometry/multiband_photometry.ipynb
```diff
@@ -232,7 +151,14 @@
     "    'r_MAGERR_AUTO', 'FLUX_24', 'FLUXERR_24', 'MAG_GALEX_NUV', 'MAGERR_GALEX_NUV',\n",
     "    'MAG_GALEX_FUV', 'MAGERR_GALEX_FUV'\n",
     "]\n",
-    "cosmos_table = Irsa.query_region(coords, catalog=\"cosmos2015\", radius=radius, selcols=','.join(cols))\n",
```
bsipocz, line 235: Please keep this one even if it's more resourceful atm than the direct TAP query (I'm working on switching out the backend in astroquery). Users are definitely not required to know about TAP, and working with coordinates this way is not really the best practice we try to show anyway (the premise of they don't need to track their coordinates are lost when they are forced to fall back to use the values as is, etc)

bsipocz:
- `concurrent` vs `multiprocessing`: @troyraen or @stargaser or @jkrick  should comment on it, I recall we went a bit back and force and they looked under the hood.

forced_photometry/multiband_photometry.ipynb
zoghbi-a: Thanks @bsipocz,
I did this because the notebook was just failing as it exhausts the available memory on a small compute job.

bsipocz: This notebook used to be able to run, are you trying somewhere else, or with a large area?

zoghbi-a: Yes. I am working with @jkrick to run the full area of the catalog.

troyraen:
The use of `concurrent.futures` predates my involvement. @stargaser maybe? 

Using `multiprocessing.Pool` makes sense to me. If `pool.map` is running smoothly, then great! But if there are problems, like if the main process is holding onto a lot of memory, `pool.imap` or `pool.imap_unordered` might be better. `imap` is a lazier version of `map` and can be helpful in cases like this where the iterable (`paramlist`) is very long.

stargaser:
I originally used `concurrent.futures` mainly for being more familiar with it. Using `multiprocessing.Pool` makes sense per Troy's comments.

jkrick:
Section 1. I agree with Brigitta that the pyvo query is not as streamlined as the astroquery query.  From a user perspective then, I prefer the astroquery because it is easier to write and easier to read.  The time this cell takes to complete is negligible compared to the rest of the notebook, so speed is not my concern here.  My understanding is that AZ will demo the pyvo capabilities before this notebook demo for HQ, so those capabilities will be covered elsewhere.  Can we leave in examples of both queries, with one commented out?

Section 2. Retrieve Image datasets from the cloud, which didn't work before, still doesn't work for me.  I get an error in the pyvo.regsearch.  I think it would be cleaner to leave this out of this PR since it is a different topic, and work on this separately. I know that @bsipocz is thinking about how best to retrieve the IRAC images.  The original version doesn't throw an error, it just doesn't do what we want it to do (which is better for running the entire code at once).

Section 3 works great with multiprocessing!  It takes 417s on the small SMCE for radius = 0.5', and used to take 426s with concurrent.futures.  I have not yet tested it on a large radius, but no need to do that before merging.

jkrick:
Section 1. Yes astroquery fails on a radius of 48', but works up to a radius of 40' and Brigitta is working on that backend anyway.  I think we are all in agreement that we keep both in there for reference.

Section 2. On the first cell: `irsa_cosmos = pyvo.regsearch(ivoid='ivo://irsa.ipac/cosmos/images')[0]`

I get the following error:
`---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [4], in <cell line: 2>()
      1 # Getting the COSMOS address from the registry to follow PyVO user case approach. We could hardwire it.
----> 2 irsa_cosmos = pyvo.regsearch(ivoid='ivo://irsa.ipac/cosmos/images')[0]
      4 # The search returns 11191 entries, but unfortunately we cannot really filter efficiently in the query
      5 # itself (https://irsa.ipac.caltech.edu/applications/Atlas/AtlasProgramInterface.html#inputparam)
      6 # to get only the Spitzer IRAC results from COSMOS as a mission. We will do the filtering in a next step before download.
      7 cosmos_results = irsa_cosmos.search(coords).to_table()

TypeError: search() got an unexpected keyword argument 'ivoid'`


And yes, I agree the speed was not the motivation for switching to multiprocessing, I was just curious if there was any difference between multiprocessing.pool and concurrent.futures.

zoghbi-a:
> Section 2. On the first cell: `irsa_cosmos = pyvo.regsearch(ivoid='ivo://irsa.ipac/cosmos/images')[0]`

Searching by `ivoid` is `pyvo>1.3` I think. What version are you using?

zoghbi-a:
[91ac740](https://github.com/fornax-navo/fornax-demo-notebooks/pull/116/commits/91ac7402bfd1796e5939010b977f44f938b3f930) puts back the `astroquery` code, and the irsa image search.

@jkrick, please let me know if there more changes that are needed.

zoghbi-a:
The last commits are to allow the notebook to run in non-interactive mode

jkrick:
> > Section 2. On the first cell: `irsa_cosmos = pyvo.regsearch(ivoid='ivo://irsa.ipac/cosmos/images')[0]`
> 
> Searching by `ivoid` is in `pyvo>=1.3` I think. What version are you using?

I am using whatever version gets installed with the requirements.txt file.  I don't know how to tell what version that is.  Also, looking at that requirements.txt file, do we need to use your fork of pyvo?  That seems unsustainable.  Would the notebook still work if we just had `pyvo` in the requirements.txt file?

bsipocz:
If we require a minimum version of a package, then the requirements file needs to be updated. 
Then of course the question is of how to sync and upgrade the requirements for a user who already have their set up is a separate problem. We can think about the "finished" notebooks as frozen versions, and now this a one of problem during development, but in practice versions won't be that frozen.

Also, the question of upgrades to the kernels on the SMCE side is afaik not a resolved issue. Some of these were started to be tracked in https://github.com/fornax-navo/fornax-images, but also on gitlab, and the internal ipac SP.

zoghbi-a:
@jkrick, My fork of pyvo is to enable the cloud features. The notebook currently does not use it as the the part of code is commented out, but the plan is to use it at some point, whether it is for this demo or later is TBD.

@bsipocz, I removed that part of the code, so there is no need for version specification

jkrick:
I think the PR is ready to merge, thanks for making this happen so that we have a fully functional notebook!  I don't understand most of what @bsipocz is referring to, so I'll let her comment on that.

bsipocz:
I would still remove the changes to the data acq cells, as if there are problems with using the current fornax workaround, then it should be fixed rather than worked around with direct urllib.request. Also, would be nice not to leave commented out code in the notebook as ultimately we need to cleanup everything for the demo.

So, what I would suggest: 1) make a copy of this branch 2) squash all things together to leave only the changes to the multiprocessing parts.

In the meantime, we already talked about addressing the COSMOS vs SEIP question, and to use the cloud cutout API rather than full image downloads.

zoghbi-a:
Is the first part of `astroquery` working, @bsipocz?
The reason those changes are there is that the engineers at SMCE want something that they can run without having to make any code change. The goal is make sure the whole thing runs.

jkrick:
> Is the first part of `astroquery` working, @bsipocz?
> The reason those changes are there is that the engineers at SMCE want something that they can run without having to make any code change. The goal is make sure the whole thing runs.

The astroquery IRSA catalog search currently works at radius= 40', but not 48.

zoghbi-a:
@bsipocz, I rechecked the original code and the following part is not working correctly. The data is not being downloaded. That is why I did a direct `urllib` download, so that we (and the engineers) can debug other parts of the notebook, while we fix this separately.

> `fornax_download(spitzer, access_url_column='sia_url', fname_filter='go2_sci', 
                data_directory='../data/IRAC', verbose=False)`

bsipocz:
OK. The I suggest to keep that temporary workaround in a separate branch, and squash things here and keep only the fixes that should go into production.

bsipocz:
> so I would like the current PR to go through as is.

Then, please make a new cell for the workaround. The current PR removes the code that worked at some point to do the download, but apparently is now buggy in favour of a workaround, while it still adds commented-out code that is not proposed to be used for the demo at all. To ultimately fix it, we need to dig the part back up again that this PR removes.
Ultimately, making direct urllib.request calls won't tick any boxes for the HQ demo while the part that got removed does. If it's broken, then yes, it needs to be fixed, but this workaround is not demoable as is, as mentioned above, it was added only as a workaround to fix the later parts.

Or, alternatively, I can start working on fixing those with the version from the base of this PR parallel, and then we can try to get the fixes merged together. The bottom line, we need to separate the parts at some points anyway. I agree it's not ideal, especially that we are bound by the limitations of notebooks.

zoghbi-a:
This last commit ([17b1f8d](https://github.com/fornax-navo/fornax-demo-notebooks/pull/116/commits/17b1f8d6aecb60071ad42f060f566f14674e49dd)) keeps only the necessary changes needed to the run the notebook as a whole. 

- The `urllib` download part is  there temporarily so we can debug others issues (memory, cpu etc.). We will go back to what is in `fornax-cloud-access-API` or `pyvo.utils.cloud` once they are working. 
- The notebook uses TAP service so we can run the full 48arcmin until the `astroquery` issue is resolved.
- The `multiprocessing` part is kept as it is.

bsipocz:
I'm fairly certain the bug is in the fornax code and how `download_file` is directly used from astropy without explicitly stating `local_path` as a kwarg (e.g. running the cell takes some non negligible time).

bsipocz:
> Section 1. Yes astroquery fails on a radius of 48', but works up to a radius of 40'

And is 48' the full dataset? I dived in a tiny bit, but fail to find a number for how big the "COSMOS" field actually is.

zoghbi-a:
> I'm fairly certain the bug is in the fornax code and how `download_file` is directly used from astropy without explicitly stating `local_path` as a kwarg (e.g. running the cell takes some non negligible time).

Thanks @bsipocz for adding a fix in the other repo. I am a little confused though. IThe code is meant to grab the data from aws, so fixing the on-prem part is not going to fix why aws is not working. I plan to spend some time today to fix this.

bsipocz:
Yes, I'm doing the aws part and will hopefully have the fix before the being in the afternoon. As I mentioned the issue is likely be due to the restricted nature of this particular bucket, and I indeed needed to fix the on prem download part, so those without the aws credentials can still run the notebook.

zoghbi-a:
I think I found the problem. There is a bug in `fornax`. I will fix it

jkrick:
> And is 48' the full dataset?

yes radius = 48' is the full COSMOS.

bsipocz:
Rebased, squashed and removed all unrelated changes. 

A copy of your original branch is located here: https://github.com/bsipocz/fornax-demo-notebooks/tree/zoghbi-a_copy_az_updates_beforemerge

zoghbi-a:
```diff"
- mpld3.display(fig)
+ #mpld3.display(fig)
```
This part is also needed in the last cell.

bsipocz:
> - mpld3.display(fig)
+ #mpld3.display(fig)

Ahh, I left that one in out saying that we should be able to run it interactively.

bsipocz:
The commented out mpl line has been discussed at the meeting, so I went ahead with the merge. The notebook runs on SMCE for me, though I canceled its run in the multiprocess cell.

Issue #107, opened 2023-05-03
speedup ZTF light curve search 


jkrick:
What are potential ways to speed up the ZTF light curve search to help run the light curve notebook at scale?

Currently we wrap an IRSA light curve API search in a for loop over each ra and dec of interest. 
Potential speedups include:
1. vectorize this light curve search to accept a table of positions so that there is no for loop, and only one archive call
2. write a wrapper for the publicly available parquet files
      Potential problems with this idea that might impede the hoped for speedup.  These parquet files are not formatted with hipscat, instead by ZTF field, so this is not an obvious wrapper to write.   The parquet files are at IRSA, so are local to the IRSA SP, but not to the SMCE.

Solutions which avoid the API have the potential to be fast because they don't have to share the API with other users, which can cause a factor of 10 slowdown in getting the light curves.

jkrick:
From short discussion with @sgroom, IRSA does not have funding from ZTF to vectorize the query, so the IRSA query will only be 1 object at a time.  

@troyraen wondered out loud if we could get ZTF light curves from any of the alert brokers
@bsipocz wondered out loud if we could get ZTF light curves from other light curve users.

jkrick:
slight improvements in speed can be made by taking some things out of the for loop over each object which does the API call.  That for loop should just do the API call, then things like unit conversion or checking for duplicates could be vectorized seperately.

PR #118, opened 2023-06-13
Get ZTF data from parquet files instead of light curve service

troyraen:
Note: I am using this PR as a demonstration of this data access method for informational and discussion purposes. I do not intend to actually merge it.

The current bottleneck in scaling up the light curve notebook is the ZTF light curve service, so we're looking at other options to get the data (see #107). This PR switches the data access method in `ZTF_get_lightcurve` to use the ZTF [parquet files](https://irsa.ipac.caltech.edu/data/ZTF/lc/lc_dr17/). 

The parquet files are organized by ZTF field, CCD, quadrant, and filter. Unfortunately, the only method I could find to lookup which files contain a given object (ra/dec) is far too slow, taking more than 30 seconds per object. If we are going to use parquet files, I think we should repartition them to be the same as the others we're currently doing (i.e., by HEALPix order 5). There are other options for getting the data, like alert brokers or others in the community, so we're looking into those as well.

I'll leave this open for a few days, in case anyone wants to comment. Then I'll close it. Pinging @xoubish @jkrick @stargaser @bsipocz

### DIFF
--- a/light_curves/code/ztf_functions.py
+++ b/light_curves/code/ztf_functions.py
@@ -1,43 +1,24 @@
+import re
+
+import astropy.units as u
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
-import requests
-from alerce.core import Alerce
 from astropy.coordinates import SkyCoord
-from astropy.io import ascii
-from tqdm import tqdm
-
+from astropy.table import Table, vstack
 from data_structures import MultiIndexDFObject
+from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


troyraen:
From discussion in the SP meeting yesterday, I'll try using a TAP query in place of the particular call in this PR that takes 30 seconds per object. That might allow us to go forward using the parquet files here. It might also demonstrate a key step or two that could be used to improve the light curve service.

PR #119, opened 2023-06-13
updated plotting

afaisst:
CHANGES:
- plotting in now done in separate function ( create_figures() )
- I renamed "plot_settings.py" to "plot_functions.py". The file includes the plot settings as well.

NOTE: I noticed that when I run the query in parallel mode, it doesn't give me the ZTF lightcurves.


jkrick (approved):
Looks good to me.
Very minor, but can you change the title of the PR to include 'forced photometry' so we know which notebook is being worked on?
Otherwise, please merge.

PR #123, opened 2023-06-28
Switch `ZTF_get_lightcurve` to get data from parquet files

troyraen:
Partially addresses #107 

- Get data from [ZTF parquet files (dr18)](https://irsa.ipac.caltech.edu/data/ZTF/lc/lc_dr18/) instead of light curve service.
- Use TAP to lookup the ZTF field, CCD, and quadrant info needed to find the object in the parquet files
- The structure of the loaded data is significantly different than what was previously being loaded from the light curve service, so the rest of the code has been updated to accommodate it.
- This no longer makes plots or displays a `tqdm` progress bar. This is because it no longer does a single for-loop over the objects. I think those were mostly for debugging, but if they are actually still wanted I can put them back in.
- Adds the option to create a `MultiIndexDFObject` with data directly rather than having to create an empty `MultiIndexDFObject` and then append the data.

With these changes, one run of the Yang sample (30 objects) took about 2 minutes which is on-par (but at the high end) with the other functions grabbing light curves.

## To do before merging

- [x] handle "repeated lightcurves"
- [x] convert hmjd -> mjd
- [x] check that this returns the same results as the original code

## Details

Data is fetched in two steps:

1. For each object:  Call TAP service to lookup the ZTF field, etc. needed to find this object in the parquet files. 
    - Takes ~3 seconds per object. estimating 17 days for 500,000 objects (without parallelization)
    - TAP service limits to one object per call
2. For each parquet file:  Load data using a filter for objectid (returned by Step 1)
    - Takes ~30-45 seconds for all files (Yang sample, 30 objects)
    - This probably scales better than 1-to-1 with the number of objects because multiple objects will be pulled from each file.

### Thoughts on parallelization (future work)

Right now, Step 1 takes 3x longer than Step 2.

One option would be to run Step 1 continuously as a background job, periodically returning batches of results. The main process could then run each batch through Step 2 and be done before the next batch is returned.

If that is too slow, Step 1 could be parallelized by making several simultaneous TAP calls. If these are returned faster than Step 2 can handle serially, run batches in parallel by launching more background workers.


troyraen:
@xoubish a comment in the original code says:

> ```
> ## Neeeded to remove repeated lightcurves in each band. Will keep only the longest following (Sanchez-Saez et al., 2021)
> ```

Would this "repeated lightcurves" problem be solved by using just the single object that is closest to the sample coordinates? Or do we really want the object with the longest light curve, even if it's not the closest spatially?

xoubish:
That is a valid point. Please check and if by changing the coordinate match strategy only one light curve per waveband is left you are right. I thought that there might have been multiple light curves saved per band for the same object.

bsipocz:
Afaik there may be duplicates for objects that fall on multiple chips?

xoubish:
You are right. Just checked the mentioned paper: "When doing this for a particular object, the ZTF API can return
more than one light curve in a given filter. This happens because ZTF treats independently the light curves observed in a particular field, filter, and CCD-quadrant, and thus, if a source appears in more than one combination of these three, ZTF will provide more than one light curve. "

troyraen:
The original code was returning MJD, but these parquet files have HMJD ("Heliocentric-based Modified Julian Date"). Does anyone know how to convert HMJD -> MJD? I don't see this method in astropy or anywhere else.

troyraen:
@xoubish The Sánchez-Sáez paper says they keep the light curve with longest timespan:

> to avoid having multiple light curves for any particular source, we only used the longest light curve (in terms of its timespan) of each filter of a particular object.

I interpreted this as meaning the largest difference between max and min MJD. But if I'm reading your original code right, I think you were keeping the one with the largest number of data points. Can you confirm which one you actually want?

bsipocz:
> The original code was returning MJD, but these parquet files have HMJD ("Heliocentric-based Modified Julian Date"). Does anyone know how to convert HMJD -> MJD? I don't see this method in astropy or anywhere else.

would this help? https://docs.astropy.org/en/stable/time/index.html#barycentric-and-heliocentric-light-travel-time-corrections

bsipocz:
(I somewhat fail to see why they used heliocentric, rather than barycentric time, but I suppose the difference small enough to treat it as barycentric. Then there are already a couple of scales for it in astropy: https://docs.astropy.org/en/stable/time/index.html#time-scale (note that mjd is a format, while here we talk about the scale, nevertheless I would suggest to check out both sections in the docs as it clears up the concepts nicely). )

troyraen:
@bsipocz thank you.

Can I ask a scientist (maybe @xoubish ?) to take on the task of converting the HMJD into whatever format you want in the `MultiIndexDFObject`? I can leave it as HMJD for now and open an issue to remind us.

bsipocz:
I suppose if we properly parse stuff into an astropy Time instance, then it should work out of the box. But then of course conversions need to be done when moving to pandas, etc as those know nothing about astronomical times.

troyraen:
I confirmed with Shooby that she wants the light curve with the most datapoints, not (necessarily) the one with the largest delta-MJD. I'm making that change now.

Unless there are objections, I'll merge this by the end of the day so it can be included in tomorrow's HQ demo dry run (@xoubish @bsipocz @jkrick?). I'll open Issues for anything that still needs to be done, but I think nothing that remains should prevent this from being included in the HQ demo.

jkrick:
No objections, but please confirm that this doesn't cause any problems downstream, since I might not have the chance to check it.

troyraen:
> please confirm that this doesn't cause any problems downstream

Meaning, 'make sure the whole notebook runs and the light curve plots look reasonable'?

troyraen:
Spoke with Jessica and decided to leave HMJD as-is and add a note that if the science depends on precise times this will need to be corrected.

troyraen:
Reading the parquet files over HTTP turned out to be much slower from the SMCE than reported in the PR's main comment (which was based on tests from the IRSA science platform), and is also not the recommended way to access those files (recommended to download). We've decided to upload the files to S3, though the specific bucket is still being determined. In the meantime, I've uploaded them to a testing bucket, and updated the code, so that this PR can move forward.

troyraen:
@jkrick @bsipocz this is ready, if you want to review. The files are still in-transit to the S3 bucket. Hopefully that will be complete by tomorrow morning. The Yang sample can be used for testing -- all of those files are in the bucket already.

I think this is still going to be slow relative to the other archive calls. I need to do more time testing to see how it scales. I have experimented with using a "helper" dataset to replace the TAP query, which I think I've mentioned offline. It looks like it could make that step 10 times faster.

This PR makes a necessary change in order to run at scale: using the parquet files instead of the light curve service. I think this PR should be merged and future improvements can go in other PRs.

troyraen:
I believe all the parquet files are in the S3 bucket now.

jkrick:
working on the IRSA SP:
in requirements.txt I added hpgeom and s3fs, but the second cell crashes with a few errors, it seems quite snarled.  last error is a PermissionError: Forbidden  
I don't know how to solve this, so switching to SMCE...

working on the SMCE: 
in requirements.txt I needed to add hpgeom to make the second cell in the notebook pull the ZTF functions
Then I can run the light curve call for a sample of 30 and it takes ~3min. This is almost exactly what the ZTF API query took on a fast ZTF API day from the IRSA SP.
I also ran it on some of the other literature coords to make sure it worked outside of our nominal sample and it did fine.

Other than the requirements update, looks good and can be merged when you are ready.

out of curiosity, you made 'service' a variable in ZTF_get_lightcurve . Does that mean there are other services?

troyraen:
Thanks @jkrick. This is in a temporary S3 bucket that is available automatically from the SMCE but requires user credentials for access from IRSA SP or elsewhere. I added an explanatory comment in the notebook.

Re: the service -- I'm not aware of another service. I made this a variable in case the call to `pyvo.dal.TAPService` is making an external call to the archive to create a client or something. No need to re-do such a call for every coord; that could negatively impact performance so better to do it once and send it in as a variable. I didn't track down whether it actually does make an external call, was just easier to be safe.

PR #133, opened 2023-08-16
Heasarc speedup

jkrick:
Following this sciserver [tutorial](https://github.com/HEASARC/sciserver_cookbooks/blob/main/source_list_querying.ipynb), and with much help from @zoghbi-a, I have changed the `HEASARC_get_lightcurves` function.  The old method was a for loop over an `astroquery.heasarc` search for each individual object.  The new method uses a pyvo tap query which cross matches a catalog of our targets with each HEASARC catalog of interest. 

This is a much faster method when running at scale, and probably preferred by HEASARC rather than sending them 500K individual requests. 

I do not know if this method will translate well to other archive searches:
1) Other archives may not allow table uploads to their tap queries (ZTF??).
2) It may not scale as well with much larger catalogs.  The HEASARC GRB catalogs we are querying are only a few hundred to a few thousand lines long, and therefore not representative of other light curve datasets.
3) other archives may require a login (Gaia??)


light_curves/multiband_lc.ipynb
```diff
@@ -195,13 +197,24 @@
...
+    "max_fermi_error_radius = str(1.0)  \n",
+    "max_sax_error_radius = str(3.0)\n",
+    "\n",
+    "#list of missions to query and their corresponding error radii\n",
+    "heasarc_cat = [\"FERMIGTRIG\", \"SAXGRBMGRB\"]\n",
+    "error_radius = [max_fermi_error_radius , max_sax_error_radius]\n",
+    "\n",
+    "#first make a VOTable from our master list of coordinates\n",
+    "#we need a VOTable as input to the pyvo TAP query \n",
+    "xmlfilename = make_VOTable(coords_list, labels_list,'sources.xml')\n",
```
troyraen, line 212: I wonder if the `HEASARC_get_lightcurves` function should handle this under the hood, so the user doesn't really have to know about it?

light_curves/code/heasarc_functions.py
```diff
@@ -1,12 +1,140 @@
...
+    #zoom in on the range of interest
+    #error radii are in units of degrees
+    plt.hist(heasarcresulttable["error_radius"], bins = 30, range = [0, 10])
+    
+    #in case anyone wants to look further at the data
+    return heasarcresulttable
+    
+#example calling sequences
+#resulttable = make_hist_error_radii('FERMIGTRIG')
+#resulttable = make_hist_error_radii('SAXGRBMGRB')
```
troyraen, line 51: Can these go in the docstring for this function? I would put it between the first sentence and the parameters

troyraen (commented):
Thanks, I'll look into whether this can be applied to the ZTF call.

Here's a couple small, random thoughts/comments I had while reading through the changes.

light_curves/code/heasarc_functions.py
jkrick: done

light_curves/multiband_lc.ipynb
jkrick: good point.  done.

light_curves/code/heasarc_functions.py
```diff
@@ -1,12 +1,148 @@
...
+        print('working on mission', heasarc_cat[m])
+
+        fermiquery=f"""
+            SELECT cat.name, cat.ra, cat.dec, cat.error_radius, cat.time,  mt.ID, mt.name
+            FROM {heasarc_cat[m]} cat, tap_upload.mytable mt
+            WHERE
+            cat.error_radius < {max_error_radius[m]} AND
+            CONTAINS(POINT('ICRS',mt.ra,mt.dec),CIRCLE('ICRS',cat.ra,cat.dec,cat.error_radius))=1
+             """
+        fermiresult = heasarc_tap.service.run_sync(fermiquery, uploads={'mytable': xmlfilename})
```
troyraen, line 127: `xmlfilename` is now an Astropy table (might want to rename the variable)

light_curves/code/heasarc_functions.py
```diff
@@ -1,12 +1,148 @@
...
+    coords_list : list of astropy skycoords
+        the coordinates of the targets for which a user wants light curves
+    labels_list: list of strings
+        journal articles associated with the target coordinates
+    sourcefilename: str
+        name of the output file; must be .xml
+    """
+    
+    tab = Table({
+        'name': labels_list,
```
troyraen, line 68:
```suggestion
        'name': [label.encode() for label in labels_list],  # encode strings for TAP
```

light_curves/code/heasarc_functions.py
```diff
@@ -1,12 +1,148 @@
...
+        'ra': [coord.ra for objectid, coord in coords_list],
+        'dec': [coord.dec for objectid, coord in coords_list],
+        'ID': [objectid for objectid, coord in coords_list]
+    })
+    #write out to an .xml file
+    tab.write(sourcefilename, format='votable', overwrite=True)
+    
+    return sourcefilename
```
troyraen, line 76:
```suggestion
    return tab
```

troyraen (commented):
In applying this table upload strategy to ZTF, I learned that you can send an Astropy table directly with the TAP query rather than having to send a file name. This would avoid having to write a temporary file to disk, which would be nice. 
Here's my [ZTF example](https://github.com/fornax-navo/fornax-demo-notebooks/blob/c2d1f8cad65808b8267300e81261e7e56f3b6365/light_curves/code/ztf_functions.py#L123).
I made some code suggestions below that are, I think, the minimum changes that would be needed to do this here.

light_curves/multiband_lc.ipynb
```diff
@@ -95,7 +97,7 @@
     "    get_green_sample, get_lyu_sample, get_lopeznavas_sample, get_hon_sample, get_yang_sample,  get_SDSS_sample, clean_sample\n",
     "from ztf_functions import ZTF_get_lightcurve\n",
     "from data_structures import MultiIndexDFObject\n",
-    "from heasarc_functions import HEASARC_get_lightcurves\n",
+    "from heasarc_functions import HEASARC_get_lightcurves, make_VOTable\n",
```
zoghbi-a, line 100: `make_VOTable` import is not needed in the notebook

light_curves/code/heasarc_functions.py
jkrick: done

light_curves/code/heasarc_functions.py
jkrick: done

light_curves/code/heasarc_functions.py
```diff
@@ -1,12 +1,148 @@
...
+        generated by `make_VOTable` functiom
+        
+    Returns
+    -------
+    df_lc : MultiIndexDFObject
+        the main data structure to store all light curves
+    """
+    #first make a VOTable from our master list of coordinates
+    #we need a VOTable as input to the pyvo TAP query 
+    xmlfilename = make_VOTable(coords_list, labels_list,xmlfilename)
```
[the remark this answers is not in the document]
jkrick: done

light_curves/code/heasarc_functions.py
```diff
@@ -1,12 +1,148 @@
...
+import pyvo
+import matplotlib.pyplot as plt
+import numpy as np
 
 from data_structures import MultiIndexDFObject
 
+#need to know the distribution of error radii for the catalogs of interest
+#this will inform the ligh curve query, as we are not intertsted in 
+#error radii which are 'too large' so we need a way of defining what that is
+def make_hist_error_radii(missioncat):
```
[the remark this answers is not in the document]
jkrick: I kept it in there but added a comment about why it is there (to help people decide on what the right threshold is for their science case).

light_curves/multiband_lc.ipynb
jkrick: removed it.

jkrick:
Thanks for the comments, I cleaned up the VOTable -> astropy table.  I agree it is much cleaner to not write output.

light_curves/multiband_lc.ipynb
```diff
@@ -473,7 +494,7 @@
     "        Gaia_get_lightcurve, (coords_list, labels_list, 1), callback=callback\n",
     "    )\n",
     "    pool.apply_async(\n",
-    "        HEASARC_get_lightcurves, (coords_list, labels_list, heasarc_radius, mission_list), callback=callback\n",
+    "        HEASARC_get_lightcurves, (coords_list, labels_list, heasarc_cat, error_radius, 'sources.xml'), callback=callback\n",
```
troyraen, line 497: `'sources.xml'` needs to be removed

troyraen (approved):
There's one missing import in heasarc_functions.py and the file name should be removed from the function call in the multiprocessing cell. I made a few other suggestions as well. Otherwise, looks good.

light_curves/code/heasarc_functions.py
```diff
@@ -1,12 +1,126 @@
...
+        #really just need to mark this spot with a vertical line in the plot, it's not actually a light curve
+        #so making up a flux and an error, but the time stamp and mission are the real variables we want to keep
+        df_fermi = pd.DataFrame(dict(flux=np.full(len(fermiresulttable),0.1), err=np.full(len(fermiresulttable),0.1), time=fermiresulttable['time'], objectid = fermiresulttable['id'], band=np.full(len(fermiresulttable),'Fermi GRB'), label=fermiresulttable['name2'])).set_index(["objectid", "label", "band", "time"])
+
+        # Append to existing MultiIndex light curve object
+        df_lc.append(df_fermi)
+    
+    return df_lc
+
+def HEASARC_get_lightcurves_forloop(coords_list,labels_list,radius, mission_list ):
```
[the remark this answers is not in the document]
jkrick: yes

light_curves/multiband_lc.ipynb
```diff
@@ -82,6 +82,8 @@
     "from astropy.table import Table\n",
     "from scipy import stats\n",
     "import pandas as pd\n",
+    "import pyvo\n",
+    "from tqdm import tqdm\n",
```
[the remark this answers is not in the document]
jkrick: done

light_curves/multiband_lc.ipynb
```diff
@@ -95,7 +97,7 @@
     "    get_green_sample, get_lyu_sample, get_lopeznavas_sample, get_hon_sample, get_yang_sample,  get_SDSS_sample, clean_sample\n",
     "from ztf_functions import ZTF_get_lightcurve\n",
     "from data_structures import MultiIndexDFObject\n",
-    "from heasarc_functions import HEASARC_get_lightcurves\n",
+    "from heasarc_functions import HEASARC_get_lightcurves#, make_coordsTable\n",
```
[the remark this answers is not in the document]
jkrick: done

light_curves/multiband_lc.ipynb
jkrick: done

PR #134, opened 2023-08-28
Speed up ZTF

### LINKED ISSUE #107: speedup ZTF light curve search

troyraen:
Closes #107 

This PR makes two improvements: 

1. switches the TAP query to use a table upload, following #133. Results of initial time tests are very good:

```
Nsample    query_time_original     query_time_with_table_upload
5                     12 sec                        2 sec
10                    30                            4.3
25                    61                            2.2
50                   160                            4.7
1,000              48 min                         3 min
```

2. Adds an option to load data from the parquet files in parallel. This parallelization is separate from the parallelization cell in the main notebook. Unfortunately the two parallelizations are mutually exclusive. For a large number of objects, it will be faster to call ZTF individually and take advantage of loading multiple files in parallel.

Here are a few load times and projections:
- 30 objects with no parallelization take a couple of minutes
- 1,000 objects with no parallelization takes about an hour
- 10,000 objects with `nworkers=8` (load 8 files in parallel) takes about an hour
- 500,000 objects with `nworkers=8`: I project 4.5 hours, but haven't done the full test


troyraen:
I believe this is ready to merge

PR #135, opened 2023-09-07
Gaia speedup

jkrick:
This PR changes the `Gaia_get_lightcurve` function to now use astroquery table upload with `launch_job_async` instead of a for loop over each individual object using `cone_search_async`.  The largest change is then in the `Gaia_retrieve_catalog` function which replaces `Gaia_retrieve_median_photometry`.  The other changes are mostly to allow for the different input/output of the astroquery search vs. the original search.  

Additionally, the code used to store the median photometry if there was not a time-series available.  Since I don't understand what this is (ie., if there is no time-series, how is there a median?), and in general we are not tracking individual photometry points, I have removed this part of the code.  I'm open to discussion of going back to tracking these values.

Lastly, I touched the plotting functions as they gave me trouble in the 'Kepler, TESS, K2' section, and I had to make small changes there to get the code to plot.  I have added this as issue #136  to make sure this section is functional going forward.

The 30 object Yang sample went from taking 1min19s to 2s.  
Testing with 10,000 sources took 10s.


light_curves/multiband_lc.ipynb
```diff
@@ -546,7 +537,7 @@
    "outputs": [],
    "source": [
     "# Save the data for future use with ML notebook\n",
-    "#parquet_savename = 'data/df_lc_051123.parquet.gzip'\n",
+    "#parquet_savename = 'data/df_lc_090723_yang.parquet.gzip'\n",
     "#parallel_df_lc.data.to_parquet(parquet_savename)\n",
```
troyraen, line 541: Why does this file name end with ".gzip", does anyone know? I see two issues with this:

1. This file is not gzip-ed -- it's just a regular Parquet file. Thus it's recommended that the name end with ".parquet" because some generic readers can use this to automatically determine the file format.
2. The content in Parquet files is generally compressed on a per-column basis (which is different than compressing the entire file together). gzip is an option for this type of compression, but it's not what's being used here. The `to_parquet` call is using the default compression, which is snappy not gzip.

light_curves/multiband_lc.ipynb
```diff
@@ -137,14 +137,14 @@
...
-    "#get_yang_sample(coords, labels)   #2018ApJ...862..109Y\n",
+    "get_yang_sample(coords, labels)   #2018ApJ...862..109Y\n",
     "\n",
     "#now get some \"normal\" QSOs for use in the classifier\n",
     "#there are ~500K of these, so choose the number based on\n",
     "#a balance between speed of running the light curves and whatever \n",
     "#the ML algorithms would like to have\n",
     "\n",
-    "num_normal_QSO = 50\n",
+    "num_normal_QSO = 0\n",
```
troyraen, line 147: Seems more natural to comment this out rather than set it to 0. 

Related question: What sample do we want to make the default in this cell? It would be good to pick one and then leave it.

light_curves/code/gaia_functions.py
```diff
@@ -387,41 +296,11 @@ def Gaia_mk_MultiIndex(coords_list, labels_list, gaia_phot , gaia_epoch_phot , v
...
-
-                    # add to table
-                    try:
-                        this_df_lc
-                    except NameError:
-                        #this_df_lc doesn't exist (yet)
-                        this_df_lc = dfsingle.copy()
-                    else:
-                        #this_df_lc_gaia exists
-                        this_df_lc = pd.concat([this_df_lc, dfsingle])
```
troyraen, line 423: If there is no match to any object in `coords_list`, `this_df_lc` never gets defined, and the `return` statement below throws an error. I suspect this is a problem with multiple archive calls, not just this one. One solution would be to return an empty dataframe, but that seems like a hassle. I wonder if a better solution will be for the archive call to return `None` in that case, and then to update the `MultiIndexDFObject.append` method to ignore `None` values.

This problem does not need to hold up this PR, but perhaps we should open an Issue that points to it.

troyraen (approved):
Looks good and runs for me on the SMCE. I commented below on a couple minor things, and also one larger issue that we should address but that doesn't need to hold up the PR. Finally, there is a comment in the Gaia cell of the main notebook that seems like it should be removed now:
```
## Note: This is very slow. Can probably make faster with direct SQL search?
```

light_curves/multiband_lc.ipynb
jkrick: done

light_curves/code/gaia_functions.py
```diff
@@ -10,6 +10,7 @@
 from tqdm import tqdm
```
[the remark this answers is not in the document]
jkrick: done

light_curves/code/gaia_functions.py
```diff
@@ -477,3 +357,55 @@ def Gaia_plot_lightcurves(df_lc , nbr_objects):
     plt.show()
     
     return(True)
+def Gaia_extract_median_photometry(gaia_table):
```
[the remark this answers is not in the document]
jkrick: @afaisst are you ok with the plan to remove the median photometry from the light curve plots?

light_curves/multiband_lc.ipynb
jkrick: I guess default so far has been Yang et al., sample.  I will change to be that

light_curves/code/gaia_functions.py
jkrick: good point, I will open an issue to maybe make this more uniform across all functions.  Heasarc currently returns an empty data frame, so does TESS/Kepler/K2.  Those are the ones that routinely will not find any matches for a small sample.  I'm going to tag you @troyraen in that issue because I don't know how to 'ignore' `none` values

light_curves/code/gaia_functions.py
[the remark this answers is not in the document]
jkrick: After discussion with Andreas, the code now keeps the median photometry function in case a future user wants to use it.  Possible uses are to compare with similar bands of other photometry, combining them into a light curve.  This sort of cross calibration is beyond the scope of this demo notebook, but we will leave the function there in case someone else wants to do it.

PR #139, opened 2023-09-12
Move root-level `code` and `data` directories under `forced_photometry`

### LINKED ISSUE #106: Re-evaluate directory structure for code and data files
Currently, the data and code used by the forced_photometry notebook is stored in the corresponding subdirectories of the repo's _root_ directory while the light_curves notebook stores in subdirectories of the light_curves directory. Now that we have two examples, we could re-evaluate how we're organizing those files.

troyraen:
Closes #106 

Moves root-level `code` & `data` directories under the `forced_photometry` directory.


jkrick (approved):
It took blowing away my fornax-demo-notebooks directory, re-cloning from GitHub, and restarting the kernel.  But now the code works on the SMCE and can run in its entirety without adding images and with all plots populated with points.  I ran it with a 4.0 arcmin radius which gives more points for that second to last plot.  It's possible that a small radius would produce an empty plot.

bsipocz (approved):
Looks good, thanks. 

One minor comment, which we may or may not want to do, and is anyway beyond the scope of this PR:
- use the `data` directory only for input data, downloads and raw stuff
- save all outputs into a separate one, to avoid mixing of what we produce ourselves and bring in from the outside

troyraen:
> * use the `data` directory only for input data, downloads and raw stuff
> * save all outputs into a separate one, to avoid mixing of what we produce ourselves and bring in from the outside

Good idea. I just spoke with @jkrick and we decided to include it in this PR. You (@bsipocz) also gave a suggestion previously to rename the `code` directly to a word not protected by python. So here's my plan to finish this PR:

- [ ] rename the directory `code` -> `code_` (We decided to keep the word "code" because that's what best describes what's in the directory. I think that adding the `_` is a common way to avoid the issue of protected words.)
- [ ] update anything that saves output so that it saves to a new directory called `output`
- [ ] leave the input data, etc. in the `data` directory

I will make these same changes in #140 for the light curve use case so that the repo is consistent.

troyraen:
@bsipocz @jkrick This repo has a [tests](https://github.com/fornax-navo/fornax-demo-notebooks/tree/main/tests) directory that only has one thing in it -- the test I wrote to make sure I hadn't broken anything in the forced photometry notebook when I reconfigured the main function parameters (#49). Since this test does not ever need to be run again (it was very specific to that PR), I'm tempted to just delete it to avoid confusion and reduce clutter. I probably shouldn't have committed it in the first place. Thoughts?

bsipocz:
>  I'm tempted to just delete it to avoid confusion and reduce clutter.

sounds reasonable. thanks!

jkrick:
I agree that removing tests directory makes sense.

bsipocz:
I'll think a bit more about what to call that `code` to see there is anything better than the `_`, but today is just swamped with meetings which is not great for any creativity that would be needed for naming things.

troyraen:
I went with `code_` for the directory but happy to change it later if someone thinks of a better name.

bsipocz:
Maybe something a bit more descriptive than the `_` would be `code_src`?

PR #140, opened 2023-09-13
Clean up imports and file organization

### LINKED ISSUE #96: Fix imports in light curve notebook

troyraen:
Closes #96

A lot of the clean up for #96 was done in #99 and #110. This PR does the rest.

- [x] remove unused imports from notebook and code files. sort remaining imports. 
- [x] remove unused installs from requirements.txt
- [x] rename the directory `code` -> `code_`
- [x] save downloaded data to a `data` directory. add this dir to .gitignore
- [x] save results to an `output` directory


.gitignore
```diff
@@ -1,7 +1,10 @@
-# directory created by notebooks to store results
+# Directories created by notebooks to store data, results, etc.
+# "data" dir to store any data that is downloaded or otherwise needs to be made available to the notebook
+# occasionally, data may need to be checked into the repo (e.g., forced_photometry/data), but this should be rare
+data/
```
troyraen, line 4: This will not affect the files in forced_photometry/data that are already checked into the repo, but it will prevent any new files in a "data" directory from being accidentally committed.

light_curves/multiband_lc.ipynb
```diff
@@ -74,34 +74,43 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# create a \"data\" directory to store any data that needs to be downloaded\n",
+    "Path(\"data\").mkdir(exist_ok=True)\n",
+    "\n",
+    "# create an \"output\" directory to store any results that we want to persist\n",
+    "Path(\"output\").mkdir(exist_ok=True)"
```
troyraen, line 113: I added this cell at the top of the notebook to create the "data" and "output" directories up front. However, I wonder if it would be better to check those two (empty) directories into the repo to avoid doing this in the notebook? I would still leave those two directories in the .gitignore file so that nothing gets accidentally committed to them. This would reduce clutter in the notebook and simplify the workflow (perhaps especially helpful if/when we start porting code out of the notebook to run large jobs). I'm thinking about this in the context of all future notebooks, not just this one. It would be nice to use the same process for all of them, and make it as simple as possible.

jkrick: @bsipocz what is your opinion on this?  I see both of Troy's points.  1) declutter so not have it in the notebook but 2) make sure the user sees the organizational tools that we have setup, so leave in the notebook.  I lean slightly towards not having it in the notebook since the SP has a nice file structure list on the left column so the directories are quite transparent.

jkrick (approved):
thanks for doing this.  Looks good to me, and runs on the SMCE as expected.

troyraen:
We had a group discussion offline and decided on a few things. Capturing what I need to do here:

- [ ] change the dir name code_ -> code_src (do for both light_curves and forced_photometry)
- [ ] create empty directories for both data and output that live in the repo. make sure they are listed in .gitignore so nothing gets accidentally committed to them. remove the various calls in notebooks and code files that create those directories.
- [ ] resolve the conflicts with the main branch

bsipocz (approved):
These all look to me, and can be merged once items from the checklist are done (https://github.com/fornax-navo/fornax-demo-notebooks/pull/140#issuecomment-1738124102)

PR #144, opened 2023-09-15
bugfix heasarc_get_lightcurves

jkrick:
- changed column names in the ADQL query which were accidentally out of sync with  `make_coordsTable()`
- renamed the query to not include the string 'fermi' since it is more generic now


troyraen (approved):
LGTM. Runs for me without error.

PR #145, opened 2023-09-21
Add `tqdm` to WISE and ZTF calls and do a bugfix for the parallel processing cell

troyraen:
Adds `tqdm` progress bars for WISE and ZTF and fixes a bug in certain types of calls to ZTF. Normally I would put these two things in different PRs but I'm trying to get this out for folks who are waiting to use it and I don't want the complication of two PRs.

## `tqdm`

Adds `tqdm` progress bars to the WISE and ZTF calls. Note that neither of these calls iterates directly over the objects in `coords_list`, so the number of iterations will not match the length of that list. Here is what they do iterate over:

**WISE**:
1. catalog partitions (only the subset of partitions that contain objects in `coords_list`)

**ZTF** has two steps that take significant time (relative to total time). A progress bar is shown for each:
1. chunks of `coords_list`, with 10,000 objects per chunk. This is for the TAP call that figures out which file each object is in.
2. files in the dataset (only the subset of files that contain objects in `coords_list`). This is the step that actually loads the light curves.

Finally, note that for both WISE and ZTF, there are short steps that occur both before and after the long-running steps that display progress bars. Those steps are short but still take a non-zero amount of time, and this time may increase with the number objects in `coords_list`. So be aware that, (1) the progress bars may not display immediately after the calls are made; and (2) the calls will still need some time to complete after their process bars hit 100%.

## Bugfix

Switches the multiprocessing to "spawn" new process by adding this to the notebook:

```python
import multiprocessing as mp

# when using multiprocessing, "spawn" new processes because it uses less memory and is thread safe
# in particular, this is required for pd.read_parquet (used by ZTF_get_lightcurve)
```

The bug only affected the parallel processing cell in cases where the serial call to ZTF had not been run first. Calls that use the ZTF function's internal parallelization already had this implemented.


troyraen:
@xoubish please look this over and let me know if anything is confusing.

For `tqdm`, I suggest running the WISE and ZTF calls individually in the "serial" section of the notebook so that you know what to expect for each before using the code in the "parallel" section of the notebook.

The bug that this PR fixes is probably something you were running into. If you are using the code from the parallel processing cell in a separate script, make sure to add the fix to your script.

PR #146, opened 2023-09-26
Samplepy

xoubish:
This has the ML (dimensionality reduction) notebook with texts added.


light_curves/sample_lc.py
troyraen, line 1: I know that we need to work out a better solution for fetching the light curves of a large number of objects than trying to run the parallel processing cell in the main notebook. However, I am concerned about adding this file because it duplicates the code in the main notebook, and thus will need to be updated manually with the same changes if the notebook gets changed. Ideally, we should come up with a solution that does not require this type of manual upkeep. If this file is intended to be a temporary solution then I am ok with it, but recommend that it live under the code directory.

troyraen (commented):
I spent a few hours trying to understand what is here and why and I left some related comments and suggestions below. I haven't gotten to looking at the notebook and related code in any detail.

- I strongly recommend that we do not commit anything in the data directory. The data in this PR is almost 0.5GB and increases the size of this repo by more than 2.5 times. Bloating the size of the repo makes it especially difficult to work with in an environment like the SMCE where user space is limited. If needed, I can help figure out what to do instead.
- The notebook output should be cleared/removed and not committed to the repo.
- `sompy` is an external package, so we should install it along with the other requirements instead of putting a copy of it in this repo. It appears that it is not published to PyPI, but I think we should still be able to pip install it using `pip install git+https://github.com/sevamoo/SOMPY`.

light_curves/code/WISE_MIR_variable_AGN_with_PS1_photometry_and_SDSS_redshift.csv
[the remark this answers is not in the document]
xoubish: This file is an input that we got from an internal collaborator. Not sure what the best way forward is, either we should get permission put it in some cloud where we can retrieve in the sp or I have to find a somewhat similar sample somewhere else.

light_curves/sample_lc.py
xoubish: I agree. I will copy it under code. Also, it still does not add wise in the loop I think.

xoubish:
> * The notebook output should be cleared/removed and not committed to the repo.

The reason I ran it was to be able to showcase the plots to people who won't have access to the data that is large ...

light_curves/code/WISE_MIR_variable_AGN_with_PS1_photometry_and_SDSS_redshift.csv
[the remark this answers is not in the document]
xoubish: maybe I add .csv to .gitignore file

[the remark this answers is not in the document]
troyraen: What is this file needed for? I'm guessing it's the redshifts. Anything else?

Because this file is for a specific sample of objects, I don't think the notebook should rely on it directly because then the notebook is tied to a specific sample. It's totally reasonable to use a file like this while developing, but I think it should just be used locally. I don't think it's something that we should try to make publicly available in the repo, S3, or elsewhere. I know that we need to work out a general way to retrieve redshifts (or try to) for any sample. I have some thoughts, but I'll leave that for another conversation.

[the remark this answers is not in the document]
xoubish: This is a specific sample which showed variability in WISE, we read the coordinates as part of the sample. As the idea of the notebook is to see if we can separate different classes of objects, this file is useful to show trends.

light_curves/sample_lc.py
troyraen: > it still does not add wise in the loop I think.

Let me know if you want my help with this. Also let me know if you think that the code that does this in the main notebook doesn't work either... as far as I know, the notebook works as intended.

jkrick (commented):
It's so fast! and so exciting to run my own Umaps.

Sorry, I feel like many of my comments were covered by Troy, so apologies for the overlap.  Mostly about file naming and locations.

I haven't looked deeply at the ML_AGNzoo.ipynb, but just checked to see that it will run in its entirety.  I feel like that is what is necessary for v1.  I only found one KeyError, but the whole code does run without that cell.

The last thing I would suggest is to follow the style guide of the other notebooks (ie., numbering sections, etc.).  https://github.com/spacetelescope/style-guides/blob/master/guides/jupyter-notebooks.md

light_curves/code/ztf_functions_old.py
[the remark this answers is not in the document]
xoubish: removed :D

light_curves/data/20210126_PS-IC40-IC86_VII (1).zip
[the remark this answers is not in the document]
xoubish: no not needed.

light_curves/ML_AGNzoo.ipynb
```diff
@@ -0,0 +1,1049 @@
...
+    "\n",
+    "for b in objid:\n",
+    "    singleobj = parallel_df_lc.data.loc[b,:,:,:]  \n",
+    "    label = singleobj.index.unique('label')\n",
+    "    bband = singleobj.index.unique('band')\n",
+    "    for bb in bband:\n",
+    "        bands = singleobj.loc[label[0],bb,:].index.get_level_values('time')[:]\n",
+    "        #bands.values\n",
+    "        #print(bb,len(bands[:]),np.round(bands[:].max()-bands[:].min(),1))    \n",
+    "        cadence[bb].append(len(bands[:]))\n",
```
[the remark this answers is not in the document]
xoubish: I can not reproduce the error, and the labels are read from the data frame I think, so the change should be ok.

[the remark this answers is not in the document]
xoubish: My experience with these key errors has been pandas/python version mismatch on different platforms and servers (a guess).

light_curves/sample_lc.py
xoubish: I think it would be great if you could take a look. I tried to run even for a very small sample (~100) and the code doesn't finish.

light_curves/code/sample_lc.py
```diff
@@ -0,0 +1,145 @@
+import sys
+import time
+import warnings
+from math import ceil
+from multiprocessing import Pool
```
troyraen, line 5:
```suggestion
import multiprocessing as mp
```

light_curves/code/sample_lc.py
```diff
@@ -0,0 +1,145 @@
...
+    wise_radius = 1.0 * u.arcsec
+    panstarrs_radius = 1.0 / 3600.0  # search radius = 1 arcsec
+    lk_radius = 1.0  # arcseconds
+    hcv_radius = 1.0 / 3600.0  # radius = 1 arcsec
+    n_workers = 8  # this should equal the total number of archives called
+    ztf_nworkers = None
+
+    parallel_starttime = time.time()
+
+    # start a multiprocessing pool and run all the archive queries
```
troyraen, line 97:
```suggestion
    # start a multiprocessing pool and run all the archive queries
    # "spawn" new processes because it uses less memory and is thread safe
    # in particular, this is required for pd.read_parquet (used by ZTF_get_lightcurve)
    # https://stackoverflow.com/questions/64095876/multiprocessing-fork-vs-spawn
    mp.set_start_method("spawn", force=True)
```

light_curves/code/sample_lc.py
```diff
@@ -0,0 +1,145 @@
...
+            Gaia_get_lightcurve, (coords_list,  labels_list , 1), callback=callback
+        )
+        pool.apply_async(
+            HEASARC_get_lightcurves, (coords_list, labels_list, heasarc_cat, error_radius), callback=callback
+        )
+        pool.apply_async(
+            HCV_get_lightcurves, (coords_list, labels_list, hcv_radius), callback=callback
+        )
+        pool.apply_async(
+            icecube_get_lightcurve, (coords_list, labels_list, 3, "./data/", 1), callback=callback
```
troyraen, line 113:
```suggestion
            icecube_get_lightcurve, (coords_list, labels_list, 3, "../data/", 1), callback=callback
```

light_curves/code/sample_lc.py
```diff
@@ -0,0 +1,145 @@
+import sys
```
troyraen, line 1:
```suggestion
import os
import sys
```

light_curves/code/sample_lc.py
```diff
@@ -0,0 +1,145 @@
...
+from sample_selection import get_lamassa_sample, get_macleod16_sample, get_ruan_sample, get_macleod19_sample, get_sheng_sample, get_green_sample, get_lyu_sample, get_lopeznavas_sample, get_hon_sample, get_yang_sample,get_SDSS_sample, get_paper_sample, clean_sample,nonunique_sample,TDE_id2coord
+from data_structures import MultiIndexDFObject
+from heasarc_functions import HEASARC_get_lightcurves
+from TESS_Kepler_functions import TESS_Kepler_get_lightcurves
+from WISE_functions import WISE_get_lightcurves
+from ztf_functions import ZTF_get_lightcurve
+from ML_utils import unify_lc, stat_bands, autopct_format, combine_bands,\
+mean_fractional_variation, normalize_mean_objects, normalize_max_objects, \
+normalize_clipmax_objects, shuffle_datalabel, dtw_distance, stretch_small_values_arctan
+
```
troyraen, line 26:
```suggestion


# make sure the data and output directories exist so that we can write to them
os.makedirs("../data", exist_ok=True)
os.makedirs("../output", exist_ok=True)

```

light_curves/code/sample_lc.py
```diff
@@ -0,0 +1,145 @@
...
+    get_macleod19_sample(coords, labels)  #2019ApJ...874....8M
+    get_sheng_sample(coords, labels)  #2020ApJ...889...46S
+    get_green_sample(coords, labels)  #2022ApJ...933..180G
+    get_lyu_sample(coords, labels)  #z32022ApJ...927..227L
+    get_lopeznavas_sample(coords, labels)  #2022MNRAS.513L..57L
+    get_hon_sample(coords, labels)  #2022MNRAS.511...54H
+    get_yang_sample(coords, labels)   #2018ApJ...862..109Y
+
+    # Variable AGN sample from Ranga/Andreas:
+    VAGN = pd.read_csv('data/WISE_MIR_variable_AGN_with_PS1_photometry_and_SDSS_redshift.csv')
```
troyraen, line 48:
```suggestion
    VAGN = pd.read_csv('../data/WISE_MIR_variable_AGN_with_PS1_photometry_and_SDSS_redshift.csv')
```

light_curves/code/sample_lc.py
```diff
@@ -0,0 +1,145 @@
...
+    get_paper_sample('2020ApJ...896...10B','Palomar variable 20',coords,labels)
+
+    #To remove duplicates from the list if combining multiple references clean_sample can be used 
+    # the call below with nonunique_sample just changes the structure to mimic the output of clean sample
+    coords_list, labels_list = nonunique_sample(coords, labels) 
+    print('final sample: ',len(coords))
+    return coords_list,labels_list
+ 
+
+def parallel_lc(coords_list,labels_list,parquet_savename = 'data/df_lc_.parquet.gzip'):
```
troyraen, line 78:
```suggestion
def parallel_lc(coords_list,labels_list,parquet_savename = '../output/df_lc_.parquet'):
```

light_curves/code/sample_lc.py
```diff
@@ -0,0 +1,145 @@
...
+    print('parallel processing took', parallel_endtime - parallel_starttime, 's')
+    
+    # # Save the data for future use with ML notebook
+    parallel_df_lc.data.to_parquet(parquet_savename)
+    print("file saved!")
+    return parallel_df_lc
+
+def main():
+    c,l = build_sample()
+    dflc = parallel_lc(c,l,parquet_savename = 'data/df_lc_smalltest.parquet.gzip')
```
troyraen, line 141:
```suggestion
    dflc = parallel_lc(c,l,parquet_savename = '../output/df_lc_smalltest.parquet')
```

troyraen (commented):
This is missing the bug fix from #145. Without it, the ZTF function will hang. In addition, the paths should be updated as follows:
- since this file moved to the code directory, paths to the data directory need to be pointed up one level
- we are switching to a workflow that puts data that we produce in a new "output" directory to avoid confusion with data that we retrieve from elsewhere, so please update your paths accordingly. (hopefully this process will be more streamlined after #140 gets merged)
- I recommend that the parquet file names end in ".parquet" instead of ".parquet.gzip". This is not strictly necessary here but may help avoid bugs and errors in the future. I can explain more if you'd like.

I have made the specific code suggestions below. If you want to accept the suggestions, you will see buttons that let you do this (either individually or altogether in a batch). If you use them, the changes will be committed directly to your branch, and then you don't have to make them separately yourself. Just be aware that if you do this you will need to do a `git pull` from your local machine (or wherever you're working on the code) before you make any other changes in order to avoid dealing with conflicts.

light_curves/sample_lc.py
troyraen: You are missing the bug fix from #145. In addition, the paths need to be updated because this file was moved. See my comments and suggestions below.

PR #147, opened 2023-09-26
Heasarc label fix

jkrick:
This tiny PR fixed a bug in the HEASARC functions which had the label for all HEASARC catalogs fixed to a single string.  It now will accurately reflect the catalog name from which the sources was obtained.  At the same time I added some print statements to let us know how many sources were found in the HEASARC catalogs.


troyraen (approved):
LGTM and runs on the SMCE.

PR #148, opened 2023-09-26
Writing out coords_list and labels_list to disk

bsipocz:


jkrick (approved):
thanks for doing this, looks good to me

bsipocz:
I've added the Table-zation of `clean_sample` and the rest, but likely won't get to it today.

PR #151, opened 2023-09-28
Light Curve Classifier V1

jkrick:
This is V1: A fully functional demo notebook which runs ~10 ML algorithms on light curve data as output from our light curve notebook to try to identify a classifier for CLAGN.  

There are many improvements I would like to make, but here is the starting point.


light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "scrolled": true
+   },
+   "outputs": [],
+   "source": [
+    "!pip install sktime\n",
+    "!pip install tqdm\n",
+    "!pip install googledrivedownloader"
```
xoubish, line 65: Does it make sense to add these to the requirements.txt? As the ML notebook requires packages such as acstools which is in the requirements anyways.

light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   "outputs": [],
+   "source": [
+    "#access structure of light curves made in the light curve notebook\n",
+    "# has CLAGN & SDSS small sample, all bands\n",
+    "# https://drive.google.com/file/d/1y92XoKuUa-auAWM39KpSaBC9VpEqXcC6/view?usp=share_link\n",
+    "\n",
+    "gdd.download_file_from_google_drive(file_id='1y92XoKuUa-auAWM39KpSaBC9VpEqXcC6',\n",
+    "                                    dest_path='./data/df_lc_458sample.parquet.gzip',\n",
+    "                                    unzip=True)\n",
+    "\n",
```
xoubish, line 114: If we run multibandlc notebook, the idea is to have the dataframe saved in output instead of data. Shouldn't we then also download this parquet into output?

light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "df_lc = pd.read_parquet(\"/home/jkrick/stagejkrick/JK_workingdir/df_lc_458sample.parquet.gzip\")\n"
+   ]
```
xoubish, line 116:
```suggestion
    "df_lc = pd.read_parquet(\"data/df_lc_458sample.parquet.gzip\")\n"
   ]
```

xoubish: or output?!

light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "\n",
+    "df_lc['label'] = df_lc.label.str.replace('MacLeod 16', 'CLAGN')\n",
+    "df_lc['label'] = df_lc.label.str.replace('LaMassa 15', 'CLAGN')\n",
+    "df_lc['label'] = df_lc.label.str.replace('Yang 18', 'CLAGN')\n",
+    "df_lc['label'] = df_lc.label.str.replace('Lyu 21', 'CLAGN')\n",
+    "df_lc['label'] = df_lc.label.str.replace('Hon 22', 'CLAGN')\n",
+    "df_lc['label'] = df_lc.label.str.replace('Sheng 20', 'CLAGN')\n",
+    "df_lc['label'] = df_lc.label.str.replace('MacLeod 19', 'CLAGN')\n",
+    "df_lc['label'] = df_lc.label.str.replace('Green 22', 'CLAGN')\n",
+    "df_lc['label'] = df_lc.label.str.replace('Lopez-Navas 22', 'CLAGN')\n",
```
xoubish, line 192: smarter than my for loop looking for the labels :D

light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,983 @@
...
+   "outputs": [],
+   "source": [
+    "#idea here is that we normalize across each object.  \n",
+    "#so the code will know for example that within one object W1 will be brighter than ZTF bands\n",
+    "#but from one object to the next, it will not know that one is brighter than the other."
+   ]
+  },
+  {
+   "cell_type": "raw",
+   "metadata": {},
```
xoubish, line 523: code

light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "from sktime.classification.deep_learning import CNNClassifier\n",
+    "from sktime.classification.ensemble import WeightedEnsembleClassifier\n",
+    "from sktime.classification.dictionary_based import IndividualTDE\n",
+    "from sktime.classification.distance_based import KNeighborsTimeSeriesClassifier\n",
+    "from sktime.classification.feature_based import RandomIntervalClassifier\n",
+    "from sktime.classification.feature_based import Catch22Classifier\n",
+    "from sktime.classification.shapelet_based import ShapeletTransformClassifier\n",
+    "from sktime.classification.dummy import DummyClassifier\n",
+    "\n",
+    "names = [\"Arsenal\",                     #kernel based\n",
```
xoubish, line 864: Would be cool to see one or two examples with plots separately (the ones which takes less time) and then leave the loop with all possible classifiers for the end of the notebook for enthusiasts who can wait

light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,983 @@
...
+    "pd.reset_option('all')"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pivoted_df_reindex"
```
xoubish, line 669: This still shows all normalized fluxes as 100 (the initialized value) for me. Not sure why they are not updated.

xoubish (approved):
All ran fine on SMCE, I didn't check for details.

light_curves/lc_classifier.ipynb
jkrick: yes, @troyraen @bsipocz @stargaser should we have a different requirements.txt file for each demo notebook in that light_curves directory?  Or all in one requirements.txt?

light_curves/lc_classifier.ipynb
jkrick: I think this parquet should be in data/ because it is required to run the lc_classifier notebook.  But this reminds me I also wanted to get rid of that .gzip

xoubish: The output of one notebook is the input to the other, so a bit confusing to me.

light_curves/lc_classifier.ipynb
jkrick: done.

light_curves/lc_classifier.ipynb
bsipocz: I think it would make sense to have different requirements.txt for each use case, so we could, if needed use different version pins, but not necessarily one for each notebook (e.g. I expect the full use case to be able to run along with each other).
If if we run into some version requirement conflict, we could still separate it out to separate files per notebook.

troyraen (approved):
Looks good. I appreciate the text and comments explaining what you're doing and why. Also the output and figures showing what the transformed data look like. It was clear and easy to follow. Fun to see the whole thing after hearing about bits and pieces for awhile.

I left a few suggestions below (also see the .gzip comment above), but no blockers. Merge when ready. This will cause a minor conflict with #140. I assume you're ready to merge fairly quickly, so I'll let you go first and then I'll deal with the conflicts.

One overall thing... I see what you mean (from previous conversations) about the hassle of setting and unsetting the index. After seeing the first light curve notebook and now this one, I think we (probably me) jumped the gun when we set the index from the beginning. It was great to carefully consider which fields/data we needed in the dataframe before starting, but we should have left them all as columns. I think as a general workflow, it's totally fine and good to set an index on the fly in spots where it's actually helpful. And this can be different indexes in different spots, no problem. This shouldn't be an expensive operation, especially compared to everything else we're doing. But, having tied ourselves to a particular index structure that is very rarely actually needed, we have to fight with it in a bunch of places where it's counterproductive. Even besides the `reset_index`s, the `.loc`s with 4 index levels are another thing that is unnecessarily complicated because of this structure. Good lessons for me especially. I'm happy to think and talk more about where to go from here, and to go back and change/fix it if we decide that's best.

light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,1029 @@
...
+    "#for each objct in the dataframe\n",
+    "for oid in tqdm(range(nobj)):\n",
+    "    try:\n",
+    "        oid_bandname = df_lc.loc[oid,:,:,:].index.unique('band').tolist()\n",
+    "    except KeyError:\n",
+    "        #some of the data cleaning above must have removed some objects\n",
+    "        print(oid, ' objectid doesnt exist')\n",
+    "    else:\n",
+    "        #this is the case where the objectid exists, so let's make sure all the bands are there\n",
+    "        oid_label = df_lc.loc[oid,:,:,:].index.unique('label').tolist()\n",
```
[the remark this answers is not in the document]
jkrick: issue #152

light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,1029 @@
...
+    "        objectid = df_reindex_reset.query(querystring).objectid.values\n",
+    "        band = df_reindex_reset.query(querystring).band.values\n",
+    "        label = df_reindex_reset.query(querystring).label.values\n",
+    "        dfsingle = pd.DataFrame(dict(flux=normflux, err=df_reindex.loc[oid,:,:,:].err, datetime=datetime, objectid=objectid, band=band,label=label))\n",
+    "\n",
+    "        #now concatenate the single new df with df_lc\n",
+    "        if oid == 0:\n",
+    "            df_norm = dfsingle\n",
+    "        else:\n",
+    "            df_norm = pd.concat([df_norm, dfsingle])\n",
```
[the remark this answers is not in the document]
jkrick: I changed the for loop and try/except statement to not have to worry about the objectid0 concern, but didn't work on the dfsingles and concat because this cell is likely going in the trash. I did the for loop just because maybe muscle memory will kick in on these loops and I will write them better on the first try one day (and I made a note to myself to check all of these in a future PR).   

I certainly need to fix the normalization, but will also save that for a different PR. My plan for the normalization is to write a new function which uses whatever Shooby used in her code since she has shown that she can correctly normalize such that her ML algorithm doesn't just find the different in brightness between the samples.

light_curves/lc_classifier.ipynb
```diff
@@ -0,0 +1,1029 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "#proove this did what I wanted it to do for one randomly chosen object\n",
+    "oid = 318\n",
+    "meanflux = df_norm.loc[oid,:,:,:].flux.mean(skipna = True)\n",
```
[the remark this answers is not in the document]
jkrick: I actually like keeping the extra `:, :, :`.  It seems more clear to me to keep them so that the structure of the thing is apparent to anyone trying to figure out what I did (or me at any time > 5 min after I wrote it).  Unless you are truly opposed, I'd like to keep them.

jkrick (commented):
Thanks for the comments everyone.  Mostly I have used them to populate new issues for future PRs or my own notes of what needs to be worked on to improve the code.  In other words, I have not made any large changes, but plan to do them in future work.

light_curves/lc_classifier.ipynb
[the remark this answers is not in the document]
troyraen: I'm not opposed and I agree that it's more clear especially in the context of this notebook where there are lots of other spots where `:` is required.

PR #154, opened 2023-10-02
BUG: Uniformly handle the case where there are no archive matches for coords_list

### LINKED ISSUE #138: uniformly handle the case where there are no archive matches for coords_list
Currently the Heasarc and TESS/Kepler/K2 get_lightcurves functions return empty data frames if no matches are found in the archives.  The gaia_get_lightcurves function will throw an error if this happens (good thing they have so much data!).  Other archive calls need to be checked and a uniform solution implemented.

Options appear to be
- return and empty data frame
- return `none` and edit `multiindexdf.append` to ignore `none`

troyraen:
Closes #138

Standardizes all `*_get_lightcurve` functions to return an empty dataframe if no matches were found for the objects in `coords_list`. The two functions actually changed here used to throw errors if they didn't find matches. All other functions already return an empty dataframe.

### DIFF
--- a/light_curves/code_src/gaia_functions.py
+++ b/light_curves/code_src/gaia_functions.py
@@ -44,6 +44,11 @@ def Gaia_get_lightcurve(coords_list, labels_list , verbose):
     
     print(gaia_table.columns)   
     
+    # if none of the objects were found, there's nothing to load and the Gaia_retrieve_EPOCH_PHOTOMETRY fnc
+    # will raise an HTTPError. just return an empty dataframe instead of proceeding
+    if len(gaia_table) == 0:
+        return MultiIndexDFObject()
+
     ## Extract Light curves ===============
     # For each of the objects, request the EPOCH_PHOTOMETRY from the Gaia DataLink Service
 
@@ -58,8 +63,7 @@ def Gaia_get_lightcurve(coords_list, labels_list , verbose):
                                     labels_list=labels_list, gaia_phot = gaia_table,
                                     gaia_epoch_phot=gaia_epoch_phot, verbose=verbose)
 
-    df_lc = MultiIndexDFObject()
-    df_lc.append(df_lc_gaia)
+    df_lc = MultiIndexDFObject(data=df_lc_gaia)
     
     
     return(df_lc)
--- a/light_curves/code_src/ztf_functions.py
+++ b/light_curves/code_src/ztf_functions.py
@@ -49,6 +49,11 @@ def ZTF_get_lightcurve(coords_list, labels_list, nworkers=6, ztf_radius=0.000278
     # use a TAP query to locate which files each object is in
     locations_df = locate_objects(coords_list, labels_list, ztf_radius)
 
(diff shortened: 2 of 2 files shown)


light_curves/code_src/data_structures.py
```diff
@@ -46,14 +46,20 @@ def append(self,x):
...
         Parameters
         ----------
-        x : Pandas dataframe
+        x : Pandas dataframe, optional
             contains columns [flux, fluxerr] and multi-index [objectid, label, band, time]
+            If x is None, calling this method has no affect (used when an archive call finds no light curve data).
         """
+        # if an archive call found no data, x is None. return without doing anything
+        if x is None:
+            return
```
bsipocz, line 55: It may or may not be applicable (and if it's, it's a more substantial fix than this current PR): over the years in astroquery, we converged to the solution of returning empty Table-s (or whatever the expected valid return type is) when there is no match, rather than None or an error, so to avoid all downstream code to deal with these types of conditionals.

troyraen: Ah great, thanks for the heads up. The functions here are handling this case in different ways (not all of them use astroquery, and even the ones that do transform the data before returning and so may still handle this differently) so the idea here is to standardize to _something_. The consensus from those who voiced opinions previously was to use `None`, but if you have a strong opinion in favor of the astroquery method I could instead make these return empty dataframes to match the astroquery workflow.

bsipocz: Nope, I don't have strong opinions, and since there aren't really downstream users here, the actual choice will matter less. It was just the experience that handling None/Error returns was more problematic at the end than returning empty containers. (E.g. it super easily distinguishes actual query errors/service outages/etc from the valid result of not finding matches, etc.)

troyraen: It was actually simpler to standardize to an empty dataframe so I just went with that. (More info in #138.)

bsipocz (approved):
LGTM

Minor note for workflow: you could maybe rebase and remove a commit rather than adding a direct revert into the history*?

* but the history of this repo is all over the place, so it doesn't really matter.

troyraen:
> you could maybe rebase and remove a commit rather than adding a direct revert into the history*?

Hm, I'd be happy to but I don't know how. I know there's different ways to "revert" but I didn't think it was possible to completely remove a commit from the history once it's been pushed, except to delete the branch and start over. I'll look into it.

bsipocz:
`git rebase -i 8a9d845f44d3ffefdace91ff8784bbc0e6bc5d7c` is the command, it brings up an editor, you remove the line for the commit that you want to delete/revert and save and exit.

More on slack

troyraen:
I did the rebase and deleted aeb1cf00064d79f4cab19a7a7e62251077d41b8e and 2118c5967fce84d653bf3d15d7ec441e8e313d14 (the initial commit that I want to remove plus my "revert" commit) but it seems like I just made the history worse rather than better?

bsipocz:
Yes, it always has to be a force-push, as the history has changed.

troyraen:
Ah, now I think it's right. Thanks for the lesson, that's useful to know.

PR #156, opened 2023-10-02
Use Table for passing along sample coordinates and labels

bsipocz:
This is to address https://github.com/fornax-navo/fornax-demo-notebooks/issues/102

Still need to go through all the functions, but pushing this up as a draft now, so @jkrick can look into the logic (e.g. heasarc is working as it suppose, but ztf is not yet, so that will change).

I also expect that this will all need to be rebased to pull in changes from #140.


light_curves/code/ztf_functions.py
```diff
@@ -97,23 +94,26 @@ def file_name(filtercode, field, ccdid, qid, basedir=None):
...
         return files[0]
 
-    f = f"{basedir}/field{field:06}/ztf_{field:06}_{filtercode}_c{ccdid:02}_q{qid}_{DATARELEASE}.parquet"
-    return CATALOG_ROOT + f
+    name = f"{basedir}/field{field:06}/ztf_{field:06}_{filtercode}_c{ccdid:02}_q{qid}_{datarelease}.parquet"
+    return CATALOG_ROOT + name
 
 
-def locate_objects(coords_list, labels_list, radius):
+def locate_objects(source_table, radius, datarelease=DATARELEASE, chunksize=10000):
```
troyraen: Same as above for the data release.

I also don't think the chunksize should be an argument here because changing it will very likely lead to worse performance. The time that one TAP query takes increases approx. linearly with the number of objects until you go over about 10,000 objects; beyond that it slows way down. I found this out by doing a lot of testing after trying to run it at scale (500k objects) ran into problems. There's nothing the user can do to change how the TAP query will respond, so I think it's confusing to make this an argument.

light_curves/code/sample_selection.py
```diff
@@ -372,68 +370,47 @@ def get_paper_sample(paper_link,label,coords,labels,verbose=1):
...
-    #keep only those entries in the resulting table which are unique 
-    uniquerows = table.unique(tjoin, keys = 'sc_id')
+    # this join will return 4 entries for each redundant coordinate:
+    # 1 for the match with itself and 1 for the match with the similar
+    # enough coord target then the same thing again for the match, but all of
+    # these 4 will have the same id in the new 'coords_id' column
+
+    # keep only those entries in the resulting table which are unique
+    uniqued_table = unique(tjoin, keys='coord_id')['coord_id', 'coord', 'label']
+    uniqued_table.rename_column('coord_id', 'object_id')
```
troyraen: I recommend that we use `objectid` instead of `object_id` because the former is how it's named in the light curve dataframes and using the same name will help avoid confusion.

light_curves/code/ztf_functions.py
```diff
@@ -56,11 +53,11 @@ def ZTF_get_lightcurve(coords_list, labels_list, nworkers=6, ztf_radius=0.000278
     ztf_df = transform_lightcurves(ztf_df)
 
     # return the light curves as a MultiIndexDFObject
-    indexes, columns = ["objectid", "label", "band", "time"], ["flux", "err"]
+    indexes, columns = ["object_id", "label", "band", "time"], ["flux", "err"]
```
troyraen: If this field name is going to be changed it will need to be updated in all three notebooks under the light_curves/ directory. I think we should avoid this unless there is some strong reason to do it.

troyraen (commented):
I'm really glad you're working on this. I have one suggestion about a field name and also some questions about why arguments are being added to the ZTF functions.

light_curves/code/ztf_functions.py
bsipocz: This is being used only internally, I would instead suggest properly indicating that it's a private function, and then the new kwargs can stay (to make our life easier)

light_curves/code/sample_selection.py
bsipocz: OK, I can change that back. Going forward though I think it would be nice to follow some more widely accepted naming scheme, snake case for variables, and full words, etc.

troyraen: Generally speaking, I'm all in favor of snake case, clear and descriptive variable names, etc. It would be nice to be able to clean up and standardize the code before it's so deeply developed that changes become a real headache. It's obviously important to let the scientists blaze ahead and hack away in a way that works for them so they can flesh out their ideas. But maybe as a group we could develop a workflow where code gets reviewed and revised earlier on, after they've been able to develop the idea a little bit but before there's so much code that changing something is a real pain.

troyraen: (Perhaps more importantly, this might help us catch and correct structural issues early on, like the split between `coords_list` and `label_list`.)

bsipocz: Or the manual looping issues. Though one wonders how much drag early reviews would add to the process, after all a lot of the original code got condensed down already and at this point, and working with the `code_src` files is way nicer and adds fewer code conflicts than it would have been with the first drafts of the notebook when everything was way too fluid.

troyraen: > Though one wonders how much drag early reviews would add to the process

Ya, I agree. It would be easy for this to just end up making the work harder and people more frustrated. Might help to get clear as a group about what should be reviewed at this stage and for the reviewers not to go beyond that scope. Maybe an initial list is:
- structural issues like `coords_list` and `labels_list`
- error-prone patterns like manual looping
- names of things that will be used extensively like `objectid`

bsipocz: > error-prone patterns like manual looping

I tried to pull these out to e.g. #157. Will rebase this PR to be based on top of that, so review can look at only the diff commits.

But some of the changes are difficult to recouple as they are connected to e.g. the structural changes, or only make sense to do once those are done. Sadly even some of the pep8 stuff is in this category, e.g. why fix whitespace when the given codeblock is known to go away in a subsequent commit.

jkrick: This is all good for me to see.  I have added this as an agenda item for wednesday's team meeting.

light_curves/code_src/ztf_functions.py
```diff
@@ -97,23 +94,26 @@ def file_name(filtercode, field, ccdid, qid, basedir=None):
...
         return files[0]
 
-    f = f"{basedir}/field{field:06}/ztf_{field:06}_{filtercode}_c{ccdid:02}_q{qid}_{DATARELEASE}.parquet"
-    return CATALOG_ROOT + f
+    name = f"{basedir}/field{field:06}/ztf_{field:06}_{filtercode}_c{ccdid:02}_q{qid}_{datarelease}.parquet"
+    return CATALOG_ROOT + name
 
 
-def locate_objects(coords_list, labels_list, radius):
+def locate_objects(source_table, radius, datarelease=DATARELEASE, chunksize=10000):
```
bsipocz, line 101: re `chunksize`, the review discussion from the previous PR has to be addressed here:

https://github.com/fornax-navo/fornax-demo-notebooks/pull/157#discussion_r1344765469

bsipocz:
I'm still running into issues with the ZTF functions, and gaia is updated in the other open PR. So would suggest to go ahead and merge this and iron out the remaining problems (with temporary ztf changes I run into some weird pyvo issues without an obvious reason, but at this point it's probably quicker to forget that commit and start over on that function).

bsipocz:
@troyraen - I'm going ahead and merge this now, and we can fix any review in a follow-up.

PR #157, opened 2023-10-03
MAINT: cleanups for tqdm usage and in icecube code

bsipocz:
This PR clears up a couple of smaller issues, and is ready to go in:

 - some bugs in icecube, I suppose it didn't work for me out of the box as I started from new and didn't have a previous reincarnation of the data around
 - removed unnecessary dependency (wget)
 - moved the unused TDE cleanup code out to a new file, so e.g. alerce can be kept as optional dependency (or at least be able to run most of the notebooks without it being installed)
 - moved a lot of hard-wired value out as a kwarg to the functions
 - changed verbose to be a boolean (the typical approach, if more fine tuning is needed, I would suggest to switch to proper logging)
 - fixed tqdm usages and removed manual loopings
 - fixed some style issues, it wasn't comprehensive and only did it for the files being touched in this PR.

### DIFF
--- a/light_curves/code_src/tde_functions.py
+++ b/light_curves/code_src/tde_functions.py
@@ -1,0 +1,27 @@
+from astropy.coordinates import SkyCoord
+
+from alerce.core import Alerce
+
+def TDE_id2coord(object_ids, coords, labels, verbose=1):
+    """ To find and append coordinates of objects with only ZTF obj name
+
+    Parameters
+    ----------
+    object_ids: list of strings
+        eg., [ "ZTF18accqogs", "ZTF19aakyhxi", "ZTF19abyylzv", "ZTF19acyfpno"]
+    coords : list of astropy skycoords
+        the coordinates of the targets for which a user wants light curves
+    labels: list of strings
+        journal articles associated with the target coordinates
+    verbose: int
+        print out debugging info (1) or not(0)
+    """
+
+    alerce = Alerce()
+    objects = alerce.query_objects(oid=object_ids, format="pandas")
+    tde_coords = [SkyCoord(ra, dec, frame='icrs', unit='deg') for ra, dec in zip(objects['meanra'], objects['meandec'])]
+    tde_labels = ['ZTF-Objname' for _ in objects['meanra']]
+    coords.extend(tde_coords)
+    labels.extend(tde_labels)
+    if verbose:
+        print('number of ztf coords added by Objectname:', len(objects['meanra']))
--- a/light_curves/code_src/ztf_functions.py
+++ b/light_curves/code_src/ztf_functions.py
@@ -5,7 +5,7 @@ import astropy.units as u
 import pandas as pd
 import pyvo
 import s3fs
-from tqdm import tqdm
+import tqdm
 
 from data_structures import MultiIndexDFObject
 from sample_selection import make_coordsTable
@@ -34,8 +34,8 @@ def ZTF_get_lightcurve(coords_list, labels_list, nworkers=6, ztf_radius=0.000278
     labels_list: list of strings
         journal articles associated with the target coordinates
     nworkers : int or None
-        number of workers in the multiprocessing pool used in the load_lightcurves function. 
-        This must be None if this function is being called from within a child process already. 
+        number of workers in the multiprocessing pool used in the load_lightcurves function.
+        This must be None if this function is being called from within a child process already.
         (This function does not support nested multiprocessing.)
     ztf_radius : astropy Quantity
         search radius, how far from the source should the archives return results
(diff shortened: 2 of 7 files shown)


troyraen:
> some bugs in icecube, I suppose it didn't work for me out of the box as I started from new and didn't have a previous reincarnation of the data around

What didn't work for you out of the box? I've run it several times starting from a clean `git clone` and it didn't throw any errors and seemed to work properly.

bsipocz:
No, the indexing with 30 threw a fit for a. 30 long 0 indexed thing.

Also, the filename in the conditional was incorrect.

troyraen:
I haven't worked on the guts of the IceCube module and don't know what it's really supposed to be doing so I'm sure I missed things.

light_curves/code_src/ztf_functions.py
```diff
@@ -179,19 +176,17 @@ def load_lightcurves(locations_df, nworkers=6):
...
-    chunksize = 100
     # make sure the chunksize isn't so big that some workers will sit idle
     if len(location_grps) < nworkers * chunksize:
         chunksize = len(location_grps) // nworkers + 1
 
     # "spawn" new processes because it uses less memory and is thread safe (req'd for pd.read_parquet)
     # https://stackoverflow.com/questions/64095876/multiprocessing-fork-vs-spawn
     mp.set_start_method("spawn", force=True)
-    
+
```
troyraen, line 189: I'm glad all this whitespace is getting cleaned up as it's a bit frustrating to work with. I've had to turn off the whitespace auto trim functionality of my IDE in order to work in this repo, but now I'm contributing to the problem. 

I wonder if we should do a separate PR that cleans up the formatting for the whole light_curves/ directory but doesn't make any substantive changes to the code? It would be easier to review that way. In addition to whitespace, the line lengths in some of these files is very long. Running autopep8 would be nice.

bsipocz: I did use a max-line-length of 120, so flake8 is now clear on this PR.

(with one exception of `light_curves/code_src/ztf_functions.py:20:60: W605 invalid escape sequence '\s'`, but I didn't want to touch the ZTF path logic in this PR knowing my other one has some changes for that already)

troyraen: > (with one exception of light_curves/code_src/ztf_functions.py:20:60: W605 invalid escape sequence '\s', but I didn't want to touch the ZTF path logic in this PR knowing my other one has some changes for that already)

Oh, this is just the separator -- telling pandas that the file is space delimited. Maybe it needs to be a raw string: `sep=r"\s+"` instead of `sep="\s+"`.

light_curves/code_src/tde_functions.py
jkrick, line 1: I believe that Shooby uses a bunch of different AGN samples in her ML_AGNzoo work.  Perhaps we could change this file name to be something more like "AGN_sample_selection.py" and we can put some of the other functions she uses in there as well at a later date.  This PR doesn't need to do that, but setting up the filename here to be more general might be helpful.

light_curves/code_src/ztf_functions.py
```diff
@@ -114,36 +114,31 @@ def locate_objects(coords_list, labels_list, radius):
...
-    # calculate the number of iterations needed. it would be easier to just use this while loop directly
-    # for the tap calls, but we want to use tqdm and that's easier in a for loop
-    niterations = 0
-    while niterations * chunksize < len(coords_tbl):
-        niterations += 1
     # do the tap calls
     locations = []
-    for i in tqdm(range(niterations)):
-        result = tap_service.run_async(query, uploads={"coords": coords_tbl[i * chunksize : (i + 1) * chunksize]})
+    for i in tqdm.trange(0, len(coords_tbl), chunksize):
```
jkrick, line 140: In this function and the one below, 'chunksize' is used to mean two different things.  While we are cleaning up this function, could we please change those variable names to be more descriptive?  Perhaps: "max_number_coords" and "max_number_files"?  I don't love those.

jkrick (approved):
I have just a few small suggestions while you are touching these functions.

light_curves/code_src/ztf_functions.py
troyraen: I think "chunk" needs to be in there somehow, otherwise it incorrectly implies that some objects/files will not be processed. `max_objects_per_chunk` and `max_files_per_chunk`?

afaisst (approved):
This all looks good. Thank you Brigitta!

light_curves/code_src/ztf_functions.py
troyraen: If "chunk" seems confusing we could also use "batch". `objects_per_batch` and `files_per_batch`? Could prepend `max_` but could also leave that out of the name and just put those details in the docstring.

bsipocz: `chuck` and `chucksize` feel more prevalent in general than batch and the rest. Maybe this is just a documentation question and it should be explained in the docstring? 

Neither of these functions should really be exposed to the user of the notebook.

Either case, I made a note of this, and propose to do the change in the other PR when I also make some of the functions a bit more formally private.

troyraen: > `chuck` and `chucksize` feel more prevalent in general than batch and the rest. Maybe this is just a documentation question and it should be explained in the docstring?

I agree and I always prefer to use the name that the underlying function uses, or some derivative of it. (One of these functions uses a multiprocessing pool which takes the argument "`chunksize`".) And I'm in favor of explaining of the docstring.

Issue #150, opened 2023-09-27
run gaia light curves at scale


jkrick:
After speeding up the functions which find the source IDs from Gaia and download the epochal photometry, the Gaia_mk_MultiIndex function now takes the majority of the time that gaia_get_lightcurves is working.  This function can likely also be sped up.

I wonder if it is possible to 
- add columns to the table with the desired units
- convert the table to pandas
- convert the data frame to a multi index data frame (organizing this properly is going to be the hard part)

There are likely other ways of doing this.

jkrick:
done as part of PR #160

PR #160, opened 2023-10-06
Gaia Speedup

jkrick:
This code reworks the Gaia-functions to facilitate running at scale
- adds chunks to overcome maximum of the Gaia datalink server
- re-writes the Gaia functions which take the products of the datalink server to a MultiIndexDFObject.  This was a bottleneck to running at scale

The re-write now takes the the datalink product which is some sort of dictionary which has a list of tables for values and converts that to a pandas data frame directly.  Then joins that data frame to a source list to include information from the original input catalog about objectid and label.  Finally it cleans up the units and makes the whole thing a nice MultiIndexDF

This PR relies on #156 so it needs to be tested and merged after that one.
As such, I can't actually test the code here.
I did run what is hopefully the exact same thing on the 'Extra Large' server on the SMCE and it made 500,000 light curves in a blazingly fast 37 minutes.

### DIFF
--- a/light_curves/code_src/gaia_functions.py
+++ b/light_curves/code_src/gaia_functions.py
@@ -12,74 +12,59 @@ from data_structures import MultiIndexDFObject
 from sample_selection import make_coordsTable
 
 
-def Gaia_get_lightcurve(coords_list, labels_list , verbose):
+def Gaia_get_lightcurve(sample_table, search_radius, verbose):
     '''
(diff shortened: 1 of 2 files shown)


jkrick:
Merging this will close #150

bsipocz (commented):
Thanks @jkrick. 

I'm going ahead and cherry-pick this commit on top of #156, so we don't end up in a non-working notebook in the repo at any point.

bsipocz:
Going ahead with this one first, and will rebase #156 if needed. Thanks @jkrick!

PR #162, opened 2023-10-19
Bug fix tqdm

troyraen:
#157 changed the way `tqdm` is imported in ztf_functions.py but some of the `tqdm` calls were not updated to accommodate. This PR makes those updates.


bsipocz (approved):
Good catch, thanks. 

(I just wish this kind of nonsense double say naming would stop, `tqdm.tqdm`, `datetime.datetime`, such a source of difficult to spot issues)

PR #169, opened 2023-11-14
IceCube Speedup

jkrick:
I updated icecube_get_lightcurve to use astropy `search_around_sky` instead of a for loop over all objects.  This should provide a large increase in speed of the ice cube function.  Currently I can't do a good scale test on this because the SDSS function isn't working, but this really should be a speedup, so I am pushing the changes without the scale test.

I would appreciate if someone could check the filtering that I did after the match, to make sure the code does what I wanted it to.  My intent was to have it match all sources within the error radius of of IceCube, then later filter that down to the top three in energy for each input source.

### DIFF
--- a/light_curves/code_src/icecube_functions.py
+++ b/light_curves/code_src/icecube_functions.py
@@ -16,7 +16,7 @@ from tqdm import tqdm
 from data_structures import MultiIndexDFObject
 
 
-def icecube_get_lightcurve(coords_list, labels_list, icecube_select_topN=3, verbose=False):
+def icecube_get_lightcurve(sample_table, icecube_select_topN=3):
     '''
     Extracts IceCube Neutrino events for a given source position and saves it into a lightcurve
     Pandas MultiIndex object.
@@ -24,18 +24,13 @@ def icecube_get_lightcurve(coords_list, labels_list, icecube_select_topN=3, verb
 
     Parameters
     ----------
-    coords_list : list of Astropy SkyCoord objects
-        List of (id,coordinates) tuples of the sources
-
-    labels_list : list of str
-        List of labels for each soruce
+    sample_table : Astropy Table
+        main source catalog with coordinates, labels, and objectids
 
     icecube_select_topN : int
         Number of top events to
 
-    verbose : bool
-        Default False. Display extra info and warnings if true.
-
+ 
     Returns
     --------
     MultiIndexDFObject: IceCube Neutrino events for all the input sources.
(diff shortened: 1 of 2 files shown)


light_curves/code_src/icecube_functions.py
```diff
@@ -44,67 +39,52 @@ def icecube_get_lightcurve(coords_list, labels_list, icecube_select_topN=3, verb
...
+    #save the icecube info in correct format for the rest of the data
+    icecube_df = pd.DataFrame({'flux': flux_icecube, 
+                               'err': np.zeros(len(obid_match)), 
+                               'time': time_icecube, 
+                               'objectid': obid_match, 
+                               'label': label_match, 
+                               'band': "IceCube"})
+
+    #now can use a groupby to only keep the top N (by GeV flux) icecube matches for each object
+    filter_icecube_df = icecube_df.groupby('objectid').head(icecube_select_topN).reset_index(drop=True)
```
troyraen, line 82: Currently this is working correctly and keeping only the largest fluxes (at least for the Yang sample), but all the indexing and filtering that happen between the sort (line 49) and this line is dangerous. In particular, the astropy docs say that the index arrays returned by `search_around_sky` may be sorted differently in the future, which could break this code.

I recommend doing the sort immediately prior to this line. Something like:

```python
icecube_df = icecube_df.sort_values(['objectid', 'flux'], ascending=[True, False])
```

Then I think you can delete line 49.

troyraen (commented):
I think the code is doing what you want, filtering within the error radius and then taking the top 3 (or N). But I do have two suggestions, one about the names of the index arrays and the other about sorting to guarantee that the N largest fluxes get returned. More in the code comments below.

There are other things here that I would comment on, but I'm guessing you're not ready to merge yet? For one, the `filter_icecube_df` that gets returned doesn't actually have a multi-index. Also, the notebook looks like it has unintentional changes. Happy to give more specific comments if/when you're ready.

light_curves/code_src/icecube_functions.py
```diff
@@ -44,67 +39,52 @@ def icecube_get_lightcurve(coords_list, labels_list, icecube_select_topN=3, verb
...
-
-    return (df_lc)
+    # create SkyCoord objects from icecube event coordinates
+    icecube_skycoords = SkyCoord(icecube_events["ra"], icecube_events["dec"], unit="deg", frame='icrs')
+
+    #here are the skycoords from mysample defined above
+    mysample_skycoords = sample_table['coord']
+
+    #Match
+    idx_icecube, idx_mysample, d2d, d3d = icecube_skycoords.search_around_sky(mysample_skycoords, 1*u.deg)
```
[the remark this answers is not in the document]
jkrick: I was thoroughly confused by this, and it is nice to hear that someone else was confused too!  I thought I was doing what the docs said, but I agree, the names should be swapped, so I will go ahead and swap them throughout.

light_curves/code_src/icecube_functions.py
jkrick: thanks, I didn't know about `sort_values`, this was helpful.

jkrick:
I believe I have addressed @troyraen comments.  

Also was able to do the scale testing and the new ice cube function was able to get "light curves" for 5000 sources in 20.5s, so should not have trouble completing 500k sources very quickly.

troyraen (approved):
LGTM. Thanks for fixing the WISE and HEASARC calls in the parallel section as well.

PR #170, opened 2023-11-15
Switch ZTF to use `sample_table`

troyraen:
Followup to #156 

Switches ZTF to use `sample_table` instead of `coords_list` and `labels_list`.


jkrick:
Thanks for finding the bug.  This PR works for me, which is really the important part.  Merge whenever you are ready.

PR #171, opened 2023-11-17
convert all remaining functions to use sample_table instead of coords_list and labels_list

jkrick:
I think I touched every file except ztf_functions.py since that is in PR #170 

I did:
- for panstarrs and TESS/Kepler/K2 functions I took the easy route of just moving the workaround of converting the table back into coords_list and labels_list into those functions and not having it in the main notebook.  My thought there is that these functions need to change in the future anyway in order to access those archives at scale, so for now we will just put a band-aid on those functions so they still work for a small sample, but not think too much about a bigger fix for now.  In other words, it is not worth spending time on these functions now, when we will have to completely re-work them later.  I just made them functional.
- changed HCV to use sample_table
- changed plotting function to be sample_table
- consistent capitalization of function names
- removed now empty function make_coordsTable from sample_selection.py that got replaced in the gaia and ZTF functions by something else.  I did not remove the import line in the ztf_functions.py because I am trying not to touch that file so it won't conflict with Troy's PR.
- Following a discussion on slack, I chose to leave the sample_table as an astropy table both because it was easier at this point with half the functions already using it, and because it is good to have some representation of skycoords in the demo.  I'm not sure it is worth the work to switch now to pandas, but if someone feels strongly that the demo notebook is too inconsistent with different data types, then we can talk about changing it.  

This now means that the light curve notebook is once again fully functional from start to finish in conjunction with PR #170

### DIFF
--- a/light_curves/code_src/gaia_functions.py
+++ b/light_curves/code_src/gaia_functions.py
@@ -9,7 +9,6 @@ from astropy.time import Time
 from astroquery.gaia import Gaia
 
 from data_structures import MultiIndexDFObject
-from sample_selection import make_coordsTable
 
 
 def Gaia_get_lightcurve(sample_table, search_radius, verbose):
@@ -33,7 +32,7 @@ def Gaia_get_lightcurve(sample_table, search_radius, verbose):
     '''
 
     # Retrieve Gaia table with Source IDs ==============
-    gaia_table = Gaia_retrieve_catalog(sample_table , labels_list,
+    gaia_table = Gaia_retrieve_catalog(sample_table , 
                                          search_radius = search_radius,
                                          verbose = verbose
                                          )
--- a/light_curves/code_src/gaia_functions.py
+++ b/light_curves/code_src/gaia_functions.py
@@ -99,7 +99,7 @@ def Gaia_retrieve_catalog(source_table , search_radius, verbose):
     results = j.get_results()
     
     if verbose : print("\nSearch completed in {:.2f} seconds".format((time.time()-t1) ) )
-    if verbose : print("Number of objects matched: {} out of {}.".format(len(results),len(coords_list) ) )
+    if verbose : print("Number of objects matched: {} out of {}.".format(len(results),len(sample_table) ) )
 
     return results
 
--- a/light_curves/code_src/icecube_functions.py
+++ b/light_curves/code_src/icecube_functions.py
@@ -16,7 +16,7 @@ from tqdm import tqdm
 from data_structures import MultiIndexDFObject
 
 
-def icecube_get_lightcurve(sample_table, icecube_select_topN=3):
+def Icecube_get_lightcurve(sample_table, icecube_select_topN=3):
     '''
     Extracts IceCube Neutrino events for a given source position and saves it into a lightcurve
     Pandas MultiIndex object.
--- a/light_curves/code_src/panstarrs.py
+++ b/light_curves/code_src/panstarrs.py
@@ -230,7 +230,7 @@ def search_lightcurve(objid):
 
 
 #Do a panstarrs search
-def panstarrs_get_lightcurves(coords_list, labels_list, radius):
+def Panstarrs_get_lightcurves(sample_table, radius):
     """Searches panstarrs for light curves from a list of input coordinates
     
     Parameters
@@ -249,6 +249,11 @@ def panstarrs_get_lightcurves(coords_list, labels_list, radius):
     """
         
     df_lc = MultiIndexDFObject()
+    
+    #first convert sample_table to coords_list, labels_list
+    coords_list = [(row['objectid'] - 1, row['coord']) for row in sample_table]
+    labels_list = [row['label'] for row in sample_table]
+    
     #for all objects in our catalog
     for objectid, coord in tqdm(coords_list):
         #doesn't take SkyCoord, convert to floats
--- a/light_curves/code_src/panstarrs.py
+++ b/light_curves/code_src/panstarrs.py
@@ -251,7 +251,7 @@ def Panstarrs_get_lightcurves(sample_table, radius):
     df_lc = MultiIndexDFObject()
     
     #first convert sample_table to coords_list, labels_list
-    coords_list = [(row['objectid'] - 1, row['coord']) for row in sample_table]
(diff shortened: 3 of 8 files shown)


jkrick:
As I see it, next steps after this gets merged are to:
- chunk up the HEASARC call so it works at scale
- check the warnings that appear only for the parallel call (deprecation warnings) and make sure we don't need to do anything about these.
- fix the plotting function to make it look good again.  The scales are off or something.  Plots get generated but they don't look good.

troyraen:
> I did not remove the import line [make_coordsTable] in the ztf_functions.py because I am trying not to touch that file so it won't conflict with Troy's PR.

I just realized that I didn't remove it either and I'll need to do that so it doesn't throw an error after this gets merged. It's probably better to do it separately to avoid dealing with conflicts between this PR and #170. I also see a few other unused imports in various files and can clean them up at the same time. I'll plan to do a quick PR after this one gets merged.

Re: sample_table as an astropy Table. I agree with your assessment and am fine to leave the Table.

light_curves/code_src/panstarrs.py
```diff
@@ -249,6 +249,11 @@ def panstarrs_get_lightcurves(coords_list, labels_list, radius):
     """
         
     df_lc = MultiIndexDFObject()
+    
+    #first convert sample_table to coords_list, labels_list
+    coords_list = [(row['objectid'] - 1, row['coord']) for row in sample_table]
```
troyraen, line 254: Same as above; should be:

```suggestion
    coords_list = [(row['objectid'], row['coord']) for row in sample_table]
```

light_curves/light_curve_generator.md
```diff
@@ -194,7 +186,7 @@ df_lc.append(df_lc_fermi)
 
 ### 2.2 IRSA: ZTF
 
-```{code-cell} ipython3
+```{raw-cell}
```
troyraen, line 189: Is this an accidental commit?

light_curves/light_curve_generator.md
```diff
@@ -273,7 +268,7 @@ df_lc.append(df_lc_HCV)
 
 ```{code-cell} ipython3
 gaiastarttime = time.time()
-df_lc_gaia = gaia_get_lightcurve(sample_table, 1/3600., 0)
+df_lc_gaia = Gaia_get_lightcurve(sample_table, 1/3600., 0)
```
troyraen, line 271: This gives me the following error on the IRSA SP using the default kernel:

```
TypeError: TapConn.execute_tappost() takes from 3 to 4 positional arguments but 5 were given
```

I have the full traceback if that's helpful, but I'm guessing this is actually an astroquery bug... This comes from astroquery version 0.4.7.dev8948, which seems to be the version that is pre-installed in the image. @stargaser do you know why we're using a "dev" version here? This works fine if I revert to the latest non-dev version, astroquery==0.4.6.

troyraen (commented):
Thanks for working through these. The biggest issue I see is with the `objectid`s. I also hit a weird error with Gaia, but I think it's an astroquery issue and nothing to do with this PR. Otherwise, this looks good and runs for me.

light_curves/light_curve_generator.md
troyraen: I wonder if we need to submit a bug report to astroquery. But I'll let @bsipocz weigh in first.

troyraen:
I just noticed that line 102 in gaia_functions.py contains `coords_list` and should be updated to `sample_table` (will error if `verbose` != 0).

jkrick:
> I just noticed that line 102 in gaia_functions.py contains `coords_list` and should be updated to `sample_table` (will error if `verbose` != 0).

got it.

jkrick:
I'm going to merge this so we can work on the plotting with both this PR and the ZTF PR.  I'll make an issue on fornax-demo-notebooks repo to figure out the gaia astroquery problem. I need plotting to work for the Dec 1 user panel meeting.

light_curves/light_curve_generator.md
stargaser: Sorry I didn't see this sooner. I've adjusted my Github notifications to track this repository. I've responded in #173 that astroquery version 0.4.7.dev8948 is expected due to their continuous release model. @bsipocz can say more about that.

bsipocz: yes, don't use 0.4.6. If the problem persist with the latest pypi available dev version, then do open a bug report, otherwise upgrade.

bsipocz: OTOH, `Gaia_get_lightcurve` doesn't really adhere to snake case naming conventions, `gaia_get_lightcurve` looked better to me (but this is very much bike shedding from me)

bsipocz:
> verbose != 0

I have cleaned up some of these integer and otherwise not really used verbose options. If `verbose` is still needed, I would suggest changing it to a boolean and having either verbose or quiet versions of the code. If any more is needed that should better be done via proper logging anyway.

light_curves/code_src/gaia_functions.py
```diff
@@ -100,7 +99,7 @@ def Gaia_retrieve_catalog(source_table , search_radius, verbose):
     results = j.get_results()
     
     if verbose : print("\nSearch completed in {:.2f} seconds".format((time.time()-t1) ) )
-    if verbose : print("Number of objects matched: {} out of {}.".format(len(results),len(coords_list) ) )
+    if verbose : print("Number of objects matched: {} out of {}.".format(len(results),len(sample_table) ) )
```
bsipocz, line 102: nitpick: avoid one-liners like this, add a newline after the conditional, it would be more readable. (and using f-strings would make it even more so)

light_curves/code_src/icecube_functions.py
```diff
@@ -16,7 +16,7 @@
 from data_structures import MultiIndexDFObject
 
 
-def icecube_get_lightcurve(sample_table, icecube_select_topN=3):
+def Icecube_get_lightcurve(sample_table, icecube_select_topN=3):
```
bsipocz, line 19: It doesn't matter here, but I like the practice of making all optional arguments mandatory keyword argument. So in user code there won't be any semi-cryptic integers, as well as a more flexible way of extending the API with new arguments. (It doesn't matter here as there are only two arguments, but when a code is expected to evolve, then this practice is useful)

PR #172, opened 2023-11-21
UMAP notebook converted to .md and fixed style

xoubish:


light_curves/code/sample_lc.py
```diff
@@ -41,32 +41,39 @@ def build_sample():
...
 
     # Variable AGN sample from Ranga/Andreas:
     VAGN = pd.read_csv('../data/WISE_MIR_variable_AGN_with_PS1_photometry_and_SDSS_redshift.csv')
     vagn_coords = [SkyCoord(ra, dec, frame='icrs', unit='deg') for ra, dec in zip(VAGN['SDSS_RA'], VAGN['SDSS_Dec'])]
     vagn_labels = ['WISE-Variable' for ra in VAGN['SDSS_RA']]
     coords.extend(vagn_coords)
-    labels.extend(vagn_labels)    
+    labels.extend(vagn_labels) 
+    
+    # George's sample with low radio to IR:
```
jkrick, line 60: I would suggest we don't include George's sample in the demo notebook.

light_curves/code/sample_selection.py
```diff
@@ -360,7 +361,7 @@ def get_SDSS_sample(coords, labels, num, verbose=1):
     """
     # Define the query
     query = "SELECT TOP " + str(num) + " specObjID, ra, dec, z FROM SpecObj \
-    WHERE ( z > 0.0 AND z < 2.0 AND class='QSO' AND zWARNING=0 )"
+    WHERE ( z > 0.0 AND z < 1.0 AND class='QSO' AND zWARNING=0 )"
```
jkrick, line 364: let's make the max_redshift a variable of get_SDSS_sample().  Sorry I didn't think of this earlier.

jkrick (commented):
General comments:
1) I added @bsipocz  as a reviewer so she can help with the conversion to .md and the conflicts which arose from that.  I think we had already converted ML_AGNzoo to a .md, and then you committed one, which is causing some of the conflicts.
2) you have a commit called 'redshifts carryover', but I don't know what that did or what that means.  I thought that meant keeping track of the redshifts of the sources, but maybe you mean something else by it, or quite possibly I missed something?

light_curves/code/ML_utils.py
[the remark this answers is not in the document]
xoubish: I added comments to the "unify_lc" code as well as some lines to the cell before these are called.

bsipocz (commented):
I would suggest to do a big squash to remove the binary/data components/format back and forths first, and then rebase to pick up the changes we already did in the repo to resolve the conflicts. 

I also see remaining ipynb files in the diff. 

Happy to help with any of these tasks, just let me know what would be the most useful (when to step in and do it).

xoubish:
@bsipocz I was not sure whether to delete the ipynb after converting or not. Please feel free to step in now I appreciate the help.

xoubish:
I get the same long warnings for plotting recently. It was not the case before. 
For unify_lc it returns four, latest here: https://github.com/fornax-navo/fornax-demo-notebooks/blob/samplepy/light_curves/code/ML_utils.py

.gitignore
```diff
@@ -18,6 +18,7 @@ cache/
 ./**/*.parquet
 
 ./data/*
+./**/data/*
```
troyraen, line 21: I just randomly noticed this and wanted to mention two things:

First, it is not necessary to make separate entries for directories with the same name at different levels. One entry using the syntax "directory-name/" will ignore every directory with that name (plus everything underneath it) no matter where it is in the repo. So the two lines here can be replaced with one, like this:

```suggestion
data/
```

Second, the .gitignore file the main branch has already taken care of this. I suggest using it instead.

bsipocz: this PR is very much outdated, the rebased version is linked above in my branch. However, I suspect that maybe outdated already as well.

bsipocz: branch is here: https://github.com/fornax-navo/fornax-demo-notebooks/pull/172#issuecomment-1835503967

PR #175, opened 2023-11-22
Update astroquery and s3fs to recommended versions

### LINKED ISSUE #173: Gaia Astroquery bug
From PR #171 

gaia_get_lightcurves 

This gives me the following error on the IRSA SP using the default kernel:

TypeError: TapConn.execute_tappost() takes from 3 to 4 positional arguments but 5 were given
I have the full traceback if that's helpful, but I'm guessing this is actually an astroquery bug... This comes from astroquery version 0.4.7.dev8948, which seems to be the version that is pre-installed in the image. @stargaser do you know why we're using a "dev" version here? This works fine if I revert to the latest non-dev version, astroquery==0.4.6.

stargaser:
* in `requirements.txt` specify the most recent pre-release version of astroquery before the Gaia bug #173 was introduced
* in `requirements.txt` specify a recent version of s3fs

Fixes #173

### DIFF
--- a/light_curves/requirements.txt
+++ b/light_curves/requirements.txt
@@ -3,7 +3,7 @@
 # the container supporting the science use case in question.
 acstools
 astropy
-astroquery>=0.4.7dev8948
+astroquery==0.4.7dev8479
 hpgeom
 lightkurve
 matplotlib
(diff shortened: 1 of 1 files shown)


troyraen (approved):
Thanks, these versions work for me. Will you add a comment to the astroquery one reminding us why it's pinned so that we can unpin it when the bug is fixed?

stargaser:
Thanks @troyraen, I have added the comment.

I will go ahead and merge this one since @jkrick has a demo to do on Friday.

PR #178, opened 2023-11-27
Break heasarc TAP calls, each call has a maximum number of sources

zoghbi-a:
Calling the HEASARC TAP with large search areas appears to time-out. At the request of @jkrick, this PR caps the maximum number of sources in the upload table to 50000, which appears to be an optimal number that avoids making many calls, and runs in a reasonable time.

For the notebook to run, I had also to delete the following line `from sample_selection import make_coordsTable` from ztf_functions.py. `sample_selection` does not have that method. @troyraen?

### DIFF
--- a/light_curves/code_src/heasarc_functions.py
+++ b/light_curves/code_src/heasarc_functions.py
@@ -1,6 +1,7 @@
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
+from astropy.table import vstack
 import pyvo
 from tqdm import tqdm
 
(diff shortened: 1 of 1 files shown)


troyraen:
> For the notebook to run, I had also to delete the following line from sample_selection import make_coordsTable from ztf_functions.py. sample_selection does not have that method. @troyraen?

This is getting fixed in #176

jkrick (commented):
looks good to me and runs for both a small sample and a sample which needs to be broken up.  I have not yet tried 500k sources.  Thanks for implementing the changes. My only request is to add some comment lines around what you are doing in heasarc_functions.py so it is clear for anyone reading the code.

jkrick:
And now I can confirm that it works on the 500k sample.

zoghbi-a:
Thanks @jkrick for testing this.
I have now rebased to include the latest changes in the main branch, and added some comments.

Issue #177, opened 2023-11-22
List of things to consider cleaning up in the light curve notebook


troyraen:
Collecting a list of things we may consider doing as we finish up the notebook light_curve_generator.md.

## Function docstrings

- [x] Review for accuracy. (#189)

## Notebook text

- [x] Review for accuracy (e.g., since we switched `coords_list` --> `sample_table`, instructions in section 1.1 should be updated). (#189)
- [x] Review sections for completeness: (#189)
    - [x] Non-standard Imports
    - [x] References
- [x] Some of the serial cells have text explaining the catalog data or query method (e.g., WISE) but some don't (e.g., ZTF). Consider standardizing this. (#189)

## Code

Note that several MAST functions still iterate over individual sample objects rather than loading in bulk. This is a known issue being handled separately (see #165, #166, #179).

### General

- [x] #174 
- [x] #155 
- [x] #184 
~~- [ ] #159~~ deferred
~~- [ ] #182~~ deferred
~~- [ ] Since it's hard to attach units, etc. to DataFrame columns/values, consider renaming columns to be explicit (e.g., "flux" --> "flux_mjy" and "time" --> "mjd"). OTOH, we may run into issues that mean this is just more confusing (to the point of deceiving), like ZTF actually uses "hmjd" rather than "mjd".~~

### fluxconversions.py

~~- [ ] See if we can use Astropy methods in place of any functions in this file (prompted by https://github.com/fornax-navo/fornax-demo-notebooks/pull/181#discussion_r1411663452).~~ moved to #188 

### heasarc_functions.py

- [x] The two lists `heasarc_cat` and `max_error_radius` have an implicit, one-to-one relationship. Replace the two lists with a dict that ties the correct values together. Then iterate over the dict instead of a list index. (#198)

### mast_functions.py

- [x] Remove this file. It's not being used anywhere. (#198)

### panstarrs.py
- [x] Remove try/except if possible. Otherwise, move as much code as possible outside the `try` block as possible. (#198)

### sample_lc.py

~~- [ ] Update it to, e.g., use sample_table. (This is a script version of the parallel section of the notebook, meant to be called from command line, but has not been kept up to date.) Alternately, consider making the modules directly callable from the command line and write a .sh script to execute them. This should speed up bulk runs because ZTF could use multiple workers.~~ (moved to #195)

### sample_selection.py

[...]

PR #181, opened 2023-11-29
Bug-fix data cleaning and plotting

troyraen:
Included changes:

- Sort data before plotting so time increases monotonically.
- Make the IceCube max search radius a variable and raise the default value to 2 degrees.
- Convert Vega -> AB system for WISE. This is mostly copied from a private conversation with @afaisst.
- Add `objectid` to plot legend.

### DIFF
--- a/light_curves/code_src/fluxconversions.py
+++ b/light_curves/code_src/fluxconversions.py
@@ -1,14 +1,44 @@
+import numpy as np
 from acstools import acszpt
 from astropy.time import Time
 
 
-def convert_wise_flux_to_millijansky(nanomaggy_flux):
+def convert_wise_flux_to_millijansky(flux_nanomaggy, *, band=None):
     """unWISE light curves flux is stored in nanomaggy. Convert to millijansky.
-    
-    See https://www.sdss3.org/dr8/algorithms/magnitudes.php
+
+    See https://www.sdss3.org/dr8/algorithms/magnitudes.php and Meisner et al. (2023, https://iopscience.iop.org/article/10.3847/1538-3881/aca2ab/pdf).
+    For Vega to AB conversion for WISE see https://wise2.ipac.caltech.edu/docs/release/allsky/expsup/sec4_4h.html
+
+    Parameters
+    ----------
+    flux_nanomaggy : float or iterable of floats
+        Flux in nanomaggy.
+    band : str
+        WISE band name corresponding to flux_nanomaggy. One of "W1", "W2", "W3", "W4", or None.
+        If None, band will be set to the name attribute of flux_nanomaggy. Useful when applying this 
+        transform to a DataFrame grouped by band.
+
+    Returns
+    -------
+    flux_mjy : float or iterable of floats 
+        flux_nanomaggy converted to millijansky.
     """
-    millijansky_per_nanomaggy = 3.631e-3
-    return nanomaggy_flux * millijansky_per_nanomaggy
+    if band is None:
+        band = flux_nanomaggy.name
+
+    # Vega to AB magnitude conversions
+    vega_to_ab_conv = {"W1": 2.699, "W2": 3.339, "W3": 5.174, "W4": 6.620}
+
+    # get Vega magnitude from nanomaggy flux as described in Meisner et al. (2023)
+    mag_vega = 22.5 - np.log10(flux_nanomaggy)
+
+    # convert Vega magnitude to AB magnitude
+    mag_ab = mag_vega + vega_to_ab_conv[band]
+
+    # convert AB magnitude to mJy
+    flux_mjy = 10**(-0.4*(mag_ab - 23.9) )  / 1e3
+
+    return flux_mjy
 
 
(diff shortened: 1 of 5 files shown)


troyraen:
@afaisst will you review this? The WISE fluxes are now quite small -- I don't think this what was hoped for. Here's a screenshot for objectid 5 to compare with what we were looking at yesterday.

Screenshot 2023-11-29 at 2 52 50 AM

afaisst:
hmm, yeah, something is wrong here.

light_curves/code_src/fluxconversions.py
```diff
@@ -1,14 +1,44 @@
...
-    millijansky_per_nanomaggy = 3.631e-3
-    return nanomaggy_flux * millijansky_per_nanomaggy
+    if band is None:
+        band = flux_nanomaggy.name
+
+    # Vega to AB magnitude conversions
+    vega_to_ab_conv = {"W1": 2.699, "W2": 3.339, "W3": 5.174, "W4": 6.620}
+
+    # get Vega magnitude from nanomaggy flux as described in Meisner et al. (2023)
+    mag_vega = 22.5 - np.log10(flux_nanomaggy)
```
afaisst, line 33: mag_vega = 22.5 - 2.5*np.log10(flux_nanomaggy)

afaisst: My fault. I forgot the "2.5" factor there. That should fix it:
`mag_vega = 22.5 - np.log10(nanomaggy_flux)`
to
`mag_vega = 22.5 - 2.5*np.log10(nanomaggy_flux)`

troyraen:
Thanks @afaisst! Looks a lot better to me now. Let me know if you still see any issues with WISE or other plotting stuff.

Screenshot 2023-11-29 at 11 55 36 AM

troyraen:
@jkrick As far as I know, this takes care of all the plotting issues. Take a look and let me know if you see anything else.

afaisst:
looks good! Looks like we are back to the original version :)

afaisst:
One more quick thing: in the legend, do we want to say "Gaia BP", "Gaia G", etc?

troyraen:
> One more quick thing: in the legend, do we want to say "Gaia BP", "Gaia G", etc?

That makes sense to me, but it would mean changing the name of the band in the light curve dataframe. Is that something we want to do? I'm not sure if those band names are hard coded anywhere, but if they are those spots would need to be updated as well.

afaisst:
> > One more quick thing: in the legend, do we want to say "Gaia BP", "Gaia G", etc?
> 
> That makes sense to me, but it would mean changing the name of the band in the light curve dataframe. Is that something we want to do? I'm not sure if those band names are hard coded anywhere, but if they are those spots would need to be updated as well.

i see. I remember we had it before like that in the data frame. I guess it's fine for now because there is no ambiguity so far (e.g., PanSTARRS has the "panstarrs [band]" label.

jkrick:
Thanks for figuring out the magnitudes bug!

could we rename the bands just for the plot legend? I agree it would be nice to have, but don't want to do a major code revision for it.  

Also, where is panstarrs for the most recent plot?  Is that a true bug or just an omission?

troyraen:
> Also, where is panstarrs for the most recent plot? Is that a true bug or just an omission?

Looks like MAST is down right now. Just getting this:

```
HTTPError: 502 Server Error: Bad Gateway for url: https://catalogs.mast.stsci.edu/api/v0.1/panstarrs/dr2/mean.csv?ra=0.28136&dec=-0.09789&radius=0.0002777777777777778
```

> could we rename the bands just for the plot legend? I agree it would be nice to have, but don't want to do a major code revision for it.

I mean... we *can*, but I don't recommend it. I think it's confusing to show a label on the plot that is different than what's in the dataframe. It would also introduce a bit of code that has to be manually maintained (because we'd have to hard-code the label mapping instead of letting it be automated; and doing this in a different spot than where the original label gets generated means that we need to remember we did that if we want to make a label change in the future). But this isn't a huge deal here (it's just labels.. it's not like we're not talking about changing data) so if that's what you really want I can do it here.

afaisst:
I wouldn't change it in the plotting routine. I would change them at the source (when we query the Gaia data). The plotting routine should not "change the data".

light_curves/code_src/fluxconversions.py
```diff
@@ -1,14 +1,44 @@
+import numpy as np
 from acstools import acszpt
 from astropy.time import Time
 
 
-def convert_wise_flux_to_millijansky(nanomaggy_flux):
+def convert_wise_flux_to_millijansky(flux_nanomaggy, *, band=None):
```
bsipocz, line 6: I'm cleaning up my github backlog, and wonder why not use astropy and its unit equivalencies instead of coding up the flux--mag conversion here?

troyraen: Perhaps we should review all the functions in fluxconversions.py to see whether we can use Astropy methods instead. Adding this to #177.

PR #183, opened 2023-11-30
Bugfix dtypes

troyraen:
I just noticed that the flux and error dtypes are "object" in the light curve dataframe. I'm not aware of any problems this currently causes, but it might in the future. These really should be floats. This is happening because the pandas `explode` method returns columns with "object" dtypes. So, this PR sets the correct dtypes right after every `explode`.

### DIFF
--- a/light_curves/code_src/WISE_functions.py
+++ b/light_curves/code_src/WISE_functions.py
@@ -76,8 +76,7 @@ def locate_objects(sample_table, radius):
 
     locations = locations_table.to_pandas()
     # locations contains one row per object, and the pixel column stores arrays of ints
-    # "explode" the dataframe into one row per object per pixel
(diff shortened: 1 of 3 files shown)


troyraen:
I think this is all that's needed, but I haven't tested it fully. I plan to get back to it next week after the User Panel.

troyraen:
@jkrick the figure at the end of lc_classifier.md section 2.5 looks good to me (comparing with main branch), but I'm less familiar with this notebook. Will you check it?

jkrick:
This bug fix works as intended in the lc_classifier.md notebook.  I don't get it to run through to a plot, another bug has come up with NaN's in the data frame, but that should be a separate PR (for me).  Ok to merge.

PR #185, opened 2023-12-05
Bugfix Table to DataFrame conversion

troyraen:
An error occurs in the WISE function under rare circumstances when the number HEALPix pixels to be searched is the same for all sample objects. In this case, Astropy attaches the pixel column as a fixed-length array rather than an "object" dtype, and a Table with this column type cannot be converted to a DataFrame. This PR switches the order of operations so the Table is first converted to a DataFrame and then the pixel column is attached.

 Steps to reproduce the error:

```
import astropy.units as u
# assuming you're in the directory light_curves/code_src
from sample_selection import clean_sample, get_yang_sample
from WISE_functions import WISE_get_lightcurves

coords =[]
labels = []
get_yang_sample(coords, labels)   #2018ApJ...862..109Y
sample_table = clean_sample(coords, labels)

bandlist = ['W1', 'W2']
WISE_radius = 1.0 * u.arcsec

# send only the first 2 sample objects. on the main branch, this fails with a ValueError.
WISE_get_lightcurves(sample_table[:2], WISE_radius, bandlist)
```

### DIFF
--- a/light_curves/code_src/WISE_functions.py
+++ b/light_curves/code_src/WISE_functions.py
@@ -61,7 +61,7 @@ def locate_objects(sample_table, radius):
     Returns
     -------
     locations : pd.DataFrame
-        dataframe of location info including ["objectid", "coord", "label", "pixel"]
+        dataframe of location info including ["objectid", "coord.ra", "coord.dec", "label", "pixel"]
     """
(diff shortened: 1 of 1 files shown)


bsipocz (approved):
I like that this also avoids making one extra copy of the Table. Thanks!

PR #186, opened 2023-12-13
MAINT: update dependencies and versions

bsipocz:
This is to close https://github.com/fornax-navo/fornax-demo-notebooks/issues/184 and to address the comment in https://github.com/fornax-navo/fornax-demo-notebooks/issues/173#issuecomment-1850792665

### DIFF
--- a/light_curves/requirements-lc_classifier.txt
+++ b/light_curves/requirements-lc_classifier.txt
@@ -9,3 +9,5 @@ sktime
 tqdm
 googledrivedownloader
 scikit-learn
+acstools
+requests
--- a/light_curves/requirements-lc_classifier.txt
+++ b/light_curves/requirements-lc_classifier.txt
@@ -1,6 +1,11 @@
 # List python dependencies here to be installed at the start of the
 # notebook. Libraries can be removed from the list when they get shipped in
 # the container supporting the science use case in question.
+numpy
+pandas
+matplotlib
+astropy
 sktime
 tqdm
 googledrivedownloader
+scikit-learn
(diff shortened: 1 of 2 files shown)


light_curves/requirements-lc_classifier.txt
troyraen, line 1: Starting from a clean conda environment, I needed the following to be added before I could import everything used in lc_classifier.md:

- `acstools`
- `requests`

troyraen (commented):
A couple comments below. Otherwise, LGTM.

light_curves/requirements.txt
```diff
@@ -3,15 +3,20 @@
 # the container supporting the science use case in question.
 acstools
 astropy
-astroquery==0.4.7dev8479 # pinned for Gaia bug astropy/astroquery#2844
+astroquery>=0.4.7.dev0
```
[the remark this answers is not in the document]
bsipocz: yes, it won't upgrade, but if you start from a new install, it'll pick up a good version. Given that the bug is in a section of dev versions but not before or after, this still feels a good way to state that.

light_curves/requirements-lc_classifier.txt
bsipocz: ahh, good. I only grepped the import from the md file, but not from the code parts that it also run.

That being said, I think we should maybe limit the usage of requests, mast and panstars should ideally be accessible with astroquery/pyvo.

bsipocz:
The review has been addressed, so I went ahead and merged this.

PR #189, opened 2023-12-16
Light curve text cleanup

jkrick:
- Uniformizing the serial function descriptions and cells, major text cleanup
- removing unused functions (ATLAS, ASAS-SN)
- not touching ML_utils.py or sample_lc.py or tde_functions.py since they are not for the light curve notebook
- somewhat cleaning docstrings in code_src, but my eyes gaze over so easily on these that I am sure I am missing things.
	- Mostly for the functions we are using from other sources (panstarrs, HCV, ..)in many cases I don't know the datatypes, and it would be super tedious to figure it out, so I am punting on this.
	- not touching plot_functions.py, that thing is a beast, and might change in the next round of cleanup that Troy will do.
- changed "non-standard imports" to list all imports (as requested at some previous time).  
- Checked to see that the list of references was complete.   I couldn't find any references for the other modules imported, so if they exist they are not obvious in 1 min of searching their pages.
- edited the main README.md for the whole repo which was cut and pasted from some proposal as a placeholder years ago.  This appears to have caused a conflict and I really don't know why.  Maybe I'm not supposed to update that file??

I can't really recommend anyone try to read through all the changes I made, but I can guarantee that the code does run through to completion.  I promise I did not make any large code changes.  I would be fine to merge this without review so that the next major work on the code cleanup can happen without delay.


troyraen (approved):
LGTM and runs on ISP. Thanks!

PR #190, opened 2023-12-16
CI: adding publishing GHA

bsipocz:
This still works with the empty notebooks, I'm still not sure what the best place would be for the set of executed ones that we could tag with versions, etc.

Anyway, I'm going ahead and merge this as we see whether this resolves the permission issues around the `_static` directory or not. 

(also, note: the newer version of the theme looks different than the one I had locally, most of it is an improvement, but e.g. now we lack the title from below the logo, and that has to be fixed)

### DIFF
--- a/.github/workflows/ci_publish.yml
+++ b/.github/workflows/ci_publish.yml
@@ -1,0 +1,43 @@
+name: Publish HTML
+
+on:
+  push:
+    branches:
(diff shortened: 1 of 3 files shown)


bsipocz:
I go ahead and merge this now as the action is supposed to run only on `main`.

PR #194, opened 2023-12-16
BUG: fix function names after rename

bsipocz:
Some of the functions unfortunately got renamed in one of the refactoring PRs, but the code that used the functions hasn't been updated.

note: the rename is unfortunate as now the functions don't follow the standard, all lowercase snake_case naming convention of python functions.


jkrick (approved):
Sorry that was probably me changing the names to keep upper/lower case names consistent between all function calls.

This looks fine to me.  

Does this fix the problem of ML_AGNzoo running through to completion?

bsipocz:
Yes, `ML_AGNzoo` runs through completion for me, but I suppose the doesn't converge issue is still around (at least the plots doesn't look particularly good to me). See more comments about what I found in the slack DM, tldr I don't think it's an issue in my rebase/cleanup of the markdown but due to changes in the code files in `code_src`.

PR #197, opened 2023-12-23
Clean up warnings

### LINKED ISSUE #174: warnings on parallel section
check the warnings that appear only for the parallel call (deprecation warnings) and make sure we don't need to do anything about these.

troyraen:
Closes #174 

- Remove warnings filter from main notebook.
- Remove `plt.tight_layout` from plotting function. It is throwing a warning, and the figures look good to me without it.
- Add comment that it's ok to ignore the lightkurve "error".
- Remove print statements from `HCV_get_lightcurves`.


troyraen:
I see no more warnings.

jkrick (approved):
This looks good to me and runs to completion on the SMCE without unexpected warnings.

Issue #203, opened 2024-01-09
Bug: Light curves containing NaN are not plotted


troyraen:
When a (single band) light curve contains at least one NaN none of its data points are plotted. For example, see SDSS (object 2, band "TESS") or (object 35, band "K2").

This is happening because `scipy.stats.zscore` "propagates" NaN by default. To ignore NaNs instead, the kwarg `nan_policy='omit'` can be used (see [scipy.stats.zscore](https://docs.scipy.org/doc/scipy/reference/generated/scipy.stats.zscore.html)).

Another option would be to remove NaNs from the dataframe before calculating the zscore.

jkrick:
I like the idea of dropping NaN's from the data frame before plotting.  Can that go in the new` _clean_lightcurves` before the `zscore` gets calculated?

sorry I read this after the PR with all the plotting, but I do think it would make sense to take care of this in that same PR #199 , so I'll copy this comment over there.

troyraen:
> I like the idea of dropping NaN's from the data frame before plotting. Can that go in the new `_clean_lightcurves` before the `zscore` gets calculated?

Yes, done. But also, I wonder if these should not be in the dataframe to begin with? In other words, maybe each archive call/function should drop rows containing NaN before returning their results. Are those rows actually useful at any point?

jkrick:
I can't think of a case where we need a NaN.  Mabe it is useful at some level to know that an archive call worked???  I'll take your recommendation on where the NaN dropping should get done, but I am find with it happening at the plotting stage.  It's easier at this point to not go back into every archive call function.

troyraen:
> I'll take your recommendation on where the NaN dropping should get done

I think this is really a science question. I suppose I could be useful at some point to just know that an observation was taken even if it didn't result in good data.

troyraen:
In that case, we can call this done now.

PR #204, opened 2024-01-09
Clean up plot_functions.py

### LINKED ISSUE #199: Clean up plot_functions.py
This issue was separated from #177.

The following refer to the `create_figures` function unless otherwise indicated.

- [x] Remove `sample_table` as an argument. Iterate over `df_lc` (grouped by `objectid`) instead of `sample_table` rows.
- [x] Remove the manual counter `cc`. If this counter is really needed, get it by adding `enumerate` to the `for` loop.
- [x] The function is very long function; breaking it up into smaller ones would improve readability. Consider making separate functions each band (or group of bands) and for the formatting.
- [x] Remove function `setup_text_plots` because it is not being used anywhere.

troyraen:
Closes #199 
(This should be merged *after* #198)

## Changes
- separate `create_figures` into smaller functions
- remove `sample_table` arg and `cc` manual counter
- remove unused function `setup_text_plots`

## Of Note

Bands with fluxes in electrons/sec ('Kepler', 'K2', 'TESS') are now scaled based on all bands with fluxes in mJy. Previously they were being scaled based only on the bands that had actually been plotted before the electrons/sec bands; and if no bands had already been plotted, the electrons/sec band was omitted. I can provide figures for comparison if anyone wants to see them.

I have removed the line that dropped data points with "time" > 65000 MJD (= 2036-11-03). I checked the Yang sample and the first 100 objects in the SDSS sample and did not find any such data points. If there was a specific need for that line, please let me know. It seems to me that, if there are any such data points, they should be dropped from the dataframe altogether, not just for plotting.

### DIFF
--- a/light_curves/code_src/plot_functions.py
+++ b/light_curves/code_src/plot_functions.py
@@ -1,10 +1,15 @@
 import os
 
-import matplotlib as mpl
 import matplotlib.pyplot as plt
 import numpy as np
 from astropy.table import Table
 from scipy import stats
+
+
+# List bands that need special treatment
+ELECTRON_BANDS = ["Kepler", "K2", "TESS"]  # flux in electrons/sec
+ICECUBE_BAND = "IceCube"  # "fluxes" are just events (= limits on plot)
+ZTF_BANDS = ["zg", "zr", "zi"]  # will be plotted in "B" zoom-in for better visibility
 
 
 def create_figures(df_lc, show_nbr_figures, save_output):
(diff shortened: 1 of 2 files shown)


jkrick (approved):
Thanks for tackling this plotting function.  I  like how you divided it up into separate functions, so now it is way easier to manage.

I will defer to Andreas to check the plotting.  I did run it on the SMCE, and the plots appear by quick glance to be reasonable.  

I agree that anything with an erroneous future dates should be removed at some data cleaning place in the ML notebook (I made a note for myself)

Now that the plotting function is more organized, is it possible to make the plotting colors band specific (ie., so in every plot W1 is always gray, W2 is always black, etc.)?  It would be a nice enhancement, not required at all.

jkrick:
I like the idea of dropping NaN's from the data frame before plotting.  Can that go in the new` _clean_lightcurves` before the `zscore` gets calculated?  for issue #203

afaisst (approved):
I ran it through on SMCE and all looks good.

Regarding "Of Note" #2: I don't know why we imposed a cut at high MJDs. There might have been a problem in the beginning (wrong data?), but that seems to not occur anymore. I agree that if there is a case like this, we would have to remove it from the data frame all together.

bsipocz (approved):
The refactoring looks good to me, thank you for doing this cleanup!

troyraen:
> Now that the plotting function is more organized, is it possible to make the plotting colors band specific (ie., so in every plot W1 is always gray, W2 is always black, etc.)? It would be a nice enhancement, not required at all.

Sure, done. Here is what the colors are now. (And if folks don't like certain assignments, they can be changed.)

lightcurve_100

troyraen:
> I like the idea of dropping NaN's from the data frame before plotting. Can that go in the new `_clean_lightcurves` before the `zscore` gets calculated? for issue https://github.com/fornax-navo/fornax-demo-notebooks/issues/203

Done

Issue #196, opened 2023-12-20
edit intro readme file to be more useful


jkrick:
I like this example here: https://github.com/heliocloud-data/science-tutorials

PR #205, opened 2024-01-11
Rendering suggestions

jkrick:
the PR is just a spelling change, but here are additional suggestions:

- Can we change the headings under the Table of Contents to be:
"Photometry" (instead of "Multi-band forced photometry) and "Time Domain"

- Then the "Photometry" heading doesn't need to be a link to a page with a link to the notebook.  Since there is only one notebook in that heading, I think it is fine to not have the heading be a link.  The notebook title under it will be the link.

- Inside the rendered notebooks, I don't need the orange # that appears next to the headings when I hover over them and appears to move that section to the top.  This is just a preference and not a strong opinion.

- For the forced photometry rendered notebook, the section heading doesn't appear to have worked for section 2.  I think "Use fornax cloud access API..." should be 2.1, etc.  Sorry if this is a problem with the original notebook, and leet me know if you want me to change anything in the .md file

- Issue #196 is already opened to edit the text blurbs in the intro readme file which will propagate to here.

### DIFF
--- a/index.md
+++ b/index.md
@@ -1,4 +1,4 @@
-# Table of Content
+# Table of Contents
 
 ```{toctree}
 ---


jkrick:
Here is a draft blurb for the "Time Domain" page:

Many fields in astronomy use the time variability to understand physical processes.  Large datasets combined with time domain studies are becoming increasingly available to astronomers (ZTF, WISE, Rubin, etc.,).  We show here two demo notebooks at the intersection of big data and time domain astronomy.   The first notebook automatedly generates light curves from NASA and non-NASA archival sources in an efficient manner allowing this to be run on big data samples.  The second takes that a light curve sample and runs machine learning classifiers on it.

jkrick:
More suggestions:
- the top menu which has the symbols for github, binder, print, etc. needs to be clicked on to click on the option within the menu, ie., you need to click on the main image, then on the option you want.  Which is one click too many if that is customizable. (This works the way I expect, with a single click, on the epyc website)
- the pdf print option for the notebooks prints the table of contents, is it possible to not print the table of contents, and just the notebooks?
- colab fails, and binder has been trying without success for the last 5 minutes, let's remove those links unless we can easily make them work.  It's not a priority to make those services work because we are saying "come use the NASA SP".  
- I do like the documentation on the main table of contents page, so let's go with it for now.  Instead of section heading "More details about Fornax portal", can we make it "Documentation"
- Also, while I am thinking about header names, maybe "Table of Contents" should be something else, it really shows the notebooks.  Let's try calling it "Tutorial Notebooks", and I will edit the main README.md to not have the heading "Fornax-demo-notebooks" since that is redundant.

I think that is it for the first round, really all cosmetic stuff.

bsipocz:
I'll keep this open as a placeholder until I go through your list of suggestions.

bsipocz:
> the top menu which has the symbols for github, binder, print, etc. needs to be clicked on to click on the option within the menu, ie., you need to click on the main image, then on the option you want. Which is one click too many if that is customizable. (This works the way I expect, with a single click, on the epyc website)

I don't think it's customisable. For the epyc version I made the mistake of not updating the CSS theme to be the latest, I'm afraid the more structured way is the future. Good news that this is a very widely used CSS theme, so the one-too-many click is exactly the same as everywhere else.

> the pdf print option for the notebooks prints the table of contents, is it possible to not print the table of contents, and just the notebooks?

not sure whether it's customisable, docs says download is for the currently viewed page, and it maybe a global configuration. Could we open a separate issue for this, so the idea is not lost (I may just need to look into more example or ping a jupyterbook contact, both of these would happen slower than I would like to get the other changes from your list merged.

> colab fails, and binder has been trying without success for the last 5 minutes, let's remove those links unless we can easily make them work. It's not a priority to make those services work because we are saying "come use the NASA SP".

OK, this is easy to do and I'll change it. Binder worked for me when I checked, but it's very much a hit and miss and very often doesn't work. Colab is interesting that doesn't work, it should, but I think some upstream fix is necessary for it to pick up the ipynb version rather than the markdown.
Other than that, I think it may even be possible to have a NASA SP link under the same launch button, but I haven't yet looked into it of what else is required to make it work.

> Instead of section heading "More details about Fornax portal", can we make it "Documentation"

Done

> Also, while I am thinking about header names, maybe "Table of Contents"...

Done

> Re Readme.md

Github picks that up for the repo readme blurb, so ideally it has some generic info about the repo. But if its content doesn't make sense for the HTML pages (I expect it is a shortened version of some parts of the documentation anyway), we can jot not put it into the index page (right now it's included as is)

bsipocz:
The `Table of Content` line is now gone, so I'm closing this PR. However, I moved all feedback comment lines to a checklist in #306.

PR #206, opened 2024-01-11
Update README.md

### LINKED ISSUE #196: edit intro readme file to be more useful

jkrick:
removed uncessesary section heading of "fornax-demo-notebooks"

edited executive summary to include more details about the notebooks.

hoping this will now stand alone nicely, and also fit in with the rendering work that Brigitta is doing.

this will close #196

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
-# fornax-demo-notebooks
-Tutorial notebooks of fully worked science use cases for the Fornax project
-
 ## Executive Summary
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,9 +1,9 @@
-# fornax-demo-notebooks
```
bsipocz, line 1: I'm fairly certain that we still need a top level heading, but we can try and see and fix only if it causes problem.

bsipocz (approved):

bsipocz:
Thanks @jkrick! I go ahead and merge this, so we see if there is any issues because of the missing top level heading, or all is fine.

PR #208, opened 2024-01-12
Convert Panstarrs -999 to NaN

### LINKED ISSUE #201: Bug: Panstarrs "-999" should be converted to NaN
Panstarrs encodes NaN as -999. Our code should check the flux and error columns for -999 and convert it back to NaN. (This is currently being done for magnitudes in the function `improve_filter_format`, but this doesn't fix it because we're using the fluxes not magnitudes.)

Object 7 in the SDSS sample has this issue and the negative value causes `create_figures` to throw an error.

troyraen:
Closes #201 

This just adds two lines to convert the values so that it doesn't throw an error when plotting.


jkrick (approved):
thanks for both finding and fixing this bug.

PR #209, opened 2024-01-12
changed to a smaller area for faster runs of the notebook

jkrick:
changed the forced_photometry area to a small area instead of the full 48' area and documented how long the code would take to run on the full area.

### DIFF
--- a/forced_photometry/multiband_photometry.md
+++ b/forced_photometry/multiband_photometry.md
@@ -126,9 +126,11 @@ except ImportError:
 coords = SkyCoord('150.01d 2.2d', frame='icrs')  #COSMOS center acording to Simbad
 
 #how large is the search radius, in arcmin
-radius = 0.5 * u.arcmin  #start with a manageable radius and increase as needed
-#full area of COSMOS is radius = 48'
-#running this code on the full COSMOS area takes ~24 hours on a 128core server
+# full area of COSMOS is radius = 48'
+# running this code on the full COSMOS area takes ~24 hours on a 128core server, therefore
+# start with a manageable radius and increase as needed
+radius = 0.5 * u.arcmin  
+
 
 #specify only select columns to limit the size of the catalog
 cols = [
--- a/forced_photometry/multiband_photometry.md
+++ b/forced_photometry/multiband_photometry.md
@@ -6,9 +6,9 @@ jupytext:
     format_version: 0.13
     jupytext_version: 1.15.2
 kernelspec:
-  display_name: Python [conda env:tractor]
+  display_name: Python 3 (ipykernel)
   language: python
-  name: conda-env-tractor-py
+  name: python3
 ---
 
 # Automated Multiband Forced Photometry on Large Datasets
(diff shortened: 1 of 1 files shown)


forced_photometry/multiband_photometry.md
```diff
@@ -6,9 +6,9 @@ jupytext:
     format_version: 0.13
     jupytext_version: 1.15.2
 kernelspec:
-  display_name: Python [conda env:tractor]
+  display_name: Python 3 (ipykernel)
   language: python
-  name: conda-env-tractor-py
+  name: python3
```
bsipocz, line 11: this looks OK and the ones in general work better for generic toolings, but we may need to find a way to ensure that this metadata is not accidentally changed based on the contributor and their workflow.

forced_photometry/multiband_photometry.md
```diff
@@ -128,7 +126,9 @@ except ImportError:
 coords = SkyCoord('150.01d 2.2d', frame='icrs')  #COSMOS center acording to Simbad
 
 #how large is the search radius, in arcmin
-radius = 48.0 * u.arcmin  
+radius = 0.5 * u.arcmin  #start with a manageable radius and increase as needed
+#full area of COSMOS is radius = 48'
+#running this code on the full COSMOS area takes ~24 hours on a 128core server
```
bsipocz, line 131: I would avoid inline comments, and in general move comment above the code line. Silly convention, but python styling likes to leave one space after the comment char and two before if it's inline.

```suggestion
# full area of COSMOS is radius = 48'
# running this code on the full COSMOS area takes ~24 hours on a 128core server, therefore
# start with a manageable radius and increase as needed
radius = 0.5 * u.arcmin  

```

bsipocz (approved):
Nitpicking on ways of commenting, but otherwise looks good!

forced_photometry/multiband_photometry.md
jkrick: I can't tell if this is an important thing or not.  When I open a notebook, sometimes I get asked which kernel I would like to use, so I click on one, and since kernels change, and I am not perfect, sometimes apparently that action actually gets recorded.  Is this something we can just delete altogether?  Is it something we can configure to not record in the first place?  It seems as though someone wanting to take this code and run it anywhere not on an SP would run into trouble with this current setup?  Or if it just gets ignored, then I am ok to ignore it as well.

bsipocz: >  When I open a notebook, sometimes I get asked which kernel I would like to use, so I click on one, and since kernels change, and I am not perfect, sometimes apparently that action actually gets recorded.

That is all good, and you should be able to choose kernels as you like. 

> Is this something we can just delete altogether?

no, the metadata cannot be deleted, and may not cause issues for people taking the notebook (maybe they get a warning if a kernel doesn't exist and it falls back to their default kernel). What we could do is to not commit the changes. 
These ones above are in fact nicer as a more generic metadata is created, so I would like to keep it going fwd.

bsipocz: I'll google around, maybe there are config options that would leave metadata changes out of the diff. If there aren't any, then adding changes as patches may work. (either `git add --patch` or using a git GUI, we should figure out a way that would be convenient for you and other team members.)

PR #212, opened 2024-01-16
Add initial Spectroscopy case

afaisst:
Add initial notebooks for spectroscopy use case. Mainly just merge this so I can create a new branch to start actual work on the spectroscopy use case.


bsipocz:
before this goes into the situation of #172 due to complications from md vs ipynb, could you change the format to md instead, right on the second commit?

afaisst:
> before this goes into the situation of #172 due to complications from md vs ipynb, could you change the format to md instead, right on the second commit?

done!

jkrick (approved):

bsipocz (approved):
Thanks!

PR #213, opened 2024-01-17
Drop TESS rows with negative or NaN flux

### LINKED ISSUE #202: Filter out TESS light curves containing negative fluxes
Some TESS fluxes are negative. Is that ok, or do we want to filter those out?

For example, see SDSS objects 2, 21, 54, 70, 72, 81, and 94.

troyraen:
Closes #202


jkrick (approved):
looks good and runs as expected.  Thanks for implementing the change.

Issue #137, opened 2023-09-08
Document amount of CPU and memory needed to run full notebook at scale


troyraen:
We have made significant improvements to several of the archive calls in the light curve notebook. These have been tested individually. The notebook now needs to be tested in full, at scale (500,000 objects) to determine the required resources (number of CPUs and GB memory). Then we should document these numbers, perhaps at the top of the notebook. I think it would also be helpful to list these numbers for some small number (100?) of objects. The notebook cell that calls all the archives in parallel will require the most resources.

jkrick:
In addition, we need some verbiage in the intro section of the code which helps people decide if they should be running the code in the parallel section, or the script.

troyraen:
In testing #215, I was able to obtain the SDSS 500k sample on irsakusp. I did this twice, and it ran smoothly both times.

- Runtimes for each archive were (2nd run):

  ```
  gaia : 73 min (1.2 hr)
  heasarc : 17 min (0.3 hr)
  icecube : 5 min (0.1 hr)
  wise : 279 min (4.7 hr)
  ztf : 456 min (7.6 hr) -- 8 workers
  ```

- CPU load average was about 2.5 (by eye).

- RAM -- I wish I would have paid more attention to this, but I think it needs between 16GB and 32GB.

- Disk-space used by each archive's parquet file containing it's light curve data:

  ```
  448M    lightcurves.parquet/mission=gaia
  209K    lightcurves.parquet/mission=heasarc
  6.0M    lightcurves.parquet/mission=icecube
  199M    lightcurves.parquet/mission=wise
  6.6G    lightcurves.parquet/mission=ztf
  7.2G    total
  ```

troyraen:
I've tested on the SMCE, and ran into at least two problems with larger sample sizes:

1. There is not enough disk space to write the light curves to parquet files. My (home?) directory can only hold 4GB total. (For SDSS 500k, ZTF parquet alone is 6.6GB.) This is the major blocker.

2. `top` can be misleading and show more total RAM than is actually allotted. E.g., a machine with 4G/4CPU shows: `GiB Mem :     15.3 total`. I'm guessing this is because it allows bursting, but eventually kills your job if it uses more than allotted for too long. This can lead one to believe that the job is within resource limits when it's actually not. This is not a major blocker, just something to be aware of.

troyraen:
[...]

Issue #207, opened 2024-01-11
Bug: Astroquery version for Gaia


troyraen:
The ISP has astroquery version 0.4.7.dev8948 installed which causes the error described in #173. Running `pip install -r requirements.txt` does not update astroquery to a newer version. So astroquery has to be manually uninstalled and reinstalled, otherwise it throws an error. Further, when light curves are fetched in parallel the error gets lost in the background and does not get display -- so Gaia data does not get loaded and the user does not get any notification that something went wrong.

(Yes, the ISP should be updated. But I think we should prevent the notebook from allowing this to happen, regardless of ISP.)

troyraen:
I wonder if we should add a try/except at the top of the notebook that does something like this:

```
try:
    from astroquery.gaia import Gaia
    Gaia.launch_job_async(args)
except TypeError:
    pip uninstall astroquery
    pip install astroquery
```

bsipocz:
Minor comment: `pip install -U --pre astroquery` should be the way to go rather than uninstall and install.

I'm close enough to releasing 0.4.7, so this may naturally get resolved soon. After that I plan to switch to a more frequent, date-based release model, so tagged releases will be more frequent.

troyraen:
In that case I think it's probably fine to wait for v0.4.7. I'll leave this open to remind us to update requirements.txt at that time.

troyraen:
Using astroquery v0.4.7, the Gaia call is throwing a new error. Pasted below. I have not looked into it and don't know whether our code needs to change or if there is a bug in somewhere else in the chain.

```python
KeyError                                  Traceback (most recent call last)
Cell In[3], line 4
      1 gaiastarttime = time.time()
      3 # get Gaia light curves
----> 4 df_lc_gaia = gaia_get_lightcurves(sample_table, search_radius=1/3600, verbose=0)
      6 # add the resulting dataframe to all other archives
      7 df_lc.append(df_lc_gaia)

File ~/raen/fdn-tmp/light_curves/code_src/gaia_functions.py:45, in gaia_get_lightcurves(sample_table, search_radius, verbose)
     40     return MultiIndexDFObject()
     42 # Extract Light curves ===============
     43 # request the EPOCH_PHOTOMETRY from the Gaia DataLink Service
---> 45 gaia_df = Gaia_retrieve_epoch_photometry(gaia_table)
     47 #if the epochal photometry is empty, return an empty dataframe
     48 if len(gaia_df) == 0:

[...]

PR #215, opened 2024-01-26
Add notebook and code to run light_curve_generator at scale

### LINKED ISSUE #195: Make a script to run light curve functions at scale
Update sample_lc.py to, e.g., use `sample_table`. (This is a script version of the parallel section of the notebook, meant to be called from command line, but has not been kept up to date.) Alternately, consider making the modules directly callable from the command line and write a .sh script to execute them. This should speed up bulk runs because ZTF could use multiple workers.

troyraen:
Closes #195, #137

Adds a new notebook, scale_up.md, and related code to run the light_curve_generator code at scale.

If reviewing, I recommend reading through the notebook first, then looking at code if interested.

New files in the light_curves directory:

- scale_up.md
    - notebook demonstrating how to launch large scale runs, monitor them automatically, and diagnose a problem (out of RAM)
- code_src/helpers/scale_up.py
    - python code to facilitate large scale runs
- code_src/helpers/scale_up.sh
    - bash script to execute and monitor large scale runs
- code_src/helpers/top.py
    - python code to parse top output into pandas dataframes and make figures
- output/lightcurves-demo-SDSS-500k/logs/
    - log files generated previously by running the bash script to get light curves for 500,000 SDSS objects -- used to demonstrate large scale runs without having to actually execute one on the fly

Updated existing files:

- README.md
    - add text from @jkrick describing the notebooks in this directory
- light_curve_generator.md
    - update the parallel section, reference the scale_up.md notebook
- code_src/ztf_functions.py
    - make the workers print their PIDs


light_curves/output/lightcurves-demo-SDSS-500k/logs/gaia.log
bsipocz, line 1: the outputs and logs are added by mistake, I presume?

troyraen: Not a mistake, that is intentional. I'm planning to clean up this PR so it's ready for review this afternoon. Then you can see how the logs are used and we can go from there.

light_curves/scale_up.md
```diff
@@ -0,0 +1,558 @@
...
+- The helper can be used in combination with any parallel processing method.
+- The helper can load `top` output from a file to pandas DataFrames and make some figures.
+
+The bash script allows the user to launch the full run with a single command and provides some automated logging and `top` monitoring.
+- When a run is launched, the script calls the helper to gather the requested sample and then launches jobs for each archive query in separate, parallel processes.
+  The script then exits leaving the archive jobs running in the background.
+- The script redirects stdout and stderr to log files, one per job.
+- The script tells the user what the process PIDs are and where the log and data files are.
+- In case the run needs to be canceled, the script can find and kill all the processes it launched.
+- The script can monitor `top` during the run, saving `top` output to a log file at a user-defined interval(s). The helper can be used to load this file in python.
```
jkrick, line 58:
```suggestion
- In a nutshell, the script does the following:
  - calls the helper to gather the requested sample and then launches jobs for each archive query in separate, parallel processes.
  - redirects stdout and stderr to log files
  - tells the user what the process PIDs are and where the log and data files are.
  - monitor `top` during the run, saving `top` output to a log file at a user-defined interval(s).
  - exits leaving the archive jobs running in the background.
In case the run needs to be canceled, the script can find and kill all the processes it launched.

```

light_curves/scale_up.md
```diff
@@ -0,0 +1,558 @@
...
+
++++
+
+### Parallel processing methods: bash script vs. python's `multiprocessing`
+
++++
+
+- Bash script. Recommended for most runs with medium to large samples. Allows ZTF to use additional parallelization internally, and so is often faster (ZTF often takes the longest and returns the most data for AGN-like samples). Writes stdout and stderr to log files, useful for monitoring jobs and resource usage. Can monitor and record `top` to help identify CPU and RAM usage/needs.
+- Python's `multiprocessing` library. Useful as a demonstration. May be convenient for runs with small to medium sample sizes. Has drawbacks that may be significant including the inability to use ZTF's internal parallelization and that it does not save the log output (stdout and stderr) to file.
```
jkrick, line 67:
```suggestion
- Python's `multiprocessing` library. May be convenient for runs with small to medium sample sizes (< few thousand). Has drawbacks that may be significant including the inability to use ZTF's internal parallelization and that it does not save the log output (stdout and stderr) to file.
```

light_curves/scale_up.md
```diff
@@ -0,0 +1,558 @@
...
+- When a run is launched, the script calls the helper to gather the requested sample and then launches jobs for each archive query in separate, parallel processes.
+  The script then exits leaving the archive jobs running in the background.
+- The script redirects stdout and stderr to log files, one per job.
+- The script tells the user what the process PIDs are and where the log and data files are.
+- In case the run needs to be canceled, the script can find and kill all the processes it launched.
+- The script can monitor `top` during the run, saving `top` output to a log file at a user-defined interval(s). The helper can be used to load this file in python.
+
++++
+
+### Parallel processing methods: bash script vs. python's `multiprocessing`
```
jkrick, line 62: I would suggest this section goes above the description of the bash script and python helper

light_curves/scale_up.md
```diff
@@ -0,0 +1,558 @@
...
+import helpers.top  # load `top` output to DataFrames and make figures
+from data_structures import MultiIndexDFObject  # load light curve data as a MultiIndexDFObject
+from plot_functions import create_figures  # make light curve figures
+```
+
+## Example 1: Multi-wavelength light curves for 500,000 SDSS AGN
+
++++
+
+This example shows how to launch a large-scale run using the bash script, monitor its performance, and diagnose a problem (out of RAM).
```
jkrick, line 95:
```suggestion
This example shows how to launch a large-scale collection of light curves using the bash script, monitor its performance, and diagnose a problem (out of RAM).
```

light_curves/scale_up.md
```diff
@@ -0,0 +1,558 @@
...
+# run_id will be used in future calls to manage this run, and also determines the name of the output directory.
+$ run_id="demo-SDSS-500k"
+
+# Execute the run:
+$ bash code_src/helpers/scale_up.sh \
+    -r "$run_id" \
+    -j '{"get_samples": {"SDSS": {"num": 500000}}, "archives": {"ZTF": {"nworkers": 8}}}' \
+    -a "core"  # shortcut for "Gaia HEASARC IceCube WISE ZTF"
+```
+
```
jkrick, line 119: clarify if this run is only for ZTF, since ZTF is explicitly mentioned, but then the line below it implies that it is searching Gaia, Heasarc, Icecube, WISE, ZTF.  

Also, one day we will need to change "core" to include MAST, or just remove it.

light_curves/scale_up.md
```diff
@@ -0,0 +1,558 @@
...
+ZTF log from 2024/03/01 (failure):
+
+```{code-cell}
+!cat output/lightcurves-demo-SDSS-500k/logs/ztf.log
+```
+
+The logs above show that the ZTF job loaded the light curve data successfully ("100%") but exited without writing the parquet file (no "Light curves saved" message like Gaia).
+The data was lost.
+There is also no indication of an error; the job just ends.
+We can spot what happened by looking at the `top` output.
```
jkrick, line 211:
```suggestion
We can diagnose what happened by looking at the `top` output.
```

light_curves/scale_up.md
```diff
@@ -0,0 +1,558 @@
...
+
++++
+
+#### `top`
+
+During a run, you can use the `-t` (top) flag to have the script monitor `top` for you and save the output to a log file.
+To capture the complete run, open a second terminal and call the script with `-t` right after launching the run.
```
jkrick, line 218: I don't understand this.  If i just run the bash scale_up.sh with a -t will it do to the top monitoring, or do I need to call the bash script twice in two different terminal windows?

light_curves/scale_up.md
```diff
@@ -0,0 +1,558 @@
...
+```{code-cell}
+fig = toplog.plot_overview(between_time=("22:55", "23:10"))
+```
+
+In the second panel above we see the ZTF worker processes (which load the light curve data) ending just after 23:00 and then the ZTF parent process continues by itself.
+Around 23:06 in the fourth panel, we see the ZTF job's memory usage rise sharply to almost 100% and then drop immediately to zero when the job terminates.
+This coincides with the total available memory dropping to near zero in the third panel.
+This shows that the machine did not have enough memory for the ZTF call to successfully transform the light curve data collected by the workers into a `MultiIndexDFObject` and write it as a parquet file, so the machine killed the job.
+
+The solution is to rerun ZTF on a machine with more RAM.
```
jkrick, line 271:
```suggestion
The solution is to rerun the ZTF archive call on a machine with more RAM.
```

light_curves/scale_up.md
```diff
@@ -0,0 +1,558 @@
...
++++
+
+```bash
+$ yaml_run_id=demo-kwargs-yaml
+$ bash code_src/helpers/scale_up.sh -r "$yaml_run_id" -d "use_yaml=true" -a "core"
+```
+
++++
+
+### Using a json string
```
jkrick, line 468: This is slightly too complicated for me.  Already the yaml file is a stretch, but I see that it is a good idea for bookkeeping.  Maybe leave this in the function documentation, but not in the tutorial notebook.

jkrick (changes requested):
I've made a first pass at the text and trying to run the notebook on irsakusp.  There is so much information here, my brain is still trying to wrap itself around the process, but to that end the code appears to be really helpful and well thought out!  My first wave of comments are below.  Maybe I will wait for you to address the fourth comment before I do further review.  

General comments
1.  After reading through it all and thinking about the overall structure of the tutorials, I would still like some parallelization to be in the light_curve_generator.md .  Two reasons, 1) the intermediate sized use cases, and 2) to make sure people see multiprocessing in as many notebooks as reasonable (in case they never click through to the  scale_[up.md](http://up.md/)).  Can we easily add back in what was in the main branch prior to this PR.  I like that code cell rather than the example in the scale_up because it doesn't use the helper and scripts so people don't necessarily need to understand that additional complication.  
2. When that gets added back in, I think we should edit the text in the cell 4. Parallel processing the archives to include something like " The below cells show how to increase the speed of the multi-archive search using python's built in parallel processing by taking advantage of the multiple available CPUs on Fornax.  The following will speed up searches of light curves for numbers of targets ranging from a few tens to a few thousand.  For larger sample sizes please see the tutorial in the related notebook scale_up.md in the same folder as this one.  Running the below multiprocessing calls on very large samples (hundreds of thousands) will not work because of the way the platform is setup to cull users which appear to be inactive. "
3. suggested text for light_curves folder readme.md below.  Now that we have so many notebooks in this folder it makes sense to describe them.
4. It would be great if the notebook could 'run' to completion.  Right now it doesn't for me, for likely two reasons. 1). the cells which have $bash commands give errors.  Can we maybe make these cells 'raw' type.  Just a suggestion for how to leave these in the notebook, but have the notebook still run.  2) in the markdown cells you have sections with ```{code-cell}  that aren't getting executed for me on irsakusp.  eg., the imports in the intro cell.  I guess I am assuming that the notebook does actually run some things, ie., reading the output of the logs and plotting some things.

Time Domain
In this set of Use Case Scenario we work towards creating multiband light curves from multiple archival and publication resources at scale and classifying and analysing them with machine learning tools.  Tutorials included in this folder are:
1. light_curve_generator.md  This notebook automatically retrieves target positions from the literature and then queries archival sources for light curves of those targets.  This notebook is intended to be run on a small number of sources (<~ few hundred)
2. scale_up.md  This notebook uses the same functions as light_curve_generator(above) but is able to generate light curves for large number of sources (~1000 -> millions?)
3. light_curve_classifier.md This notebook takes output from light_curve generator and trains a ML classifier to be able to differentiate amongst the samples based on their light curves.  
4. ML_AGNzoo.md This notebook takes output from the light_curve_generator(above) and visualizes/compares different labelled samples on a reduced dimension grid.

troyraen:
@jkrick Thanks for all of your feedback. I haven't gone through all of it yet, but noticed your comments about the formatting issues in particular. The notebook ran to completion for me, though I manually converted it to a .ipynb first. The imports cell should obviously be a code cell (and it appears to be in the raw markdown, but not the rendered version). For the cells with bash commands, I chose the markdown format because it provides language-aware syntax highlighting and so is easier to read, though I'm still trying to figure out if there's a better way to include it in a notebook. But regardless, those cells should definitely not be executing. Seems like the cell definitions got mangled. I'll work on it.

bsipocz:
>  For the cells with bash commands, I chose the markdown format because it provides language-aware syntax highlighting and so is easier to read, though I'm still trying to figure out if there's a better way to include it in a notebook. But regardless, those cells should definitely not be executing. Seems like the cell definitions got mangled. I'll work on it.

It should be possible to choose a shell type for a code block, not just python. But if they are not supposed to be executing as part of the notebook, then choosing the markdown and syntax highlight is certainly the right way to go.

light_curves/scale_up.md
troyraen: I've updated this text using most of what you suggested, but correcting the bit about `top` which I see was confusing.

light_curves/scale_up.md
troyraen: Done

light_curves/scale_up.md
troyraen: Clarification added. 

Also, "core" and "all" were useful shortcuts for me so that I didn't have to remember them every time. But if you'd prefer not to expose them here in the notebook we could just list out the archive names.

light_curves/scale_up.md
troyraen: To have the bash script monitor top, it needs to be called separately from the call that launches the run. I have clarified the text in various places in the notebook.

light_curves/scale_up.md
troyraen: Removed

troyraen:
This is ready to review again. There are two potential issues:

1. TESS and HCV return no data for the Yang sample. The helper still writes parquet files for these, they are just empty. This sometimes causes the next cell (read the parquet files to a dataframe) to throw an error. I'm not sure why it works sometimes but not other times. Right now I think the best option is probably to avoid writing the parquet file if there is no data.
2. The Gaia call throws an error when using astroquery v0.4.7 (newest version). See #207.

@jkrick I think I've addressed all of your comments.

> 1. ... Can we easily add back in what was in the main branch prior to this PR.

Done

> 2. ... edit the text in the cell 4. Parallel processing

Done. I also added the same keyword arguments to this section as given in the serial section for all archive calls to close the loop on https://github.com/fornax-navo/fornax-demo-notebooks/pull/230#discussion_r1496759366.

> 3. suggested text for light_curves folder readme.md below.

Added.

> 4. It would be great if the notebook could 'run' to completion.

I fixed all the formatting issues, at least as far as I can tell. Current version runs to completion for me on irsakusp (.md file) and smce (after manually converting .md -> .ipynb). One note: The bash syntax highlighting in the markdown cells renders nicely when using Theme: JupyterLab Light, but has some shadow-like effect that makes it difficult to read using JupyterLab Dark. This is unfortunate, but I don't know of a better solution at the moment.

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+sample_table = helpers.scale_up.run(build="sample", run_id=yaml_run_id, use_yaml=True)
+```
+
+Bash example:
+
++++
+
+```bash
+$ yaml_run_id=demo-kwargs-yaml
+$ bash code_src/helpers/scale_up.sh -r "$yaml_run_id" -d "use_yaml=true" -a "core"
```
jkrick, line 487: pointing out another place where "core" gets used.

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+For example, progress may to slow to a crawl, or it may run smoothly for several hours and then crash suddenly.
+If the job is running in the background, print statements and error messages may get lost and never be displayed for the user if they are not redirected to a file.
+
++++
+
+### Needs and wants for large-scale runs
+
++++
+
+The main goal is to speed up the run, so we want to look for opportunities to parallelize.
```
jkrick, line 517:
```suggestion
The main goal is to speed up total time it takes to run the full code, so we want to look for opportunities to parallelize.
```

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+
+Scaling up to large sample sizes brings new challenges.
+Even if things go smoothly, each function call may need to use a lot more resources like CPU, RAM, and bandwidth, and may take much longer to complete.
+Also, we'll want to run some functions in parallel to save time overall, but that will mean they must compete with each other for resources.
+
+These issues are complicated by the fact that different combinations of samples and archive calls can trigger different problems.
+Inefficiencies in any part of the process -- our code, archive backends, etc. -- which may have been negligible at small scale can balloon into significant hurdles.
+
+Problems can manifest in different ways.
+For example, progress may to slow to a crawl, or it may run smoothly for several hours and then crash suddenly.
```
jkrick, line 508:
```suggestion
For example, progress may slow to a crawl, or it may run smoothly for several hours and then crash suddenly.
```

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+
++++
+
+### Needs and wants for large-scale runs
+
++++
+
+The main goal is to speed up the run, so we want to look for opportunities to parallelize.
+We can group the light_curve_generator code into two main steps: (1) gather the target object sample; then (2) generate light curves by querying the archives and standardizing the returned data.
+All of the archive calls have to wait for the sample to be available before starting, but then they can run independently in parallel.
```
jkrick, line 519: Maybe add that gathering the sample does not take long compared to the archive calls, so the real opportunity for parallelization lies in the archive calls.

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+
+We need to be able to monitor the run's resource usage and capture print statements, error messages, etc. to log files in order to understand if/when something goes wrong.
```
jkrick, line 521:
```suggestion
It is useful to be able to monitor the run's resource usage and capture print statements, error messages, etc. to log files in order to understand if/when something goes wrong.
```

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+
+If the run fails, we'd like to be able to restart it without having to redo steps that were previously successful.
+To accomplish this, the inputs and outputs need to be less tightly coupled than they are in the light_curve_generator notebook.
+Specifically, we want to save the sample and light curve data to file as soon as each piece is collected, and we want the archive functions to be able to get the `sample_table` input from file.
+
+The python helper and bash script were specifically designed to fulfill many of these wants and needs.
+
++++
+
+### What the user should do
```
jkrick, line 533:
```suggestion
### Advice for the user
```

jkrick (approved):
One requirements error and a bunch of comments.  Sorry if I got a little carried away on the editing.

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+While the jobs are running, the user can call the script again with a different flag to have it save `top` output to a log file at a user-defined interval.
+The helper can be used to load this file to pandas DataFrames and make some figures.
+
++++
+
+### Imports
+
+```{code-cell}
+# Ensure all dependencies are installed
+!pip install -r requirements.txt
```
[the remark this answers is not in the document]
bsipocz: hmm, should we suggest `pip install -U -r requirements.txt`, or to work from a clean environment first and only run the install at the first time? I'm not exactly sure, but am rather certain these types of issues will pop up for users all the time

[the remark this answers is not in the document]
troyraen: Oh, the final line says `PermissionError: Forbidden`. I'm guessing you encountered this on irsakusp? ZTF and WISE are both still in that private bucket and I'm guessing you don't have credentials for that bucket setup there. SMCE has the credentials pre-packaged in the image.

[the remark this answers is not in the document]
troyraen: Re: the `pip` dependency resolver error, I could switch this to use a pyarrow filesystem object which I think(?) doesn't rely on boto3/botocore. But I don't know whether other code in the light_curves directory will require boto anyway.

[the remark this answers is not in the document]
bsipocz: a full astroquery/pyvo install will bring it on, though I don't think we do a full dependency install for those, or in fact for any of the packages.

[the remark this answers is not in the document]
bsipocz: but overall, this is one of those questions that could be moved out to a separate issue from this PR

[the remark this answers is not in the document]
jkrick: and yes, I forgot about the bucket credentials thing and was running on the ISP.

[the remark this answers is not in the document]
troyraen: #262

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+
++++
+
+## Example 2: Parallelizing the light_curve_generator notebook
+
++++
+
+This example shows how to parallelize the example from the light_curve_generator notebook using the helper and python's `multiprocessing`.
+An advantage of the example shown here is that it automatically saves the sample and light curve data to disk after loading them and can automatically skip those steps in subsequent calls and use the files instead.
+
```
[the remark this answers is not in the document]
troyraen: Actually there's no difference in performance here vs the light_curve_generator parallel call. The only advantages are what I mention above about saving files and that this will print out the PIDs so you know which job is running in which process. I clarified by adding:

> This is a small sample (Yang, 30 objects). If you want a sample larger than a few hundred, consider using the bash script instead.

light_curves/code_src/helpers/scale_up.py
```diff
@@ -0,0 +1,356 @@
...
+    if build == "lightcurves":
+        return _build_lightcurves(**my_kwargs_dict)
+
+    return _build_other(keyword=build, **my_kwargs_dict)
+
+
+# ---- build functions ---- #
+
+
+def _build_sample(*, get_samples, consolidate_nearby_objects, sample_file, overwrite_existing_sample, **_):
```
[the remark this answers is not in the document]
troyraen: The `_` indicates that they are considered "private" (== the user should not have to interact with them), but still, they really should have docstrings. I made sure the public functions were documented but skipped the private ones on the first pass. I will open an issue.

light_curves/scale_up.md
troyraen: I replaced "core" in all the function and script calls with the explicit list of archives. I also scrubbed the reference out of all the documentation. The actual list still exists and can be used, but the user would have to look directly at the `ARCHIVE_NAMES` dictionary where it's defined in order to find it. I also I renamed it to "scaled" since this is the subset that is expected to run successfully at scale. Probably still not the best name, but better than "core".

light_curves/code_src/helpers/scale_up.py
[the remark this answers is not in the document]
troyraen: #263

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+df_lc.sample(10)
+```
+
+Now we can make figures:
+
+```{code-cell}
+_ = create_figures(df_lc=MultiIndexDFObject(data=df_lc), show_nbr_figures=1, save_output=False)
+```
+
```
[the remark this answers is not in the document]
troyraen: I confirmed that ZTF data is collected and displayed here when credentials for the relevant AWS bucket are available.

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+This notebook shows how to collect multi-wavelength light curves for a large sample of target objects.
+This is a scaling-up of the [light_curve_generator](light_curve_generator.md) and assumes you are familiar with the content of that notebook.
+This code should be used for sample sizes greater than a few hundred.
+Many of the challenges, needs, and wants discussed in the appendix and addressed throughout the notebook are common when scaling up code and the ideas can be applied to other use cases.
+
+Notebook sections are:
+
+- "Overview" describes functionality of the included bash script and python helper functions. Compares some parallel processing options.
+- "Example 1" shows how to launch a large-scale run using the bash script, monitor its progress automatically, and diagnose a problem (out of RAM).
```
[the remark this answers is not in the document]
troyraen: I agree and have tried but am also unable to get it to work. I tried the only option from the [myst-parser docs](https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html) that seemed like it should work in all cases (Annotating a syntax block with `(target)=`). External links to files or URLs seem to work, but not links to a section within the notebook. I'm guessing this doesn't work because the markdown in different notebook cells is isolated from the others in some way.

Interested to hear if @bsipocz knows of a solution.

light_curves/scale_up.md
```diff
@@ -0,0 +1,559 @@
...
+
+```{code-cell}
+toplog.pids_df.sample(5)
+```
+
+```{code-cell}
+fig = toplog.plot_overview()
+```
+
+In the figure above, "Total" is the machine total, the "%CPU" panel is grouped by PID (process ID), and the "%MEM" panel is grouped by job.
```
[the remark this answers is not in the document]
troyraen: Done

PR #217, opened 2024-01-29
Bugfix ASCII Table reads in HCV

troyraen:
The HCV function has been using `ascii.read`. I assume this was meant to be using `astropy.io.ascii` but the module does not import it. This results in an error because the builtin `ascii` module is used instead.

This PR switches to use `Table.read` instead of `ascii.read`.


bsipocz (approved):

bsipocz:
Thanks!

PR #225, opened 2024-02-09
Shooby feb

xoubish:
Hi all,

I think this notebook should be ready to be merger. Run this with the sample I copied in smce ~/efs/sampleAGN_lc if you want to test.

PR #229, opened 2024-02-17
MAINT: cleanup shoobyFeb after premature merge of #225

bsipocz:
Merging PR #225 has been manually reverted from the repo history as it was prematurely merged containing a lot of unrelated files. This PR is the exact copy without those files.


troyraen (approved):
I'm fine with all of this, but it's not quite what I was expecting. @xoubish will need to confirm what she wants. I thought we were dropping these:

- data_structures.py
- sample_lc.py` (this whole file will be removed in #215)
- sample_selection.py
- requirements.txt

xoubish:
I no longer use 
- data_structures.py 
- sample_lc.py
- sample_selection.py 
so reverting to what existed or dropping in case it was not there is ok.
In requirements.txt I have changed umap-learn[plot] to umap-learn so the change is ok.

bsipocz:
Ahh, I was too fast to merge it and didn't cleanup those.

For `sample_lc`, I would suggest to go ahead with the removal in the other PR, then the changes can stay until then. Also, wasn't sure why the changes were in requirements, it may not pick up all the dependencies without specifying the extras. Anyway, cleanup to these files can totally come in a follow-up PR, the diff is not in the order of thousands of lines.

PR #230, opened 2024-02-17
Standardize function names and signatures

troyraen:
This standardization is being done in preparation for #215 to make the calls for bulk runs easier.

Changes:

- Rename sample functions using syntax `get_<name>_sample`, snake case.
- Rename archive modules using syntax `<archive>_functions.py`, snake case.
- Rename archive functions using syntax `<archive>_get_lightcurves`, snake case.
- Standardize all `get_*_sample` and `*_get_lightcurves` function signatures:
    - Make all arguments keyword-only, except `coords` and `labels` (sample) or `sample_table` (lightcurves).
    - Set defaults for all keyword arguments.
    - Remove astropy units from arguments (this makes it easier to call these functions from, e.g., the command line -- more complex types like astropy units would require special handling in #215).
- Make IceCube use an absolute path to the 'data/' directory.


light_curves/code_src/sample_selection.py
```diff
@@ -318,7 +318,7 @@ def get_graham_sample(coords, labels, verbose=1):
 
 #SDSS QSO sample of any desired number
 #These are "normal" QSOs to use in the classifier
-def get_SDSS_sample(coords, labels, num,zmin=0,zmax=1, verbose=1):
+def get_sdss_sample(coords, labels, *, num=10, zmin=0, zmax=10, verbose=1):
```
jkrick, line 321: Did you intentionally change this to zmax = 10 from 1?

light_curves/light_curve_generator.md
```diff
@@ -372,30 +359,14 @@ callback = parallel_df_lc.append  # will be called once on the result returned b
...
-    pool.apply_async(
-        TESS_Kepler_get_lightcurves, (sample_table, lk_radius), callback=callback
-    )
-    pool.apply_async(
-        WISE_get_lightcurves, (sample_table,  wise_radius, bandlist), callback=callback
-    )
-    pool.apply_async(
-        ZTF_get_lightcurve, (sample_table, ztf_nworkers), callback=callback
-    )
+    pool.apply_async(gaia_get_lightcurves, args=(sample_table,), callback=callback)
```
jkrick, line 362: what is the rationale for hiding the arguments of the functions other than sample_table?  
I would have thought that some of these arguments the user might actually want to change, or at least know they could change, eg., heasarc_catalogs, so if we show the syntax in the tutorial, then it is more clear to the user what the free parameters are.

light_curves/light_curve_generator.md
```diff
@@ -80,19 +80,19 @@ from astropy.table import Table
...
 from sample_selection import (clean_sample, get_green_sample, get_hon_sample, get_lamassa_sample, get_lopeznavas_sample,
-    get_lyu_sample, get_macleod16_sample, get_macleod19_sample, get_ruan_sample, get_SDSS_sample, get_sheng_sample, get_yang_sample)
-from TESS_Kepler_functions import TESS_Kepler_get_lightcurves
+    get_lyu_sample, get_macleod16_sample, get_macleod19_sample, get_ruan_sample, get_sdss_sample, get_sheng_sample, get_yang_sample)
+from tess_kepler_functions import tess_kepler_get_lightcurves
 # Note: WISE and ZTF data are temporarily located in a non-public AWS S3 bucket. It is automatically
 # available from the Fornax SMCE, but will require user credentials for access outside the SMCE.
-from WISE_functions import WISE_get_lightcurves
-from ztf_functions import ZTF_get_lightcurve
+from wise_functions import wise_get_lightcurves
```
jkrick, line 94: I don't care about capitalization (which is potentially why the code needed this sort of PR), but I don't know if others care about capitalization.  For example, WISE is capitalized because it is an acronym, so are we doing something grammatically incorrect (or worse, offending someone) by changing it to lower case?  Same for the other telescope names.

light_curves/code_src/sample_selection.py
troyraen: Yes. Should it stay at 1? I was concerned that `num=500_000` would return fewer than 500K sources, and the user may not realize they need to increase `zmax`. I didn't actually try it. (I am skeptical of these z limits being here in the first place. I think they were committed by accident.)

light_curves/light_curve_generator.md
troyraen: Now that everything but `sample_table` has a default they aren't required here anymore, and the calls are simpler without them. The serial section shows what they are at how to use them (at least most of them). This section  gives one example (ZTF) of how to submit them in this call, and I figured it was fine to leave the others out and let the user add them if they want.

That was my rational. but I'm happy to add them back in if you want.

light_curves/code_src/sample_selection.py
xoubish: I think its good to have z_min, z_max in the sample build_up where possible. I might have added those in. And here's how the query can change to not sort by redshift by default I think:

query = f"SELECT TOP {num} specObjID, ra, dec, z FROM SpecObj \
WHERE ( z > 0.1 AND z < {zmax} AND class='QSO' AND zWARNING=0 ) \
ORDER BY NEWID()"

light_curves/light_curve_generator.md
troyraen: In python it is pretty standard to name functions and modules using snake case (including lower case). By contrast, classes typically use camel case.

troyraen: #215 really benefits from these using a uniform syntax, including the case. It could deal with something non-uniform but would require more code and/or rely on the user to enter the name exactly right including case.

light_curves/code_src/sample_selection.py
jkrick: good point about the 500k needing potentially a larger redshift range.  However, if I want a smaller sample, I want to know for instance that the lower redshift ones are high enough to not be extended sources, and the higher redshift ones are observable with some reasonable telescope to do follow up, ie., not z = 10.  Can we keep these variables visible so it is really obvious to users (including me!) what exactly is getting selected?  I think having them more visible would be a good nudge on the shoulder to not forget these other aspects of sample selection.

troyraen: What do you mean by "more visible"?

jkrick: ahhhh, maybe that is the difference between me and someone who writes better code than me....I think function definitions in a different file than the main notebook are one level removed, and therefore 'hidden' or 'less visible' to the user.  I assume most users won't actually look at the function definitions.  Maybe you don't start with that assumption?

troyraen: @xoubish do you want me to add that option to randomize z? If so, do you want the randomization on or off by default?

Randomizing z will result in different samples being returned each time (at the very least, they will be in a different order). This means the sample is not reproducible. That's my biggest hesitation, so I would leave it off by default.

troyraen: @jkrick would I agree that code in a different file than the notebook is "less visible" and I don't expect most users to look at it. Are you just wanting the z arguments to be shown in the main notebook where this gets called?

xoubish: Why would ordering by ID be randomized and not reproducible?

troyraen:
```
ORDER BY NEWID()
```

As I understand it, this creates a new, random ID every time. So ```SELECT TOP N``` will return different rows every time.

xoubish: You are right. And apparently it doesn't get a seed. I am not sure if people want to build sample more than once that they care about getting the same sample. It doesn't bother me much to go with sorted redshifts either. I will use this on the forked version for publication but here it can go either way.

troyraen: > I will use this on the forked version for publication but here it can go either way.

Ah, ok then I'm happy to add it here. There are some things that will need to be different in your fork (otherwise you wouldn't need a fork), but functionality like this doesn't need to be one of them. The less out-of-sync your fork is with this repo, the easier it will be to manage.

troyraen: I've added both a `randomize_z` option in this function and a demo of `z` kwargs in the main notebook.

light_curves/light_curve_generator.md
jkrick: I think we agreed to add redshift info back to the serial calls, but leave the parallel section as is until the review for PR #215

jkrick (approved):

PR #231, opened 2024-02-19
Enhance sample_selection.py

troyraen:
Merge _after_ #230.

Changes to sample_selection.py:

- Add ~~`get_tde_sample`~~`get_ztf_objectid_sample` - Load sample by ZTF objectid.
- Add `get_csv_sample` - Load sample from CSV file containing (at least) RA and Dec.
- Add `get_papers_list_sample` - Wrapper for `get_paper_sample` to fetch multiple papers with one call (helpful for #215).
- Replace `nonunique_sample` with a kwarg in `clean_sample` called `consolidate_nearby_objects`.

### DIFF
--- a/light_curves/code_src/sample_selection.py
+++ b/light_curves/code_src/sample_selection.py
@@ -1,6 +1,7 @@
 from requests.exceptions import ConnectionError
 
 import astropy.units as u
+from alerce.core import Alerce
 from astropy.coordinates import SkyCoord
 from astropy.table import Table, join, join_skycoord, unique
 from astroquery.ipac.ned import Ned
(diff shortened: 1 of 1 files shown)


light_curves/code_src/sample_selection.py
```diff
@@ -316,6 +318,30 @@ def get_graham_sample(coords, labels, *, verbose=1):
...
+        journal articles associated with the target coordinates
+    object_ids: list of strings
+        eg., [ "ZTF18accqogs", "ZTF19aakyhxi", "ZTF19abyylzv", "ZTF19acyfpno"]
+    verbose: int
+        print out debugging info (1) or not(0)
+    """
+    alerce = Alerce()
+    objects = alerce.query_objects(oid=object_ids, format="pandas")
+    tde_coords = [SkyCoord(ra, dec, frame='icrs', unit='deg') for ra, dec in zip(objects['meanra'], objects['meandec'])]
+    tde_labels = ['ZTF-Objname' for _ in objects['meanra']]
```
jkrick, line 338: @xoubish these presumably TDEs are coming in with label = 'ZTF..'.  Should they have a label which includes the word 'TDE', or should we rename the function to be something broader like 'get_ZTF_objname', since the function looks like it doesn't know anything about what sort of objects these are.

light_curves/code_src/sample_selection.py
```diff
@@ -374,15 +404,62 @@ def get_paper_sample(coords, labels, *, paper_link="2019A&A...627A..33D", label=
         print("number of sources added from "+str(label)+" :"+str(len(paper_coords)))
 
 
-def clean_sample(coords_list, labels_list, verbose=1):
-    """Makes a unique sample of skycoords and labels with no repeats. Attaches an object ID to the coords.
+def get_papers_list_sample(coords, labels, *, paper_kwargs=[dict(),]):
```
jkrick, line 407: @troyraen I thought you once suggested that I not write 1 or 2 liner functions.  I don't care at all here, but am trying to learn from you.  So what is your logic in making this a function?

light_curves/code_src/sample_selection.py
xoubish: Yes, a general name would be better.

light_curves/code_src/sample_selection.py
troyraen: This makes it a lot easier for #215 to gather samples from more than one paper. Without this, `get_paper_sample` would require special handling in #215 because the code would have to check whether the user wanted more than one paper, and then call `get_paper_sample` multiple times. It's fewer lines of code to add this function than it would be to do the special handling.

(Shooby had asked me for a sample including multiple papers, so I had to think about the best way to do this. Otherwise, I probably wouldn't have noticed or spent any time on it.)

troyraen: But, there's no great reason for `get_papers_list_sample` and `get_paper_sample` to be two different functions. There, I was just choosing the option that was least intrusive to the existing code. But they are both pretty simple and could be combined, if you want?

jkrick: No need to combine the functions

jkrick (approved):

PR #234, opened 2024-02-29
Add redshifts to `sample_table`

troyraen:
This PR updates `clean_sample` to create a column for redshifts if any are included in `coords_list`. 

**Any `get_*_sample` function can now be updated to include redshifts for `sample_table`. To do this, the function should return a `coords_list` that contains a tuple of (skycoord, redshift) for each object instead of just the skycoord.**

This PR also updates `get_sdss_sample`and `get_csv_sample` to include the redshifts.

### DIFF
--- a/light_curves/code_src/sample_selection.py
+++ b/light_curves/code_src/sample_selection.py
@@ -380,7 +380,7 @@ def get_sdss_sample(coords, labels, *, num=10, zmin=0, zmax=10, randomize_z=Fals
         SDSS_coords = [SkyCoord(ra, dec, frame='icrs', unit='deg') for ra, dec in zip(res['ra'], res['dec'])]
         SDSS_labels = ['SDSS' for ra in res['ra']]
 
-        coords.extend(SDSS_coords)
+        coords.extend(zip(SDSS_coords, res['z']))  # add tuple (skycoord, redshift) for each object
         labels.extend(SDSS_labels)
 
     if verbose:
@@ -430,7 +430,8 @@ def get_papers_list_sample(coords, labels, *, paper_kwargs=[dict(),]):
         get_paper_sample(coords, labels, **kwargs)
 
 
-def get_csv_sample(coords, labels, *, csv_path, label, ra_colname="ra", dec_colname="dec", frame="icrs", unit="deg"):
+def get_csv_sample(coords, labels, *, csv_path, label,
+                   ra_colname="ra", dec_colname="dec", z_colname=None, frame="icrs", unit="deg"):
     """Loads coordinates from file at `csv_path` and adds them to `coords` and `lables`
 
     Parameters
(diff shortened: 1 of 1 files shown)


light_curves/code_src/sample_selection.py
```diff
@@ -378,9 +378,9 @@ def get_sdss_sample(coords, labels, *, num=10, zmin=0, zmax=10, randomize_z=Fals
         res = SDSS.query_sql(query, data_release = 16)
 
         SDSS_coords = [SkyCoord(ra, dec, frame='icrs', unit='deg') for ra, dec in zip(res['ra'], res['dec'])]
-        SDSS_labels = ['SDSS' for ra in res['ra']]
+        SDSS_labels = ['SDSS_QSO' for ra in res['ra']]
 
-        coords.extend(SDSS_coords)
+        coords.extend(zip(SDSS_coords, res['z']))  # add tuple (skycoord, redshift) for each object
```
bsipocz, line 383: At some point, I lost track that we're still mutating the input lists rather than properly returning a new one? This is a rather unexpected/dangerous behavior and if it's not too late would suggest refactoring.

bsipocz (commented):
I would strongly suggest of avoiding mutating input parameters in a function rather than returning an output. 

Also, the changing to use (coordinate, redshift) tuples in one of the functions but not in the others could cause some surprises, and would instead suggest to handle them separately. Maybe return everything in a Table or dataframe which could have different columns (coordinate and label for most data services, but coordinate, redshift, and label for others)

light_curves/code_src/sample_selection.py
troyraen: > At some point, I lost track that we're still mutating the input lists rather than properly returning a new one?

Yes, all of the `get_*_sample` functions do this. This PR doesn't change that in any way. I agree that a refactor would would be great, but it has never risen to a high priority for me.

bsipocz: yes, I suppose we missed all the functions in `sample_selection` when we did #156 and #171

jkrick: I agree that changing all the get_*_sample functions at this point is not warranted.

jkrick (approved):

troyraen:
> Also, the changing to use (coordinate, redshift) tuples in one of the functions but not in the others could cause some surprises, and would instead suggest to handle them separately. Maybe return everything in a Table or dataframe which could have different columns (coordinate and label for most data services, but coordinate, redshift, and label for others)

@bsipocz Is there anywhere in particular that you're worried about surprises popping up? As far as I know, the `clean_sample` function is the only thing using the results of the `get_*_sample` functions. Also, the user essentially _has_ to run `clean_sample` because that's what turns the lists in to a table for input to the archive calls.

I realize that surprises may pop up when adding new code to this section, but hopefully that's done now. This was supposed to be just a quick PR adding functionality that is actively being used. If you think it really does need to be handled more robustly to the point of needing to refactor all the other get-sample functions, I'm guessing Jessica would prefer that we don't make this change at all.

bsipocz:
Overall I feel that the API considerations are important (while see why it's frustrating to keep refactoring things). 
Here using lists and tuples of lists is an oversight from the previous refactor PRs and all the concerns I raised that prompted those original PRs there still stand: most importantly that changing a mutable input parameter without an explicit return should be avoided, as well as using the same call signature but with significantly different expectation depending on which functions there are used (e.g. list vs tuple of lists). So, I think it's worth finishing up the refactoring and working with a table input parameter here, basically the same we do elsewhere. Then extending that table with e.g. redshift, or anything else in the future (or by the user when they customise for their use case) all the while it ensures that the different lists and tuples don't get out of sync of each other.

troyraen:
I agree that we should try to follow all of those best practices, but I also understand the desire to not spend more time on this right now since it's no longer causing any actual problems. (Switching to use `sample_table` in the archive functions was very helpful and fixed the rest of the issues for downstream functions, as far as I can tell.) In any case, I have been asked not to refactor the `get_*_sample` functions so I won't unless that officially changes.

This PR was the simplest reasonable way I could think of to add the redshift. It does add some brittleness to the code, though I think not much more than was already here. But I understand still not wanting to add the complexity. It sounds like the best option is to close this PR without merging. @jkrick FYI.

Issue #240, opened 2024-03-11
GSoC2024: Dask Starter Problem


troyraen:
## Dask Starter Problem

This is a toy problem meant as a pre-application exercise for the GSOC project [Enable Dask execution of NASA time-domain analysis](https://openastronomy.org/gsoc/gsoc2024/#/projects?project=enable_dask_execution_of_nasa_time-domain_analysis). 

**Estimated completion time**: a few hours

Please complete all required tasks and whichever optional task(s) allows you to convey your thought process most easily. The goal is to see what your starting points are, not what you could do with several days of research and optimization.

### Overview

The `gsoc/2024/dask-toy-problem` branch contains a directory called `gsoc-dask-toy-problem/` with one file, `archives.py`. The file contains four public functions, `get_*_lightcurves`. These functions vary in their resource usage and runtimes. They take at least one argument, `num_sample`, which may vary between about 5 and 500,000.

The task is to parallelize these functions using Dask so that they run efficiently for large sample sizes. A basic script to run the functions serially looks like this:

```python
import archives  # be sure to import this **first**
import pandas as pd

num_sample = 100  # may vary between about 5 and 500,000

gaia_df = archives.get_gaia_lightcurves(num_sample)
heasarc_df = archives.get_heasarc_lightcurves(num_sample)
wise_df = archives.get_wise_lightcurves(num_sample)
ztf_df = archives.get_ztf_lightcurves(num_sample)

lightcurves_df = pd.concat([gaia_df, heasarc_df, wise_df, ztf_df])
```

### Instructions

1. Clone this repo and check out the branch `gsoc/2024/dask-toy-problem`.
2. Write code.
    - _Required_: Start a Dask cluster and stop it when finished. A local cluster is sufficient.
    - _Optional_: Execute the four `archives` functions and concatenate the results into a single Pandas DataFrame. Use any Dask method(s) you want. Your code should complete at least as fast as the serial script above. You can look at the code in archives.py, but do not alter it.
3. _Optional_: Write text. 300 words max, included as a '.md' file.
    - For any code that you did not write but would if you had more time, write down what you would do.
    - Write down any questions you have that, if answered, might help guide your design choices.
    - Test your code to determine how it scales as `num_sample` increases. Write down the results, your interpretation of the results, and/or what you would try next to see if it improves your code.
[...]

PR #246, opened 2024-03-15
GSoC2024 Solution Dask Starter Problem

lucasmg18:
#  GSoC2024 Solution Dask Starter Problem
Here I present the solution I implemented for this Dask Starter Problem (GSoC2024).

## Solution Description
As my solution, I implemented a function which adds to the new cluster client the functions to execute, with the number of samples introduced to the function and executes them. Then it gathers all the results and concatenates them into a single Pandas DataFrame

## Results

After executing the notebook created, the times obtained where this:

- Serial:
>Time for Serial execution with num_samples = 5:  14.392019033432007
Time for Serial execution with num_samples = 100:  252.61303806304932

- Parallel:
>Time for Serial execution with num_samples = 5:  10.859020471572876
Time for Serial execution with num_samples = 100:  201.532731034003235

## Future Improvements
With more time, I would evaluate each of the functions that are needed to be executed and see how much time each one consumes, because if one of them takes considerably more time than the others, the total time of parallel execution would be determined by this function, and maybe looking into the archive.py functions, another type of parallelization could be done to execute the slowest function itself and save execution time.

Issue number: #240


jkrick:

Issue #243, opened 2024-03-12
GSoC2024: DL Starter Problem


xoubish:
## GSOC2024 ML/DL Starter Problem

This is a simple toy problem meant as a pre-application exercise for the GSOC2024 project [Astronomical data enhancement with DL](https://openastronomy.org/gsoc/gsoc2024/#/projects?project=astronomical_data_enhancement_with_dl). 

**Estimated completion time**: a few hours

Please complete all required tasks and whichever optional task(s) allows you to convey your thought process most easily. The goal is to see what your starting points are, not what you could do with several days of research and optimization.

### Overview

Here we will have a more simplified case of the actual project with no time information. We have measurements of galaxies in different wavelengthts (i.e., broadband filters) in five different fields and the task is to bring them all onto a same wavelength footing. A simple notebook to read the galaxy data in the initial filters is in this repository (gsoc-ML-exercise/ReadCandels.md).

### Instructions

1. Clone this repo and checkout the branch gsco-ML-exercise.
2. Write code.
    - _Required_: A simple way to combine all five fields in optical and NIR filters and output one file in the requested wavelengths.
    - _Optional_: Use ML or DL to do this combination.
    - _Optional_: Add plots to show what you did makes sense.
    - _Optional_: Use prior information in ML by grouping galaxies at similar redshifts/stellarmasses/ etc. which are columns in the initial catalogs.
3. _Required_: Write text. 300 words max, included as a '.md' file.
    - Explain what you did and how you approached the problem.
    - For any code that you did not write but would if you had more time, write down what you would do.
4. _Required_: Open a PR with your code and writeup to merge to the main branch. Have GSoC2024 in the title of your PR.

If you have a question, please ask in a comment on this issue.

Shashankss1205:
Hello @xoubish,
I was confused whether we have to somehow train the data on the centerwave dictionary(which is extremely small in length) or on the complete dictionary corresponding to each field?

Thank you!
Shashank Shekhar Singh

Shashankss1205:
> @Shashankss1205 Sorry if it was not clear enough. If you open the notebook you will see how I read in five different datasets which correspond to five different parts in the sky. These fields do not necessarily have observations in the exact same filters (i.e., same wavelength). The toy problem is to bring them all into a same footing which is also written in the notebook. Hope this helps.

Okay now I got a clear idea of the problem statement. Thank you @xoubish for your response.
Will send my implementation as a PR in an hour.

[...]

PR #247, opened 2024-03-17
GSoC2024 DL toy problem solution

abed11326:
# Read Candels Solution
To solve the problem, I used a simple linear iterpolation. First, iterate over the input values, for a given a galaxy, extract positive fluxes and the corresponding wavelengths. The interpolation is applied for every galaxy to predict the fluxes of the required wavelengths given the fluxes of the input wavelenghts. Then, all the data is stored in a single dictionary that maps the field to a 2D array of galaxies and predicted fluxes. If I had more time, I would train a DL model to perform the task. The DL model can either be a regression model that predicts a numerical flux value of a required wavelength given fluxes of specific wavelengths as input, or the it can be a Density estimation model that builds a probability distribution and sample from it the flux values. 
# Plot
To visualize what I did, I re-used the visualization code of the original notebook and added the predicted fluxes of the required wavelengths for a random galaxy in the GOODS-S field as shown below.
plot
For issue #243


jkrick:

PR #250, opened 2024-03-19
DOC: Adding deployed URL to main README

bsipocz:
Also fixed a few typos

To close https://github.com/fornax-navo/fornax-demo-notebooks/issues/236

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,19 @@
 ## Executive Summary
+
 The Fornax Initiative is a NASA Astrophysics Archives project to collaboratively among the three archives HEASARC, IRSA, and MAST, create cloud systems, cloud software, and cloud standards for the astronomical community.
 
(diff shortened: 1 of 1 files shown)


bsipocz:
OK, so CI now works on PRs. It still make no sense to run it for PRs where we don't yet render the notebooks, but this should be nice to have for the cases where we already deployed the notebook. 

Maybe we should have another version where everything is included, even the WIP ones?

jkrick:
I'm not sure I understand the question.  Are you asking if we should implement the CI on the work in progress notebooks?  I don't think we need that.

I do think it would be nice to have something like a rendered version of the work in progress notebooks?  We have talked about how it would be useful to have fully worked versions somewhere including the final plots etc., for all notebooks, deployed and works in progress.

jkrick (approved):
looks good to me.

PR #254, opened 2024-03-21
GSoC2024 DL toy problem

VladZenko:
Added my solution to the DL toy problem + work outline file


jkrick:

PR #266, opened 2024-04-06
Update toctree and add links to readme

troyraen:
Simple PR to update the README:
- add notebooks to the toctree
- add links to notebook names in the text


bsipocz (approved):
Great that we ship these notebooks now, too!

PR #268, opened 2024-04-10
ML_AGNzoo suggestions

troyraen:
This is a partial review of ML_AGNzoo.md.

## Suggested changes

- remove unused imports
- remove the `warnings` filter
- set the `logging` level to `WARNING`
- consolidate duplicate code

## `logging` and `warnings`

Silencing all warnings hides important information about things that may be going wrong. I see that there are a large number of log messages being generated in the notebook that are not warnings but are more like info messages and lower. I think this is happening because `sompy` sets logging levels to `NOTSET` and `DEBUG`, which is very noisy. I think a good solution is to reset the logging level to the default `WARNING`, which is what I've done here. This silences most of the log messages without hiding the warnings.

With the logging level set to `WARNING`, several cells are now showing warnings like `invalid value encountered in divide`. I recommend checking each of these since they indicate there may be a problem with the math.

## Consolidate code

There is quite a bit of duplicated code in this notebook, mostly for plotting. I picked one cell (which plots ZTF data) and consolidated the code to show an example of how it can be done. It would be nice if the rest of the code were similarly consolidated. Even if we don't plan to do much more work on this notebook, I think the futures users would appreciate it. It will make the notebook easier to read and much easier to change (which I think we should expect that users will do) without introducing bugs.

### DIFF
--- a/light_curves/ML_AGNzoo.md
+++ b/light_curves/ML_AGNzoo.md
@@ -72,9 +72,6 @@ import logging
 logger = logging.getLogger()
 logger.setLevel(logging.ERROR)
 
-import warnings
-warnings.filterwarnings('ignore')
-
 plt.style.use('bmh')
 colors = [
     "#3F51B5",  # Ultramarine Blue
--- a/light_curves/ML_AGNzoo.md
+++ b/light_curves/ML_AGNzoo.md
@@ -218,35 +218,21 @@ The combination of the tree bands into one longer arrays in order of increasing 
 
 ```{code-cell} ipython3
 r = np.random.randint(np.shape(dat)[1])
-plt.figure(figsize=(18,4))
-plt.subplot(1,3,1)
 
-for i,l in enumerate(bands_inlc):
+_, axs = plt.subplots(1, 3, figsize=(18, 4))
+ztf_data = [dat_notnormal, dat, datm]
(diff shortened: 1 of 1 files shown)


xoubish:
Thanks Troy. All looks good to me. One reason for the repetitions in plotting was if people wanted to use parts of the code standalone, they could without the need to refer too much to previous cells. I also added gdown which skips the google virus scanner you mentioned.

troyraen:
The code within each cell could be consolidated independently from the other cells.

light_curves/code_src/ML_utils.py
troyraen, line 1: @xoubish did you mean to change this file?

xoubish: Yes, I removed numba. It was complaining about numpy version and there was a mismatch. And numba was not necessary.

light_curves/code_src/ML_utils.py
```diff
@@ -2,7 +2,7 @@
 from scipy import stats
 from scipy import interpolate
 import matplotlib.pyplot as plt
-import numba
+#import numba
```
bsipocz, line 5: Generally, do remove lines that are not working (or not used) rather than comment them out. I would think the only exception to this rule of thumb is the code snippets that are around to show how to scale up the same thing for a larger sample, or how to make a temporary data dump, etc.

PR #269, opened 2024-04-10
Spectroscopy faisst jan1624

afaisst:
Add the new use case spectroscopy notebook to the main.

Change includes a working version of the spectroscopy generator (only outline in the current main branch).

### DIFF
--- a/spectroscopy/spectra_generator.md
+++ b/spectroscopy/spectra_generator.md
@@ -49,7 +49,8 @@ The notebook may focus on the COSMOS field for now, which has a large overlap of
 | MAST    | HST*         | Slitless spectra would need reduction and extraction. There are some reduced slit spectra from COS in the Hubble Archive | `astroquery.mast`? | Implemented using `astroquery.mast` |
 | MAST    | JWST*        | Reduced slit MSA spectra that can be queried | `astroquery.mast`? | Should be straight forward using `astroquery.mast` |
 | SDSS    | SDSS optical| Optical spectra that are reduced | [Sky Server](https://skyserver.sdss.org/dr18/SearchTools) or `astroquery.sdss` (preferred) | Implemented using `astroquery.sdss`. |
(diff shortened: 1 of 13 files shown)


spectroscopy/spectra_generator.md
```diff
@@ -76,22 +77,189 @@ Andreas Faisst, Jessica Krick, Shoubaneh Hemmati, Troy Raen, Brigitta Sipőcz, D
...
+```
+
+### 1.2 Write out your sample to disk
+
+At this point you may wish to write out your sample to disk and reuse that in future work sessions, instead of creating it from scratch again.
+
+For the format of the save file, we would suggest to choose from various formats that fully support astropy objects(eg., SkyCoord).  One example that works is Enhanced Character-Separated Values or ['ecsv'](https://docs.astropy.org/en/stable/io/ascii/ecsv.html)
+
+```python
+sample_table.write('data/input_sample.ecsv', format='ascii.ecsv', overwrite = True)
```
jkrick, line 162: This line gives me errors on the first run through, but on a second run it completes.  Now I can't reproduce, but the error implied to me that the 'data' directory doesn't exist before this line.

spectroscopy/spectra_generator.md
```diff
@@ -76,22 +77,189 @@ Andreas Faisst, Jessica Krick, Shoubaneh Hemmati, Troy Raen, Brigitta Sipőcz, D
...
+
+```python
+%%time
+## Get DESI and BOSS spectra with SPARCL
+df_spec_DESIBOSS = DESIBOSS_get_spec(sample_table, search_radius_arcsec=5)
+df_spec.append(df_spec_DESIBOSS)
+```
+
+```python
+df_spec.data
```
jkrick, line 248: So I can show off the data structure a little more cleanly, I suggest:
```suggestion
df_spec.data.sort_values(by = ['objectid'])
```

jkrick (approved):
Two minor comments that will help me get the demo going.

spectroscopy/spectra_generator.md
afaisst: yes, that's correct. "data/" needs to exist.

PR #275, opened 2024-04-19
outline of new notebook

jkrick:
Draft outline of new notebook to explore Euclid data based on Vandana's outline from fornax-sciplatform slack.

### DIFF
--- a/spectroscopy/explore_Euclid_data.md
+++ b/spectroscopy/explore_Euclid_data.md
@@ -1,0 +1,72 @@
+<!-- #region -->
+# Explore Euclid Data
+***
+
+## Learning Goals    
+By the end of this tutorial, you will be able to:
+
+ &bull; work with data in the cloud
+
+ &bull; download and visualize Euclid images
+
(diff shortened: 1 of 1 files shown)


vandesai1:
According to Anastasia, there will be two ways to get a Euclid spectrum:

1. Do a TAP search on the MER photometry catalog. The results will include a service descriptor to get the spectrum for rows of interest.
2. Query the Euclid datalink service directly for spectra.

Currently, there is a spectrum datalink service, but Anastasia feels it is not quite ready to play with. Maybe in a couple weeks.

Depending on our schedule, we can wait for the IRSA service to firm up, OR we could choose to learn to do these kinds of searches using similar services. Anastasia says that CADC runs almost entirely off datalink - their SIA and ObsTAP services return datalink tables that you have to follow to get to any of their products.

This dealing with datalink to get data is entirely new to me and the part I am most worried about. I don't know what the ramifications are for pyvo or astroquery, if any. So regardless of the input sample or even the dataset, I think this is the technical part that we should try to understand first.

Issue #283, opened 2024-06-13
Test plotting of Herschel data after PR #281 closes


jkrick:
plotting of herschel_get_spec data was working but throwing errors (RuntimeWarning: Mean of empty slice) and not tested on the newest version of the plotting function. Testing needs to be done to make sure herschel data is well integrated into the plotting.

afaisst:
Tested and it works.

Issue #282, opened 2024-06-13
fix open 'try except' call in herschel_get_spec function


jkrick:
I can't figure out how to catch the right exception in the Herschel_get_spec function.  When a target has many observations, about 20 minutes after starting, the ESA Herschel archive drops the connection, which throws about 5 different Errors. I spent a morning testing different except schemes, but couldn't figure out which exception to catch to make the code run to completion even if the archive failed.  So, my temporary fix was to have an open 'except' clause.  I know this is not the best coding practice.  This issue is to re-visit if there is a way to clean up this open ended except.

troyraen:
I'm happy to help with stuff like this. I'll take a look when I review this notebook (July?) if no one else gets there first.

afaisst:
Hi Troy, 
if you want to fix this, please go ahead. The Herschel part of the spectroscopy notebook has now been merged to the main.

PR #284, opened 2024-06-13
Add herschel_get_spec functionality to spectroscopy notebook

jkrick:
This new module downloads and extracts "reduced" Herschel spectroscopy for a set of sources.  It is completely functional, but not perfect.  Description of imperfections (with short rants included) below:

1. Some decisions had to be made as this is a complicated dataset with complicated sets of rules about when a final spectrum is generated and which fluxes of the many reported are the best to use for any given target.  

   -  only observations which produce a HPSSPEC directory are considered.  There seems to be no way to ascertain this in the archive search, so the code downloads all tar files for all observations and then checks to see if this directory exists.
  
   - source resolution matters, ie., recommendations for which reduced spectrum to use are based on wether or not it is a point source or how extended it is. I have followed the advice in https://www.cosmos.esa.int/documents/12133/996891/Product+decision+trees to use the flux column which has the largest sum.  It is beyond the scope of this notebook to calculate if the sources is extended in Herschel band or not.

   - The Herschel directory tree in the downloaded tar files is extensive, and there is no way to customize and only download the one file that I need, so there is a lot of code around tracking filenames and paths and searching paths for filenames.

   - The Herschel archive breaks the connection at some point (testing on Arp220 which has many observations > 50, will hit this error).  I cannot for the life of me figure out how to catch the specific exceptions, and have spent too long trying to do this. So I know that blanket exceptions are not good practice, but they are the only way I know how to get this to work.  I have opened an issue on this  both here (#282 ) and at astroquery (to try to fix this at the source).

   - I have written into the function an option to delete tar files after the spectrum is extracted.  

2. I re-numbered the sections to account for the fact that Herschel data is being accessed through the ESA archive and not IRSA as was originally suggested in the outline.  The use of astroquery's ESA module  made this module possible in an automated fashion from python, so it is not possible to switch to IRSA.  IRSA only houses an API (which may even just access the ESA API???).  

3. It works with the plotting function before Andreas's newest PR #281 , but probably needs to be checked for compatibility with new function.   I have opened an issue for this (#283 ).

### DIFF
--- a/spectroscopy/code_src/herschel_functions.py
+++ b/spectroscopy/code_src/herschel_functions.py
@@ -1,0 +1,152 @@
+## Herschel PACS & SPIRE (from ESA TAP)
+from astroquery.esa.hsa import HSA
+from astroquery.exceptions import LoginError
+from requests.exceptions import ChunkedEncodingError
+import tarfile 
+from astropy.io import fits
+import glob
+import pandas as pd
+import astropy.constants as const
+from astropy import units as u
+import os
+
+from data_structures_spec import MultiIndexDFObject
+
+def find_max_flux_column(df):
+  """
+  Analyzes a DataFrame with flux columns and returns the column with the largest sum.
+
+  Args:
+      df (pandas.DataFrame): The DataFrame containing columns with "flux" in the name.
+
+  Returns:
+      str: The name of the column with the largest sum of values containing "flux".
+  """
+
+  # Filter column names containing "flux"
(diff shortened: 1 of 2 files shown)


afaisst:
Works. The only problem is that it takes ages (meaning hours) to get the Herschel data for only 5 sources.

afaisst (approved):
looks good!
I also checked that it works with the modified plotting code.

PR #286, opened 2024-07-16
MAINT: changing github org name

bsipocz:
I kept the licence file changes separate in case we want to deal with it differently.

I'm opening the PRs first and then change org, and thus I expect CI to not work properly until the actual name change is done.


bsipocz:
I'm going ahead and merge this now, if anyone sees any issue with the content of the PR, we should do a follow-up.

PR #289, opened 2024-07-24
commenting out herschel module

jkrick:
super small PR to
- complete the table at the top of the code to indicate the source of the herschel data
- comment out the lines in the herschel module which do the actual work.

In the SP meeting Jul/24 we decided that this module is functional but too slow to include.  We will leave it commented out in the notebook so that CI/CD can run, and people can use it if they like, but the whole notebook will not be slowed down by the herschel archive.

### DIFF
--- a/spectroscopy/spectra_generator.md
+++ b/spectroscopy/spectra_generator.md
@@ -241,6 +241,7 @@ df_spec.append(df_spec_HST)
 ### 2.3 ESA Archive
 ```python
 # Herschel PACS & SPIRE from ESA TAP using astroquery
+
 herschel_radius = 1.1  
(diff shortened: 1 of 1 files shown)


afaisst (approved):
perfect!

spectroscopy/spectra_generator.md
```diff
@@ -241,6 +241,7 @@ df_spec.append(df_spec_HST)
 ### 2.3 ESA Archive
 ```python
 # Herschel PACS & SPIRE from ESA TAP using astroquery
+
```
bsipocz, line 244: This PR seems to be empty, do you have one more commit locally?

PR #294, opened 2024-07-31
Catch specific exceptions in Herschel_get_spec

### LINKED ISSUE #282: fix open 'try except' call in herschel_get_spec function

troyraen:
Closes #282 

I added the specific exceptions I ran into so this is no longer a bare `except`.

### DIFF
--- a/spectroscopy/code_src/herschel_functions.py
+++ b/spectroscopy/code_src/herschel_functions.py
@@ -1,7 +1,7 @@
 ## Herschel PACS & SPIRE (from ESA TAP)
 from astroquery.esa.hsa import HSA
 from astroquery.exceptions import LoginError
-from requests.exceptions import ChunkedEncodingError
+from requests.exceptions import ChunkedEncodingError, ConnectionError
 import tarfile 
 from astropy.io import fits
 import glob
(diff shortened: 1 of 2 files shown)


jkrick (approved):
This works for me in the sense that the Herschel module runs to completion.   Thanks for looking into it.

Since I ran it just now, can you please edit the main notebook Herschel cell:
"#This search is fully functional, but is commented out because it is very slow." to "#This search is fully functional, but is commented out because it takes ~4 hours to run to completion"  I ran it on the ISP, so can't really say on Fornax, or specify the server, but I am imagining they will be similar since the data source is the same.  We can update this in the future if anyone runs it on Fornax.

PR #301, opened 2024-08-14
Add runtime & explanatory text to light_curve_classifier.md

### LINKED ISSUE #292: add more mark up around "read in a dataset of archival light curves"
a comment from Booz-Allen Hamilton:
"Looking at say ‘light_curve_classifier.md’ more mark-up around “Read in a dataset of archival light curves” and following sections to add explanations of what the code is doing, so that it may be easy for someone to alter the script for their own purposes."

jkrick:
Superquick PR 
Will close #292 

and I really thought there was an issue for adding runtime info to the notebooks, but can't find it at the momement.

### DIFF
--- a/light_curves/light_curve_classifier.md
+++ b/light_curves/light_curve_classifier.md
@@ -120,20 +120,27 @@ pd.options.mode.copy_on_write = True
 ```
 
 ## 1. Read in a dataset of archival light curves
+ We use here a sample of AGN including known CLAGN & random SDSS AGN
+ 
+ If you want to use your own sample, you can use the code [light_curve_generator.md](https://nasa-fornax.github.io/fornax-demo-notebooks/light_curves/light_curve_generator.html) in this same repo to make the required pandas dataframe which you will need to run this notebook. 
 
 ```{code-cell} ipython3
-#access structure of light curves made in the light curve generator notebook
-# has known CLAGN & random SDSS small sample of targets, all bands
+# First we want to load light curves made in the light curve generator notebook
+
+# The data is on google drive, this will download it for you and read it into 
+# a pandas dataframe
 savename_df_lc = './data/small_CLAGN_SDSS_df_lc.parquet'
 gdd.download_file_from_google_drive(file_id='1DrB-CWdBBBYuO0WzNnMl5uQnnckL7MWH',
                                     dest_path=savename_df_lc,
                                     unzip=True)
 
+#load the data into a pandas dataframe
 df_lc = pd.read_parquet(savename_df_lc)
 ```
 
 ```{code-cell} ipython3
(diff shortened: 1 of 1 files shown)


light_curves/light_curve_classifier.md
```diff
@@ -4,11 +4,11 @@ jupytext:
...
     format_version: 0.13
-    jupytext_version: 1.15.2
+    jupytext_version: 1.16.0
 kernelspec:
-  display_name: Python 3 (ipykernel)
+  display_name: science_demo
   language: python
-  name: python3
+  name: conda-env-science_demo-py
 ---
```
bsipocz, line 12: We'll need to find a way to not add changes to these metadata in unrelated PRs. E.g. changing the kernelname will certainly trip the CI in the future.

For now, my only solution is to `git add` patches rather than full files, either on the command line or with a GUI, but I'll google around if there is a nicer solution.

light_curves/light_curve_classifier.md
```diff
@@ -63,6 +63,9 @@ Trained classifiers as well as estimates of their accuracy and plots of confusio
 - `scipy` for statistical analysis
 - `json` for storing intermediate files
 - `google_drive_downloader` to access files stored in google drive
+
+## Runtime
+This notebook takes ~170s to run to completion on Fornax using the 'Astrophysics Default Image' and the 'Large' server with 16GB RAM/ 4CPU.
```
bsipocz, line 68: Maybe add as "As of 2024 Aug", or some other date indicator.

bsipocz (approved):
Looks good. Minor comment about adding some kind of timestamp for the timing info, but otherwise can be merged (even with the metadata changes, I'll the the problematic ones back when I run into issues).

light_curves/light_curve_classifier.md
jkrick: I don't know how to do this, so maybe it is something to add to your git demo for GRITS this fall.  Of course, you can always explain to me before then.  I am running these timing tests on Fornax, and the metadata changes are happening all behind the scenes.

light_curves/light_curve_classifier.md
jkrick: done

light_curves/light_curve_classifier.md
bsipocz: OK, I can certainly show how to use `git add --patch filename`, as it's a useful thing for many situations.

And also, I'm looking at whether removing all these metadata from the files is feasible. So merge away, and I'll follow-up separately.

bsipocz: xref: https://github.com/nasa-fornax/fornax-demo-notebooks/issues/303

PR #316, opened 2024-08-22
Fix path to nway.py in multiband_photometry

### LINKED ISSUE #313: nway.py call is not general
https://github.com/nasa-fornax/fornax-demo-notebooks/blob/fa9688e6631d9a5f59e13f6f3e1587dde8d83057/forced_photometry/multiband_photometry.md?plain=1#L793

The above line fails when nway is installed in the system. In that case `!nway.py ...` should work. A general solution would be to change the code to be: `!nway.py`, but add `/home/jovyan/.local/bin/nway.py` to $PATH, so if `nway` is installed in the user space it also works. i.e.:

Somewhere at the top:
```python
os.environ['PATH'] = f"/home/jovyan/.local:{os.environ['PATH']}"
```

Then change the above line to:
```python
!nway.py 'data/Chandra/COSMOS_chandra.fits' :ERROR_RADIUS 'data/multiband_phot.fits' 0.1 --out=data/Chandra/chandra_multiband.fits --radius 15 --prior-completeness 0.9
```

zoghbi-a:
Fixes #313

### DIFF
--- a/forced_photometry/multiband_photometry.md
+++ b/forced_photometry/multiband_photometry.md
@@ -115,6 +115,9 @@ except ImportError:
     print("tractor is missing")
     pass
 
+# Add ~/.local/bin to path so executables installed with pip are available (needed for nway)
+os.environ['PATH'] = f"{os.environ['HOME']}/.local/bin:{os.environ['PATH']}"
+
 
 %matplotlib inline
 ```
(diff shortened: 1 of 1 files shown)


jkrick:
> The content change looks all good, however I would ping @jkrick for a review here to see what she thinks about the best location for modifying the PATH.
> 
> (I would feel it could/should be done at the very top where we already add `code_src` to the path)

Thanks for asking, but I have no strong opinions on this, and would be happy to go with whatever you both feel is best.

bsipocz:
OK, then this is now in a conflict and have to be rebased anyway, can I ask you @zoghbi-a to move the PATH change up to the top, or let me know if I should go ahead and do it.

Thanks!

zoghbi-a:
I tried to resolve the conflict using github web editor (for the first time) and that seems to have messed things up. I'll go back to using in the cli.

zoghbi-a:
This is should be ready now!

bsipocz (approved):

bsipocz:
thank you!

PR #317, opened 2024-08-22
fix np.savez for the list of tuples in multiband_photometry

### LINKED ISSUE #312: forced photometry notebooks failing at np.savez
I am finding this cell failing at the `np.savez` line in the photometry notebook.

```python
    from  multiprocessing import Pool
    t0 = time.time()
    with open('output/output.log', 'w') as fp: fp.write('')
    with Pool() as pool:
        results = pool.map(calculate_flux, paramlist)
    dtime = time.time() - t0
    np.savez(fname, results=results)
    print(f'Parallel calculation took {dtime} seconds')
```
with the error:
```
--> 740     val = np.asanyarray(val)
    741     # always force zip64, gh-10776
    742     with zipf.open(fname, 'w', force_zip64=True) as fid:

ValueError: setting an array element with a sequence. The requested array has an inhomogeneous shape after 2 dimensions. The detected shape was (600, 2) + inhomogeneous part.
```

`results` is a list of tuples, and cannot be converted to a numpy array. A dtype may need to be passed explicitly, e.g.
```python
np.savez(fname, results=np.array(results, dtype=object))
```
Is this related to some numpy update?

I am using numpy 1.26.4

zoghbi-a:
Fixes #312

### DIFF
--- a/forced_photometry/multiband_photometry.md
+++ b/forced_photometry/multiband_photometry.md
@@ -598,7 +598,7 @@ else:
     with Pool() as pool:
         results = pool.map(calculate_flux, paramlist)
     dtime = time.time() - t0
-    np.savez(fname, results=results)
+    np.savez(fname, results=np.array(results, dtype=object))
     print(f'Parallel calculation took {dtime} seconds')
 ```
 


bsipocz:
Thank you!

PR #318, opened 2024-08-27
MAINT: Uniformise how to handle dependency install cell

bsipocz:
I opted to keep using the requirement files here, as opposed to list the packages in the pip install line as we already assume the users work from a cloned/full repo version for these notebooks (as opposed to the IRSA-tutorials that are expected to be used as self contained standalone notebooks).

Also, for easier book-keeping, I added one requirements file for each of the notebooks. They can still refer one another (see the one for scale_up), but this would make it easier in the CI and image generation to pick and choose dependency installs as needed.

My editor did a lot of whitespace cleanups, there is a `Hide whitespace` in the github diff to make it a cleaner review.


troyraen (commented):
I reviewed the changes in #321 since those changes were done on top of these.
I think these specific changes are working fine, though I found several requirements that I needed to add.

bsipocz:
> I think these specific changes are working fine, though I found several requirements that I needed to add.

I would say to go ahead with this then, I can then rebase the other PRs and they will become much cleaner in their diffs.

I also opened https://github.com/nasa-fornax/fornax-documentation/pull/43 that super briefly summarizes that the approach is to have a separate requirements file for each and every notebook, so they can be run separately if needed. From your review I see that I missed one undeclared `scikit-image`, but the other things are recovered.

PR #329, opened 2024-09-04
MAINT: update logo and favicon

bsipocz:
Tuned down the resolution of the original logo file, let me know if it was too much.


vandesai1:
Looks great! When the Fornax website is more developed, we can think about making the logo a link to it. But for now, I think we are good to go!

vandesai1 (approved):
Looks great!

bsipocz:
I'll iterate a bit more on the favicon as it looks really bad for me (too busy for the size of it). I'll either use a better resolution one, or reuse the old logo as it's less busy

vandesai1:
The favicon looks good to me!

vandesai1 (commented):
The favicon looks good!

Issue #311, opened 2024-08-20
MAINT: cleanup internal IRSA bucket usage


bsipocz:
- [ ] Spitzer COSMOS Images for the forced photometry notebook (see #121); change to use SEIP from the cloud, or use COSMOS from on-prem. This blocks testing of forced_photometry.md
- [x] #367 . This blocks testing of light_curve_generator.md.
- [x] unWISE data usage for the light curve notebook https://github.com/nasa-fornax/fornax-demo-notebooks/blob/main/light_curves/code_src/wise_functions.py#L104 (#330). This blocks testing of light_curve_generator.md.

troyraen:
Thanks for opening this. I added one more for the unWISE data in light curve notebook.

troyraen:
I have removed the ZTF DR18 Lightcurves from the bucket because the files are no longer usable by our code. See #371. I left the checkbox for ZTF data unchecked above because I am in the process of putting a HATS version of the ZTF DR23 Lightcurves in the testing bucket, so those will need to be cleaned up. Our proposal to get ODR space for ZTF catalogs has been approved, but there will be an interim period during which we'd like to have these files available in this testing bucket before making them publicly available in the ODR bucket.

Issue #141, opened 2023-09-13
Update the ZTF call to use a newer data release


troyraen:
The ZTF call uses DR18, which is a bit out of date. It also uses the mission-produced parquet files, which are organized in a way that is difficult for our use case. It would be nice to use a newer DR that is partitioned by HEALPix (either a single order or a hipscat variable order).

Currently, the DR18 files are in a temporary S3 bucket. The plan is to put them in a more permanent bucket in the AWS Open Data program. That would be a good time to make this switch.

jkrick:
This is superseded by #367 and #371 
I don't think we need three issues about this.

PR #333, opened 2024-09-16
Avoid hitting s3 bucket on module import

### LINKED ISSUE #332: BUG: ztf_functions.py should be importable even when there is no access permission to the s3 bucket
ztf_functions.py [accesses a remote bucket during import](https://github.com/nasa-fornax/fornax-demo-notebooks/blob/0434f1e25390a90f5544f3d39c3179f2ca5b0661/light_curves/code_src/ztf_functions.py#L22). This is probably not a good idea as it prone to failure when the permission to the bucket is not setup correctly.

troyraen:
Fixes #332 by lazy-loading the list of catalog files.


bsipocz (approved):

troyraen:
It would be better not to use a global variable but that's difficult with the current code. It could be fixed more easily using a HEALPix partitioned version of these files, so I'm punting that to #141.

bsipocz:
@troyraen - Any workaround is good for now, a better solution then can come with #141 and #311

PR #330, opened 2024-09-13
Change unWISE bucket to nasa-irsa-wise

troyraen:
The unWISE dataset accessed by the light curve notebooks is now in a NASA ODR bucket. This PR updates the code to access that bucket.

PR #334, opened 2024-09-17
Remove reference to incorrect AWS region

troyraen:
Removes a reference to an incorrect AWS region left over from #330


bsipocz (approved):

troyraen:
CI is failing on scale_up.md but this PR didn't touch that notebook (didn't even touch any code at all; only comments). Had a Slack conversation with @bsipocz. Will Merge now, and if it still fails on `main` we'll look into it further.

PR #337, opened 2024-09-18
BUG: fix notebook format to be myst markdown

bsipocz:
First step in the tech review: fix the format.

Note to all: we need the notebooks to be MyST markdown not vanilla markdown.

(I'm planning to open small, reasonably independent PRs, so they can be merged quickly and incrementally)


bsipocz:
This should be non-controversial, so I go ahead with the merge, and then open follow-up PRs for the requirements and rendering inclusion.

Issue #165, opened 2023-11-01
Notebooks: Speedup MAST PanSTARRS light curve search at scale


jkrick:
I have asked MAST helpdesk for help with how to run this search efficiently at scale

jkrick:
This should now be possible with a hipscat version of panstarrs in S3.  s3://stpubdata/panstarrs/ps1/public/hipscat/

The goal of this task is to make a standalone function which can query panstarrs for a large number of targets (500k) in a reasonable amount of time (< 1 day).

troyraen:
This lsdb (hipscat) tutorial looks directly relevant: https://lsdb.readthedocs.io/en/stable/tutorials/pre_executed/ztf_bts-ngc.html

troyraen:
If we don't want to use Dask here (which lsdb requires), we could still use the hipscat files with the same techniques we're using for the unWISE catalog in this notebook.

jkrick:
@troyraen Is there a reason not to use Dask?

troyraen:
I don't know exactly. We should probably try it and see what happens. It may not work well with the parallelized notebook cell.

jkrick:
I can confirm that running dask.distributed.client inside of multiprocessing crashes the kernel. I could see how this would cause problems if multiprocessing tries to make many distributed clients.  Instead of trouble shooting this part of this problem, I am going to cheat a tiny bit and move the panstarrs_get_lightcurves outside of the multiprocessing.pool call.  The dask client is already doing the multiprocessing after all, so the cell is still parallelizing the work.

troyraen:
> I can confirm that running dask.distributed.client inside of multiprocessing crashes the kernel.

Not unexpected but good to know for sure, thanks for trying it. 

> I am going to cheat a tiny bit and move the panstarrs_get_lightcurves outside of the multiprocessing.pool call. The dask client is already doing the multiprocessing after all, so the cell is still parallelizing the work.

Sounds like a good thing to try. If PanSTARRS/Dask works well well this way we might want to consider switching the other ones from multiprocessing to dask so that they can all run at the same time again.

troyraen:
WIP #344

jkrick:
[...]

PR #344, opened 2024-09-23
panstarrs_get_lightcurve using lsdb

jkrick:
having some trouble catching exceptions for FileNotFound in the call to lsdb's crossmatch with the Panstarrs catalog.  It would be nice if we could somehow catch this error so that the crossmatch doesn't fail if a file is missing.  Instead, it would be ideal if one of the objects fails the crossmatch, that the crossmatch still finishes with the other objects complete, and the final result just looks like no match was found for that trouble-causing object/file.  

Helpdesk request submitted to MAST 9/19 about the missing file.

### DIFF
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -9,7 +9,7 @@ from data_structures import MultiIndexDFObject
 
 
 #panstarrs light curves from hipscat catalog in S3 using lsdb
-def panstarrs_get_lightcurves(sample_table, radius):
+def panstarrs_get_lightcurves(sample_table, radius = 1):
     """Searches panstarrs hipscat files for light curves from a list of input coordinates.  
     
     Parameters
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -72,7 +72,6 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
         radius_arcsec=radius, 
         n_neighbors=1, 
         suffixes=("", "")
-
     )
         
     # plan to join that cross match with detections to get light-curves
@@ -91,7 +90,7 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
         #and the join with the detections table
         matched_df = matched_lc.compute()
 
-    #catch the case where there are no matches and return empty df
+    #handle the case where there are no matches and return empty df
     if len(matched_df["filterID"]) == 0:
         return MultiIndexDFObject(data=pd.DataFrame())
 
@@ -107,21 +106,17 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
     get_name_from_filter_id = np.vectorize(filter_id_to_name.get)
     filtername = get_name_from_filter_id(matched_df["filterID"])
       
-    # setup to build dataframe 
-    t_panstarrs = matched_df["obsTime"]
-    flux_panstarrs = matched_df['psfFlux']*1E3  # in mJy
-    err_panstarrs = matched_df['psfFluxErr'] *1E3
-    lab = matched_df['label']
-    objectid = matched_df['objectid']
-
     #make the dataframe of light curves
-    df_lc = pd.DataFrame(
-        dict(flux=pd.to_numeric(flux_panstarrs, errors='coerce').astype(np.float64), 
-             err=pd.to_numeric(err_panstarrs, errors='coerce').astype(np.float64), 
-             time=pd.to_numeric(t_panstarrs, errors='coerce').astype(np.float64), 
-             objectid=pd.to_numeric(objectid, errors='coerce').astype(np.int64), 
-             band=filtername, 
-             label=lab.astype(str))).set_index(["objectid","label", "band", "time"])
-
+    #the data conversions are to change from pyarrow datatypes to numpy datatypes
+    
+    df_lc = pd.DataFrame({
+        'flux': pd.to_numeric(matched_df['psfFlux'] * 1e3, errors='coerce').astype(np.float64),
+        'err': pd.to_numeric(matched_df['psfFluxErr'] * 1e3, errors='coerce').astype(np.float64),
+        'time': pd.to_numeric(matched_df['obsTime'], errors='coerce').astype(np.float64),
+        'objectid': matched_df['objectid'].astype(np.int64),
+        'band': filtername,
+        'label': matched_df['label'].astype(str)
+    }).set_index(["objectid", "label", "band", "time"])
+    
     return MultiIndexDFObject(data=df_lc)
     
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -8,8 +8,8 @@ from dask.distributed import Client
 from data_structures import MultiIndexDFObject
 
 
-#panstarrs light curves from hipscat catalog in S3 using lsdb
-def panstarrs_get_lightcurves(sample_table, radius = 1):
+# panstarrs light curves from hipscat catalog in S3 using lsdb
+def panstarrs_get_lightcurves(sample_table, *, radius=1):
     """Searches panstarrs hipscat files for light curves from a list of input coordinates.  
     
     Parameters
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -125,4 +125,5 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
              band=filtername, 
              label=lab.astype(str))).set_index(["objectid","label", "band", "time"])
 
-    return df_lc
+    return MultiIndexDFObject(data=df_lc)
+    
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -8,7 +8,7 @@ from dask.distributed import Client
 from data_structures import MultiIndexDFObject
 
 
-#panstarrs light curves from hipscat catalog using LSDB
+#panstarrs light curves from hipscat catalog in S3 using lsdb
 def panstarrs_get_lightcurves(sample_table, radius):
     """Searches panstarrs hipscat files for light curves from a list of input coordinates.  
     
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -69,7 +69,7 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
 
     #plan to cross match panstarrs object with my sample 
     #only keep the best match
-    yang_ps = panstarrs_object.crossmatch(
+    matched_objects = panstarrs_object.crossmatch(
         sample_lsdb, 
         radius_arcsec=radius, 
         n_neighbors=1, 
@@ -78,7 +78,7 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
     )
         
     # plan to join that cross match with detections to get light-curves
-    yang_ps_lc = yang_ps.join(
+    matched_lc = matched_objects.join(
         panstarrs_detect,
         left_on="objID",
         right_on="objID",
@@ -91,7 +91,7 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
     with Client():
         #compute the cross match with object table
         #and the join with the detections table
-        matched_df = yang_ps_lc.compute()
+        matched_df = matched_lc.compute()
     
         
     #cleanup the filter names to the expected letters
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -3,8 +3,6 @@ import pandas as pd
 from astropy.table import Table
 import lsdb
 from dask.distributed import Client
-
-
 from data_structures import MultiIndexDFObject
 
 
@@ -92,8 +90,11 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
         #compute the cross match with object table
         #and the join with the detections table
         matched_df = matched_lc.compute()
-    
-        
+
+    #catch the case where there are no matches and return empty df
+    if len(matched_df["filterID"]) == 0:
+        return MultiIndexDFObject(data=pd.DataFrame())
+
     #cleanup the filter names to the expected letters
     filter_id_to_name = {
     1: 'Pan-STARRS g',
@@ -102,13 +103,10 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
     4: 'Pan-STARRS z',
     5: 'Pan-STARRS y'
     }
-    if len(matched_df["filterID"]) > 0:
-        get_name_from_filter_id = np.vectorize(filter_id_to_name.get)
-        filtername = get_name_from_filter_id(matched_df["filterID"])
-    else:
-        # Handle the case where the array is empty
-        filtername = []
-    
+
+    get_name_from_filter_id = np.vectorize(filter_id_to_name.get)
+    filtername = get_name_from_filter_id(matched_df["filterID"])
+      
     # setup to build dataframe 
     t_panstarrs = matched_df["obsTime"]
     flux_panstarrs = matched_df['psfFlux']*1E3  # in mJy
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -25,7 +25,7 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
         the main data structure to store all light curves
     """
     
-    #read in the hipscat object table to lsdb
+    #read in the panstarrs object table to lsdb
     #this table will be used for cross matching with our sample's ra and decs
     #but does not have light curve information
     panstarrs_object = lsdb.read_hipscat(
@@ -37,7 +37,7 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
             "nStackDetections",  # some other data to use
         ]
     )
-    #read in the hipscat light curves to lsdb
+    #read in the panstarrs light curves to lsdb
     #panstarrs recommendation is not to index into this table with ra and dec
     #but to use object ids from the above object table
     panstarrs_detect = lsdb.read_hipscat(
@@ -46,7 +46,6 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
         columns=[
             "objID",  # PS1 object ID
             "detectID",  # PS1 detection ID
-            "ra", "dec",
             # light-curve stuff
             "obsTime", "filterID", "psfFlux", "psfFluxErr",
         ],
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -1,236 +1,16 @@
 import numpy as np
 import pandas as pd
-import requests
 from astropy.table import Table
-from tqdm.auto import tqdm
+import lsdb
+from dask.distributed import Client
+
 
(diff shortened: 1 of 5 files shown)


jkrick:
MAST helpdesk responded that they will fix the file problem on S3 before Oct 7, so this PR will wait for that fix to move forward, so that testing can be completed.

jkrick:
Now fully functional!
This PR includes:
- new panstarrs function which uses lsdb to access a hipscat parquet in S3
- removes the 'old' functions which use the panstarrs API
- changes name in plotting to be in line with actual name of the dataset "Pan-STARRS" (but is too hard for me to type every time)
- added runtime info to the intro cell (800s to run the full notebook) @bsipocz is this too long?

Notes
- along the way I discovered missing files in the panstarrs S3 parquets, MAST fixed those, and Troy assures me that he has checks for these things when he pushes data to S3.  So I think that is one problem that got fixed, that I hope we will not have to see in the future.
- This function is not actually faster for the sample size of 30 that we use in the notebook, but will be faster than the API for larger samples, and in general accessing data in parquet format in the cloud is the direction we want to be going.
- this might be the first calls to lsdb in our notebooks?  but probably not the last.

This will close #165 
This uses help from: https://github.com/astronomy-commons/lsdb/issues/416

bsipocz:
> 800s to run the full notebook

It's a bit long, but for fornax it's totally acceptable unless we hit a VM limit. We currently skip this notebook due to the ZTF private bucket access.

(and I don't know why we see the error with the classifier notebook, that should not be affected by this PR)

light_curves/requirements_light_curve_generator.txt
```diff
@@ -14,6 +14,9 @@ astroquery>=0.4.8.dev0
 acstools
 lightkurve
 alerce
+lsdb
+s3fs
```
bsipocz, line 18: We don't directly use s3fs so don't need to add it to this file.

Also do we have to manually install it as it's an undeclared dependency of lsdb? (would be nice to be able to pick it up automatically though an extras, although it's a theoretical ask for the lsdb developers as here we already install it as part of `pandas[aws]`

light_curves/requirements_light_curve_generator.txt
```diff
@@ -14,6 +14,9 @@ astroquery>=0.4.8.dev0
...
+nested_pandas
```
bsipocz, line 19: no need to declare it as we don't directly use it (and lsdb should install it if it requires it)

light_curves/light_curve_generator.md
```diff
@@ -131,9 +135,9 @@ get_yang_sample(coords, labels)   #2018ApJ...862..109Y
 # a balance between speed of running the light curves and whatever
 # the ML algorithms would like to have
 
-# num_normal_QSO = 5000
-# zmin, zmax = 0, 10
-# randomize_z = False
+#num_normal_QSO = 30
+#zmin, zmax = 0, 10
+#randomize_z = False
```
bsipocz, line 140: are these changes intended?

light_curves/light_curve_generator.md
```diff
@@ -84,6 +86,8 @@ import astropy.units as u
 import pandas as pd
 from astropy.table import Table
 
+import numpy as np
+
```
bsipocz, line 90: This is not used anywhere
```suggestion
```

light_curves/code_src/panstarrs_functions.py
```diff
@@ -1,236 +1,16 @@
...
-    
-    return(dresults)
+from data_structures import MultiIndexDFObject
 
 
-#Do a panstarrs search
-def panstarrs_get_lightcurves(sample_table, *, radius=1/3600):
-    """Searches panstarrs for light curves from a list of input coordinates.  This is the MAIN function.
+#panstarrs light curves from hipscat catalog in S3 using lsdb
+def panstarrs_get_lightcurves(sample_table, radius):
```
bsipocz, line 12: Please keep the API: radius should have a sensible default and then being an optional but mandatorily keyword argument. Or at least keep it as a mandatory keyword argument. (e.g. I want to avoid having functions that end up with a call signature of positional arguments like this `functionname(a, b, None, False, True, True)`)

light_curves/light_curve_generator.md
```diff
@@ -250,19 +258,35 @@ print('WISE search took:', time.time() - WISEstarttime, 's')
...
-The function to retrieve lightcurves from Pan-STARRS currently uses their API; based on this [example](https://ps1images.stsci.edu/ps1_dr2_api.html).  This search is not efficient at scale and we expect it to be replaced in the future.
+The function to retrieve lightcurves from Pan-STARRS uses a version of both the object and light curve catalogs that are stored in the cloud and accessd using [lsdb](https://docs.lsdb.io/en/stable/).  This function is efficient at large scale (sample sizes > ~1000).
 
 ```{code-cell} ipython3
 panstarrsstarttime = time.time()
 
-panstarrs_search_radius = 1.0/3600.0    # search radius = 1 arcsec
+panstarrs_search_radius = 1.0 # search radius in arcsec
 # get panstarrs light curves
-df_lc_panstarrs = panstarrs_get_lightcurves(sample_table, radius=panstarrs_search_radius)
```
bsipocz, line 260: keep this one. Being explicit, especially in a demo, is important.

light_curves/light_curve_generator.md
```diff
@@ -362,7 +386,7 @@ n_workers = 8
 heasarc_kwargs = dict(catalog_error_radii={"FERMIGTRIG": "1.0", "SAXGRBMGRB": "3.0"})
 ztf_kwargs = dict(nworkers=None)  # the internal parallelization is incompatible with Pool
 wise_kwargs = dict(radius=1.0, bandlist=['W1', 'W2'])
-panstarrs_kwargs = dict(radius=1.0/3600.0)
+panstarrs_search_radius = 1.0 # arcsec
```
bsipocz, line 389: nitpick, but I think it would ultimately be nice if these radiuses could be any astropy quantity rather than a float, of which the user should keep track if arcsec, or arcmin, or degrees, depending on which datasets they are querying. This comment likely will go beyond this PR into a follow-up

bsipocz (commented):
Some comments. Mostly minor, but I picked up on some things to make the API nicer (e.g. quantity inputs, or being explicit with parameters)

light_curves/light_curve_generator.md
jkrick: From my perspective, it doesn't matter what the values of these are since they are commented out.  I do change them when I am working on the code and testing different sample sizes, and then most often forget what they originally were and leave them at whatever the last value I tried.  Does that cause a problem?  Or is it the spaces between the # and the commands that is concerning?

light_curves/requirements_light_curve_generator.txt
jkrick: I removed s3fs from the requirements file.  I don't quite understand what the ask is for lsdb, so please open an issue over there if it is desirable.

light_curves/code_src/panstarrs_functions.py
jkrick: done

light_curves/light_curve_generator.md
jkrick: done

light_curves/light_curve_generator.md
jkrick: done

light_curves/light_curve_generator.md
bsipocz: It's the adding unrelated and unnecessary changes. I see why it happens, but it would be nice to figure out a convenient way to change the workflow so only intended and necessary changes get added. 
I added this topic to my list for the ipac visit.

light_curves/light_curve_generator.md
jkrick: I turned this into a new issue to track it, but not necessarily add it to this PR.

light_curves/requirements_light_curve_generator.txt
jkrick: I removed nested_pandas from the requirements file.

jkrick (commented):
Thanks for looking it over and the comments, I think I got them all.

light_curves/light_curve_generator.md
jkrick: yes!  I don't know how to do this, so perfect to discuss later this month.

light_curves/requirements_light_curve_generator.txt
```diff
@@ -14,6 +14,9 @@ astroquery>=0.4.8.dev0
 acstools
 lightkurve
 alerce
+lsdb
+#s3fs
+#nested_pandas
```
bsipocz, line 19: I'm sorry for nitpicking, but these need to be removed and not commented out. Rule of thumb, only leave commented out lines when they serve a clear purpose or a message to the user. Leaving in stuff that are not needed any more doesn't fall in that category.
```suggestion
lsdb
dask[distributed]
```

light_curves/light_curve_generator.md
```diff
@@ -250,19 +256,35 @@ print('WISE search took:', time.time() - WISEstarttime, 's')
...
+The function to retrieve lightcurves from Pan-STARRS uses a version of both the object and light curve catalogs that are stored in the cloud and accessd using [lsdb](https://docs.lsdb.io/en/stable/).  This function is efficient at large scale (sample sizes > ~1000).
 
 ```{code-cell} ipython3
 panstarrsstarttime = time.time()
 
-panstarrs_search_radius = 1.0/3600.0    # search radius = 1 arcsec
+panstarrs_search_radius = 1.0 # search radius = 1 arcsec
 # get panstarrs light curves
-df_lc_panstarrs = panstarrs_get_lightcurves(sample_table, radius=panstarrs_search_radius)
+df_lc_panstarrs = panstarrs_get_lightcurves(sample_table, radius = panstarrs_search_radius)
```
bsipocz, line 266:
```suggestion
df_lc_panstarrs = panstarrs_get_lightcurves(sample_table, radius=panstarrs_search_radius)
```

light_curves/light_curve_generator.md
```diff
@@ -131,9 +133,9 @@ get_yang_sample(coords, labels)   #2018ApJ...862..109Y
 # a balance between speed of running the light curves and whatever
 # the ML algorithms would like to have
 
-# num_normal_QSO = 5000
-# zmin, zmax = 0, 10
-# randomize_z = False
+#num_normal_QSO = 30
+#zmin, zmax = 0, 10
+#randomize_z = False
```
bsipocz, line 138:
```suggestion
# num_normal_QSO = 5000
# zmin, zmax = 0, 10
# randomize_z = False
```

light_curves/light_curve_generator.md
```diff
@@ -70,9 +72,9 @@ MAST, HEASARC, & IRSA Fornax teams
 
 This cell will install them if needed:
 
-```{code-cell} ipython3#
+```{code-cell} ipython3
 # Uncomment the next line to install dependencies if needed.
-# !pip install -r requirements_light_curve_generator.txt
+!pip install -r requirements_light_curve_generator.txt
```
bsipocz, line 77: Don't commit changes to this line.
```suggestion
# !pip install -r requirements_light_curve_generator.txt
```

light_curves/code_src/panstarrs_functions.py
```diff
@@ -1,236 +1,16 @@
...
-    return(dresults)
+from data_structures import MultiIndexDFObject
 
 
-#Do a panstarrs search
-def panstarrs_get_lightcurves(sample_table, *, radius=1/3600):
-    """Searches panstarrs for light curves from a list of input coordinates.  This is the MAIN function.
+#panstarrs light curves from hipscat catalog in S3 using lsdb
+def panstarrs_get_lightcurves(sample_table, radius = 1):
+    """Searches panstarrs hipscat files for light curves from a list of input coordinates.
```
bsipocz, line 13:
```suggestion
# panstarrs light curves from hipscat catalog in S3 using lsdb
def panstarrs_get_lightcurves(sample_table, *, radius=1):
    """Searches panstarrs hipscat files for light curves from a list of input coordinates.  
```

light_curves/code_src/panstarrs_functions.py
```diff
@@ -1,236 +1,16 @@
...
-    
-    """
-    dconstraints = {'objID': objid}
-    dcolumns = ("""objID,detectID,filterID,obsTime,ra,dec,psfFlux,psfFluxErr,psfMajorFWHM,psfMinorFWHM,
-            psfQfPerfect,apFlux,apFluxErr,infoFlag,infoFlag2,infoFlag3""").split(',')
-    # strip blanks and weed out blank and commented-out values
-    dcolumns = [x.strip() for x in dcolumns]
-    dcolumns = [x for x in dcolumns if x and not x.startswith('#')]
+import lsdb
+from dask.distributed import Client
```
bsipocz, line 5: dask is directly used, and thus has to be added to the requirements file.

light_curves/requirements_light_curve_generator.txt
bsipocz: xref: https://github.com/astronomy-commons/lsdb/issues/436

bsipocz (approved):
cleanup changes to be committed and then this is good to go.

light_curves/light_curve_generator.md
troyraen: > it would ultimately be nice if these radiuses could be any astropy quantity

Generally, I agree, but in this case (as I recall) the scale_up.md notebook needs these to be floats (or ints) because the functions are called from the command line. So if they get changed to astropy quantities we will need to implement a workaround in scale_up.md.

OTOH, we'll probably want to revisit the parallel cell in this notebook because this PR removes Pan-STARRS from the `multiprocessing` call so now it has to wait for the others to finish and then it runs separately. It would be nice to get them all to actually run in parallel again. Potentially, we can come up with a solution that actually works at scale, and then we wouldn't need scale_up.md anymore. The two biggest reasons we wrote scale_up.md[*] in the first place were that (1) the notebook would not continue running long enough for the parallel cell to complete at scale (servers were being shut down automatically) and (2) the ZTF function was taking the longest by far so we implemented `multiprocessing` directly in that function, but that could not be used in conjunction with the `multiprocessing` in the notebook's parallel cell. The first might be solved by infrastructure updates and the second by getting ZTF into HATS format -- if so, maybe we can switch `multiprocessing` -> `dask` and run all the calls in parallel with Dask directly from this notebook, at scale.

[*] scale_up.md describes the underlying bash code but doesn't actually execute it at scale -- the user is instructed to do that directly from the command line.

troyraen (approved):
Thanks for doing this, glad it's working! I left a mixture of comments below ranging from minor naming suggestions to code structure questions that will take a little more work to figure out. Anything you can't or don't want to address here can go into an issue and be assigned to me.

light_curves/light_curve_generator.md
troyraen: I've seen a new (to me anyway) "git" tab on the LHS of the Fornax console that gives GUI access to changed and staged files, etc. Maybe there's a way to select and commit individual lines/changes from there? I haven't played around with it at all but seems worth checking out.

bsipocz: Yeap, I put on the list for my visit to experiment something out.

light_curves/code_src/panstarrs_functions.py
```diff
@@ -244,53 +24,106 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1/3600):
...
+    objectid = matched_df['objectid']
+
+    #make the dataframe of light curves
+    df_lc = pd.DataFrame(
+        dict(flux=pd.to_numeric(flux_panstarrs, errors='coerce').astype(np.float64), 
+             err=pd.to_numeric(err_panstarrs, errors='coerce').astype(np.float64), 
+             time=pd.to_numeric(t_panstarrs, errors='coerce').astype(np.float64), 
+             objectid=pd.to_numeric(objectid, errors='coerce').astype(np.int64), 
+             band=filtername, 
+             label=lab.astype(str))).set_index(["objectid","label", "band", "time"])
```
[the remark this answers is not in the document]
jkrick: I changed the code to do all the steps within the dataframe assignment to hopefully mitigate this and make it more efficient.  I have not tested it with large samples to know if it is more efficient, but have tested it with small samples to know that it is still functional.

jkrick (commented):
I think I have addressed all the comments and made all the requested changes.  Let me know if I missed anything.

bsipocz:
Your comments and experience around working with the dtypes would be a super useful feedback to the lincc team I think.

light_curves/code_src/panstarrs_functions.py
```diff
@@ -244,53 +24,106 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1/3600):
...
+    err_panstarrs = matched_df['psfFluxErr'] *1E3
+    lab = matched_df['label']
+    objectid = matched_df['objectid']
+
+    #make the dataframe of light curves
+    df_lc = pd.DataFrame(
+        dict(flux=pd.to_numeric(flux_panstarrs, errors='coerce').astype(np.float64), 
+             err=pd.to_numeric(err_panstarrs, errors='coerce').astype(np.float64), 
+             time=pd.to_numeric(t_panstarrs, errors='coerce').astype(np.float64), 
+             objectid=pd.to_numeric(objectid, errors='coerce').astype(np.int64),
```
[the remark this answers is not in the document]
troyraen: Oh interesting. I wonder if it's fundamentally a result of choices made during the catalog import process rather than LSDB itself. I'll try to look at the data types more closely before I import another catalog to try to avoid these hurdles with the ones we put out.

troyraen:
LGTM. Thanks!

light_curves/code_src/panstarrs_functions.py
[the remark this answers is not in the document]
bsipocz: This difference between pyarrow/parquet vs pandas/numpy dtypes certainly came up before, and for end users I see all the reasons to provide the outputs in the latter, even if under the good it keeps mapping to what parquet likes.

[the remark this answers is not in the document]
troyraen: Good to know, thanks!

jkrick:
> Your comments and experience around working with the dtypes would be a super useful feedback to the lincc team I think.

message received, I put this in as feedback to lsdb in issue #441 over at astronomy-commons/lsdb

jkrick:
@bsipocz when you get a chance, can you please merge this PR?  I don't know why it is failing ci.

bsipocz:
It seems to be something unrelated, but I'm not sure I can dive into the details before I get back home.

jkrick:
I'm not in a hurry, thanks.

bsipocz:
OK, my latest commit should fix the failure. Basically adding some dask would trigger `sktime` to check if `dask[dataframe]` is all installed, and it wasn't.

bsipocz:
Hmm, apparently this is in fact an `sktime` issue: https://github.com/sktime/sktime/issues/7250. Someone reported that downgrading sktime solves the issue, so I was trying that. If it doesn't work, I'll look into more hackery.

bsipocz:
I'm not sure why the classifier notebook run into VM limits now, I'm doing some more debugs and see if it is still OK with GHA. If yes, then I'll bring over the configs from the IRSA notebooks to we can easily opt in GHA build here, too.

jkrick:
> see if it is still OK with GHA
@bsipocz what is "GHA"?

bsipocz:
GHA: GitHub Actions. We use GHA for doing the HTML rendering and hosting the pages with github. But for the PRs we use a different system, and their limitations differ a bit. It was already close to the limit but from below, I suspect adding the new dependencies may changed it worse the worse a bit, but we can't do much about it.

TL;DR, I may need to tweak the configs but I don't think there is much here to do about it. So I would merge this now and figure out issues with circleCI if/when they arise on main.

Issue #304, opened 2024-08-14
speedup ML_AGNzoo section 6


jkrick:
I'm running our notebooks on the new daskhub and simultaneously capturing runtimes to add to the text.  However this notebook takes a long time, which is also not good for the CI/CD process.  

Specifically Section 6 is taking a long time.  @xoubish can you suggest a way to speed up this part?  Otherwise, we could comment out the dtw_distance and make a comment to say that if someone wants to run it they can un-comment (this is what we have done for other notebooks).

PR #345, opened 2024-09-24
added runtime and commented out dtw lines

### LINKED ISSUE #324: ML_AGNzoo doesn't finish executing
I've had this notebook running for about an hour and it's still not done. It has been stuck on the second cell in section '4) Repeating the above, this time with ZTF + WISE manifold' for most of the time. It hasn't crashed (`top` shows that the CPU is still in use), though there are a bunch of warnings. I don't know whether this is normal/expected or not.

_Originally posted by @troyraen in https://github.com/nasa-fornax/fornax-demo-notebooks/pull/321#discussion_r1733620683_

jkrick:
small PR to cleanup a few issues. 
- added runtime and server size info directly under the intro
- commented out uses of dtw distance to make the notebook run in a reasonable amount of time.

This should close #324 and #304 

It also fulfills our desire to add runtime info to all the notebooks, although I can't find an issue for that right now.


light_curves/ML_AGNzoo.md
```diff
@@ -50,7 +51,7 @@ This cell will install them if needed:
 
 ```{code-cell} ipython3
-# !pip install -r requirements_ML_AGNzoo.txt
+#!pip install -r requirements_ML_AGNzoo.txt
```
bsipocz, line 54: Adding a linter is on the list, but in the meantime I'll add a commit that cleans up pep8 things.

(cc @troyraen @zoghbi-a - do you know a jupyterlab linter extension that you would recommend? I think it's time to bite the bullet and set up something (I don't want to go nuclear and black all the notebooks, but we could enable to fix some non-controversial things like comments and long lines, etc.; but that whole work only makes sense if we provide something that helps to write the code that way from the beginning rather than ending up going back and forth)

bsipocz (changes requested):
Your changes look good, but I'll block merging this for now with a TODO for myself to add the notebook back to the executed list.

light_curves/ML_AGNzoo.md
troyraen: > do you know a jupyterlab linter extension that you would recommend?

I don't. I usually work with notebooks as python files and lint before converting to markdown.

bsipocz:
@troyraen - now that I have added a lot of cleanup in here, could you please do the merge?

 (I really just broke up extra long lines and added some whitespace according to PEP8 as well as renamed one variable that overrode astropy units.)
 
 There seems to be a lot of code repetition for plotting, as well as non-descriptive variables all over.  Cleaning those up is a low-hanging fruit for a rainy day.

troyraen (approved):
LGTM

PR #354, opened 2024-12-03
Clean up formatting of spectra_generator.md and related .py files

troyraen:
Clean up formatting by:

- Organize imports and remove unused
- Run `autopep8` with a max line length of 100
- Manually wrap long lines of text in spectra_generator.md and comments in .py files


troyraen:
@bsipocz I ran `autopep8` to clean up the formatting. Does it look reasonable to you, or do you have other suggestions? The biggest remaining issue that I see is that some lines are still quite long. I'll point to one below. I will wrap long lines of text/comments manually (about to do that now) but I'm not planning to wrap lines of code manually. I tried using the `--aggressive` flag but that made it look really ugly in some spots, at least to my eye.

spectroscopy/code_src/herschel_functions.py
```diff
@@ -55,84 +57,89 @@ def Herschel_get_spec(sample_table, search_radius_arcsec, datadir,
...
- 
+
+        # first find the object ids from herschel then download the data for each observation id
+        # query_hsa_tap doesn't accept an upload_table, so do this so do this as a for loop over each instrument and object..
+
         for instrument_name in ['PACS', 'SPIRE']:
-            querystring = "select observation_id from hsa.v_active_observation join hsa.instrument using (instrument_oid) where contains(point('ICRS', hsa.v_active_observation.ra, hsa.v_active_observation.dec), circle('ICRS', "+str(search_coords.ra.deg)+", " + str(search_coords.dec.deg) +", " + str(search_radius_arcsec) +"))=1 and hsa.instrument.instrument_name='"+str(instrument_name)+"'"
+            querystring = "select observation_id from hsa.v_active_observation join hsa.instrument using (instrument_oid) where contains(point('ICRS', hsa.v_active_observation.ra, hsa.v_active_observation.dec), circle('ICRS', "+str(
+                search_coords.ra.deg)+", " + str(search_coords.dec.deg) + ", " + str(search_radius_arcsec) + "))=1 and hsa.instrument.instrument_name='"+str(instrument_name)+"'"
```
troyraen, line 80: Here's one example. Line 79 is still 232 characters long.

spectroscopy/code_src/herschel_functions.py
```diff
@@ -1,45 +1,48 @@
...
 from data_structures_spec import MultiIndexDFObject
 
+
 def find_max_flux_column(df):
-  """
-  Analyzes a DataFrame with flux columns and returns the column with the largest sum.
+    """
+    Analyzes a DataFrame with flux columns and returns the column with the largest sum.
+
+    Args:
```
bsipocz, line 21: we should probably also run some docstring formatting, e.g. these should be numpydoc formatted like the rest.
(beyond scope for this PR)

spectroscopy/code_src/herschel_functions.py
bsipocz: this is what you got with the auto formatted? I agree that it's still very ugly and not really an improvement compared to the one super long line.

spectroscopy/code_src/sample_selection.py
```diff
@@ -1,15 +1,13 @@
-import astropy.units as u
-import numpy as np
-from astropy.coordinates import SkyCoord
 from astropy.table import Table, join, join_skycoord, unique
-#from astroquery.ipac.ned import Ned
-#from astroquery.sdss import SDSS
-#from astroquery.simbad import Simbad
-#from astroquery.vizier import Vizier
+# from astroquery.ipac.ned import Ned
```
bsipocz, line 2: we'll need to clean up all the commented-out unused code, too.

spectroscopy/code_src/sdss_functions.py
```diff
@@ -32,40 +24,40 @@ def SDSS_get_spec(sample_table, search_radius_arcsec, data_release):
...
+                                         )).set_index(["objectid", "label", "filter", "mission"])
             df_spec.append(dfsingle)
 
         else:
             print("Source {} could not be found".format(stab["label"]))
 
-
-    return(df_spec)
\ No newline at end of file
+    return (df_spec)
```
bsipocz, line 63: not sure if we can clean all these up without using ruff or something similar. Just leaving the comment for the example
```suggestion
    return df_spec
```

spectroscopy/code_src/spitzer_functions.py
```diff
@@ -32,71 +27,81 @@ def SpitzerIRS_get_spec(sample_table, search_radius_arcsec , COMBINESPEC):
...
+
+        if len(tab) > 0:  # found a source
+
+            # If multiple entries are found, pick the closest.
             # Or should we take the average instead??
             if len(tab) > 0:
-                print("More than 1 entry found" , end="")
+                print("More than 1 entry found", end="")
                 if not COMBINESPEC:
                     print(" - pick the closest")
```
bsipocz, line 53: Beyond scope of this PR, but I noticed this:
in the lightcurves notebook/code we had a verbose option, so we don't get all the noise from these prints.

bsipocz (approved):
Looks good to me, and an imrpovement. While I left some comments, I would suggest to go ahead and merge this sooner rather than later, so any later cleanups won't be lost in the noise of these trivial changes.

bsipocz:
and we need a `skip ci` label, as there is no reason for run CI for these notebooks until they are actually opted into the rendering/CI

spectroscopy/code_src/herschel_functions.py
troyraen: I'm working on the docstrings now in a separate branch. If you know of a tool that will format them, let me know.

bsipocz: numpydoc comes with a linter/validator but it really lights up as a chrismass tree when I run it on the code directories here.

spectroscopy/code_src/herschel_functions.py
troyraen: Yes, just autopep8 so far, no manual adjustments.

spectroscopy/code_src/sdss_functions.py
troyraen: I'll take them out when fix the docstrings since I'll be checking the return objects then. This is much easier to do manually than, say, deciding where to break long lines of code.

troyraen:
Thanks @bsipocz. 

I'm merging this now. More PRs will follow to complete the tech review.

spectroscopy/code_src/spitzer_functions.py
troyraen: Most of the other spectroscopy modules have the verbose option. Not sure why it was left out here.

PR #356, opened 2024-12-04
Format docstrings in spectroscopy functions

troyraen:
- Format docstrings so they conform to numpydoc style.
- Remove references to light curves (these were introduced when copying code from the light curve generator).
- Remove parentheses around returned objects (fixes https://github.com/nasa-fornax/fornax-demo-notebooks/pull/354#discussion_r1868535320).  

I checked these with `numpydoc lint` and fixed most of the issues. I ignored a few, most notably "Summary should fit in a single line" to avoid making substantive changes to the text with this PR.


bsipocz (approved):
🎉

PR #357, opened 2024-12-05
Reduce indentation by inverting `if` conditions

troyraen:
Some of the spectroscopy functions have quite a few levels of indentation (almost 10 in some cases) due to nested `for` loops and `if` statements. This tends to squish the code into a narrow region on the right making it more difficult to read. This PR reduces some of the indentation by inverting the conditions of `if` statements where possible.

Simple example with just 2 levels (benefits are greater when there are more levels). This:
```python
for sample in sample_table:
    table = query_archive(...)
    if len(table) > 0:
        # long block of code
        # ...
    else:
        print("no results found")
```

becomes this:
```python
for sample in sample_table:
    table = query_archive(...)
    if len(table) == 0:
        print("no results found")
        continue
    # long block of code. now has more room horizontally.
    # ...
```

### DIFF
--- a/spectroscopy/code_src/spitzer_functions.py
+++ b/spectroscopy/code_src/spitzer_functions.py
@@ -42,65 +42,66 @@ def SpitzerIRS_get_spec(sample_table, search_radius_arcsec, COMBINESPEC):
                                 spatial="Cone", radius=search_radius_arcsec * u.arcsec)
         print("Number of entries found: {}".format(len(tab)))
 
-        if len(tab) > 0:  # found a source
+        if len(tab) == 0:
+            continue
 
-            # If multiple entries are found, pick the closest.
-            # Or should we take the average instead??
-            if len(tab) > 0:
-                print("More than 1 entry found", end="")
-                if not COMBINESPEC:
-                    print(" - pick the closest")
-                    sep = [search_coords.separation(SkyCoord(tt["ra"], tt["dec"], unit=u.deg, frame='icrs')).to(
-                        u.arcsecond).value for tt in tab]
-                    id_min = np.where(sep == np.nanmin(sep))[0]
-                    tab_final = tab[id_min]
-                else:
-                    print(" - Combine spectra")
-                    tab_final = tab.copy()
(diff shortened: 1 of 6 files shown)


bsipocz:
Some of these conditionals and loops are also very long, and for those the rule of thumb I was taught was to avoid any such section that is longer than one editor screen. (aka refactoring is advised if/when one ends in that situation)

troyraen:
@bsipocz I totally agree, just trying not to spend that much time on this code right now. The changes I made here were some low-hanging fruit that will make the rest of my review a little easier. This tends to happen a lot with these notebooks, and I think it would be good for us to come up with a small number of relatively simple things the scientists can think about / do to avoid ending up in this situation with future code.

bsipocz:
Yes, I totally meant it for that simple list of guidelines, rather than to fully rewrite of what we have. (OTOH, you practically did a couple of the refactorings in this PR, so we can even link to this as real life example of how to refactor a long conditional by e.g. inverting)

bsipocz (approved):
All looks correct, thanks!

spectroscopy/code_src/spitzer_functions.py
```diff
@@ -42,65 +42,66 @@ def SpitzerIRS_get_spec(sample_table, search_radius_arcsec, COMBINESPEC):
...
-                    tab_final = tab[id_min]
-                else:
-                    print(" - Combine spectra")
-                    tab_final = tab.copy()
+        if len(tab) == 0:
+            continue
+
+        # If multiple entries are found, pick the closest.
+        # Or should we take the average instead??
+        if len(tab) > 0:
```
troyraen, line 50: This code block (lines 50-62) is weird in two ways:
- It seems to think the condition was `if len(tab) > 1:` instead of `if len(tab) > 0:` (which would make more sense given that the `== 0` case was already handled).
- The resulting variable `tab_final` is not used. Maybe the following line of code was supposed to use it instead of `tab`?

The diff in this PR makes it difficult to look at, and I'd rather not make substantive changes in this PR anyway. Opening an issue to be addressed separately.

PR #361, opened 2024-12-05
Deal with warnings

troyraen:
Clean up warnings in the spectra_generator notebook:

- [x] MAST: `NoResultsWarning` -- Suppress these since we don't expect all-sky coverage and we explicitly handle the no-results cases.
- [x] Keck and SDSS: `RuntimeWarning: divide by zero` -- These occur in error calculations when the inverse variance is zero, resulting in infinite error. Leave this bad data in and ignore the warning.

### DIFF
--- a/spectroscopy/code_src/keck_functions.py
+++ b/spectroscopy/code_src/keck_functions.py
@@ -64,6 +64,7 @@ def KeckDEIMOS_get_spec(sample_table, search_radius_arcsec):
         # Prepare arrays
         wave = spec["LAMBDA"][0] * u.angstrom
         flux_cgs = spec["FLUX"][0] * 1e-17 * u.erg/u.second/u.centimeter**2/u.angstrom
+        # [FIXME] Next line raises 'RuntimeWarning: divide by zero encountered in divide'
         error_cgs = np.sqrt(1 / spec["IVAR"][0]) * 1e-17 * u.erg/u.second/u.centimeter**2/u.angstrom
 
         # Create MultiIndex object
--- a/spectroscopy/code_src/keck_functions.py
+++ b/spectroscopy/code_src/keck_functions.py
@@ -64,8 +64,10 @@ def KeckDEIMOS_get_spec(sample_table, search_radius_arcsec):
         # Prepare arrays
         wave = spec["LAMBDA"][0] * u.angstrom
         flux_cgs = spec["FLUX"][0] * 1e-17 * u.erg/u.second/u.centimeter**2/u.angstrom
-        # [FIXME] Next line raises 'RuntimeWarning: divide by zero encountered in divide'
-        error_cgs = np.sqrt(1 / spec["IVAR"][0]) * 1e-17 * u.erg/u.second/u.centimeter**2/u.angstrom
+        # Inverse variances may be zero, resulting in infinite error.
+        # We'll leave these in and ignore the "divide by zero" warning.
+        with np.errstate(divide='ignore'):
+            error_cgs = np.sqrt(1 / spec["IVAR"][0]) * flux_cgs.unit
 
         # Create MultiIndex object
         dfsingle = pd.DataFrame(dict(wave=[wave],
--- a/spectroscopy/code_src/mast_functions.py
+++ b/spectroscopy/code_src/mast_functions.py
@@ -1,10 +1,12 @@
 import io
 import os
 import shutil
+import warnings
 from contextlib import redirect_stdout
 
 import astropy.constants as const
 import astropy.units as u
+import astroquery.exceptions
 import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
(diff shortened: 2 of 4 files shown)


troyraen:
So far I've just added comments (including the full warning message) in the code where they're raised. I'd like feedback on how to deal with them. 
- `NoResultsWarning`: My sense is that the MAST queries are just being extra noisy we can safely suppress these, but I'm not sure.
- `RuntimeWarning: divide by zero`: Seems like we need to fix these. But as I recall, we ran into these with the light curves notebook and no one was concerned about it. So again, not sure.

@bsipocz and @jkrick what are your thoughts?

spectroscopy/code_src/mast_functions.py
```diff
@@ -336,6 +340,9 @@ def HST_get_spec(sample_table, search_radius_arcsec, datadir, verbose,
             continue
 
         # Download
+        # Observations.download_products prints an info message for every download, like:
```
bsipocz, line 343: There should be a kwarg in astroquery.mast to shut these up, if it's missing then please open an issue and we'll get it fixed.

spectroscopy/code_src/mast_functions.py
```diff
@@ -96,6 +96,8 @@ def JWST_get_spec_helper(sample_table, search_radius_arcsec, datadir, verbose,
 
         # Query results
         search_coords = stab["coord"]
+        # [FIXME] Next code line raises a warning when stab['label'] == 'NGC3049'.
+        # 'WARNING: NoResultsWarning: Query returned no results. [astroquery.mast.discovery_portal]'
```
bsipocz, line 100: We can filter these out as it's expected that not all input coordinate/objects will have coverage. But I would do that filter very explicitly with detailed comments, and here and only within this scope rather than in the notebook. As I would love to avoid teaching users to ignore warnings by default.

jkrick:
I'll defer to you both @bsipocz and @troyraen as to which things are important enough to fix and which things aren't really broken.  I know Brigitta doesn't want to teach us to ignore warnings, and while my codes may not do it explicitly, my brain certainly does ignore warnings in terms of skimming over them when looking for output.

bsipocz:
> my brain certainly does ignore warnings in terms of skimming over them when looking for output.

and thus we should reduce the noise in tutorial notebooks, and only expose the ones that a user should do something about, as sometimes they contain useful info :) (this is so much easier to say than to do though, I heavily rely on CI turning warning to exceptions at a lot of places, but that works much better in libraries than with notebooks)

troyraen:
The 'divide by zero' warnings are occurring in error calculations when the inverse variance is zero, so the error is infinite. I think we decided in the light curves notebook to leave bad data in. The plotting function in this notebook masks these bad data points, so there's no problem leaving them in from a coding standpoint. So I've used the `np.errstate` context manager to ignore the warnings.

spectroscopy/code_src/mast_functions.py
troyraen: I don't see a kwarg, so I'll open an issue.

troyraen: https://github.com/astropy/astroquery/issues/3146

troyraen: Ok, there's a kwarg `verbose=False` that will do this, but we need the latest "dev" version of astroquery (0.4.8.dev9474) to get it. Fornax has 0.4.7 which is the latest version available through pip unless you add the `--pre` flag. @bsipocz have we figured out how to handle these "dev" dependencies in requirements.txt files?
I see `astroquery>=0.4.8.dev0` in the light_curves requirements. Do I remember correctly that if there is any dev version already installed (like on the ISP) it won't be upgraded, but that's still the best option because something like `astroquery>=0.4.8.dev9474` won't work? A slightly older version (0.4.8.dev9306) throws an error, so we'd need more than "dev0" to ensure a working version.

bsipocz: I plan to do a release before the end of the year, so for now update your install manually. 
And in the new year astroquery will get a new packaging overhaul and a better system for frequent releases, so no workaround will be needed

troyraen:
@bsipocz I think this is ready to merge.

spectroscopy/code_src/keck_functions.py
```diff
@@ -64,7 +64,10 @@ def KeckDEIMOS_get_spec(sample_table, search_radius_arcsec):
         # Prepare arrays
         wave = spec["LAMBDA"][0] * u.angstrom
         flux_cgs = spec["FLUX"][0] * 1e-17 * u.erg/u.second/u.centimeter**2/u.angstrom
-        error_cgs = np.sqrt(1 / spec["IVAR"][0]) * 1e-17 * u.erg/u.second/u.centimeter**2/u.angstrom
+        # Inverse variances may be zero, resulting in infinite error.
+        # We'll leave these in and ignore the "divide by zero" warning.
+        with np.errstate(divide='ignore'):
+            error_cgs = np.sqrt(1 / spec["IVAR"][0]) * flux_cgs.unit
```
bsipocz, line 70: This maybe a good example for a docs of good practices, e.g. deal with expected Warnings that are due to the nature of the functionality (e.g. either this or the occasional no result warning coming from astroquery when we loop through a list of coordinates where we expect to have no matches!); and any warnings the end user should/can do nothing about.

bsipocz (approved):
Look good to me, thank you for doing this cleanup!

PR #362, opened 2024-12-08
Fix Spitzer logic

### LINKED ISSUE #358: `SpitzerIRS_get_spec` seems to have two bugs
From https://github.com/nasa-fornax/fornax-demo-notebooks/pull/357/files#r1870578195

This code block (lines 50-62) is weird in two ways:
1. It seems to think the condition was `if len(tab) > 1:` instead of `if len(tab) > 0:` (which would make more sense given that the `== 0` case was already handled).
2. The resulting variable `tab_final` is not used. Maybe the following line of code was supposed to use it instead of `tab`?

The diff in this PR makes it difficult to look at, and I'd rather not make substantive changes in this PR anyway. Opening an issue to be addressed separately.

troyraen:
Closes #358 

Includes two code changes, explained below.

### DIFF
--- a/spectroscopy/code_src/spitzer_functions.py
+++ b/spectroscopy/code_src/spitzer_functions.py
@@ -11,7 +11,7 @@ from data_structures_spec import MultiIndexDFObject
 
 def SpitzerIRS_get_spec(sample_table, search_radius_arcsec, COMBINESPEC):
     """
-    Retrieve HST spectra for a list of sources.
+    Retrieve Spitzer spectra for a list of sources.
 
     Parameters
     ----------
@@ -53,7 +53,7 @@ def SpitzerIRS_get_spec(sample_table, search_radius_arcsec, COMBINESPEC):
                 print(" - pick the closest")
                 sep = [search_coords.separation(SkyCoord(tt["ra"], tt["dec"], unit=u.deg, frame='icrs')).to(
                     u.arcsecond).value for tt in tab]
-                id_min = np.where(sep == np.nanmin(sep))[0]
+                id_min = np.nanargmin(sep)
                 tab_final = tab[id_min]
             else:
                 print(" - Combine spectra")
--- a/spectroscopy/code_src/spitzer_functions.py
+++ b/spectroscopy/code_src/spitzer_functions.py
@@ -47,7 +47,7 @@ def SpitzerIRS_get_spec(sample_table, search_radius_arcsec, COMBINESPEC):
 
         # If multiple entries are found, pick the closest.
         # Or should we take the average instead??
-        if len(tab) > 0:
+        if len(tab) > 1:
             print("More than 1 entry found", end="")
             if not COMBINESPEC:
                 print(" - pick the closest")
@@ -63,7 +63,7 @@ def SpitzerIRS_get_spec(sample_table, search_radius_arcsec, COMBINESPEC):
 
         # Now extract spectra and put all in one array
         specs = []
-        for tt in tab:
+        for tt in tab_final:
             url = "https://irsa.ipac.caltech.edu{}".format(tt["xtable"].split("\"")[1])
             spec = Table.read(url, format="ipac")  # flux_density in Jy
             specs.append(spec)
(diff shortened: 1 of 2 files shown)


spectroscopy/code_src/spitzer_functions.py
```diff
@@ -47,7 +47,7 @@ def SpitzerIRS_get_spec(sample_table, search_radius_arcsec, COMBINESPEC):
 
         # If multiple entries are found, pick the closest.
         # Or should we take the average instead??
-        if len(tab) > 0:
+        if len(tab) > 1:
             print("More than 1 entry found", end="")
             if not COMBINESPEC:
                 print(" - pick the closest")
```
troyraen, line 53: This code change:
- appears to be what the original author intended. (```print("More than 1 entry found"```)
- does *not* change the results but does make the code more efficient. (The code block that finds the closest entry now gets skipped when `len(tab) == 1`.)

spectroscopy/code_src/spitzer_functions.py
```diff
@@ -63,7 +63,7 @@ def SpitzerIRS_get_spec(sample_table, search_radius_arcsec, COMBINESPEC):
 
         # Now extract spectra and put all in one array
         specs = []
-        for tt in tab:
+        for tt in tab_final:
```
troyraen, line 66: This code change *does* change the results, and appears to be what the author intended. The code block above this produces `tab_final` which contains the single row from `tab` that is closest to the target if `COMBINESPEC` is True, else a full copy of `tab`. In the original code, `tab_final` was not ever used, so `COMBINESPEC` had no effect -- spectra were always combined.

spectroscopy/code_src/spitzer_functions.py
jkrick: It makes sense to me that this is a typo and not intentional.  I agree with the change.

Along the same lines, do we need the 'else' statement for this 'if' statment?  since we have already covered the case where it is 0, it should never get to that else.  Which may mean that we can get rid of the whole 'if len(tab)> 1 ...else' and reduce the indentation???  Somehow I can't leave this comment on the appropriate line, sorry.

spectroscopy/code_src/spitzer_functions.py
jkrick: yes, I agree that the original intent of the code is to use COMBINESPEC to determine if the spectra should be combined or chosen based on the nearest, and I agree that is not happening (good catch!).  I agree this change you suggest makes sense.

jkrick (approved):
Thanks @troyraen for catching those logic issues.  I think your suggested changes are sensible and I left comments on them in response to your comments. 

I don't know why I can't leave comments on other places in the code, maybe because I wasn't tagged to review?

Some additional suggestions for the code in addition to the inline comments:
- line 14: change 'HST' to 'Spitzer' - which makes me think that the HST function should be checked for the same logic problems being fixed in this PR (Hey!  I read the doctstring)
- could simplify the np.nanmin line to be np.nanargmin which returns the index directly - not important at all, but I had to look that one up and found the simpler solution

@afaisst if you have time this week, could you check this logic stuff, but if not, I am pretty confident that the changes are what the code was intended to do.

troyraen:
> I don't know why I can't leave comments on other places in the code

You may be running into github's limit on the number of lines above or below an actual change that's being made in the PR.

spectroscopy/code_src/spitzer_functions.py
troyraen: > Along the same lines, do we need the 'else' statement for this 'if' statement?

That code block wasn't getting hit originally but it will now when `len(tab) == 1`. And/but I agree that this could be made simpler (== easier to follow).

troyraen:
I applied your suggestions @jkrick. Good call on `np.nanargmin`. The HST logic is different but there were similar problems in the docstrings (probably from me copy/pasting in #356).

I think this is ready but I'll wait a bit to see if @afaisst wants to review.

troyraen:
Rebased on top of main to pull in recent changes.

troyraen:
Changes approved by @afaisst. Merging now.

PR #365, opened 2024-12-13
Minor bug fixes for spectra_generator

troyraen:
- Commit data/ and output/ directories, empty except for READMEs. (These are already ignored in .gitignore to prevent actual data files from being committed.)
- In plotting function, make sure output/ dir exists before writing to it.
- Add a target so DESIBOSS_get_spec returns data


bsipocz (approved):
Thanks!

PR #368, opened 2025-02-11
MAINT: updating googledrivedownloader usage 

bsipocz:
following its refactoring upstream that included the change in the importable package name and the removal of the class structure


bsipocz:
@sawdog - this should fix the error you run into, they completely refactored the package we use for accessing google drives so now we need to add a version limit to it.

PR #373, opened 2025-02-21
change order, add jupytext command

trjaffe:
Personally I think the User Docs for Fornax should come first.  Otherwise people launch into the notebooks without knowing anything about how to use the platform.

Also added the jupytext command.

### DIFF
--- a/index.md
+++ b/index.md
@@ -28,3 +28,6 @@ forced_photometry/README
 light_curves/README
 ```
 
+Note that we store these notebooks as markdown files.  In the Fornax Science Console, these should open automatically in the runnable jupyter notebook format.  If they do not, or you run them elsewhere, you may need to convert them explicitly with 
+
+`jupytext --to notebook yourfile.md`
--- a/index.md
+++ b/index.md
@@ -4,6 +4,18 @@ NASA Astrophysics is developing the Fornax Initiative, a cloud-based system that
 brings together data, software, and computing so that researchers can focus on science.
 
 
+## User Documentation
+
+```{toctree}
+---
+maxdepth: 2
+caption: User Documentation
+---
+
+documentation/README
+
+```
+
 ## Tutorial Notebooks
 
 ```{toctree}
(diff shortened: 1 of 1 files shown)


bsipocz:
The documentation is more about the console and Fornax in general and not the notebooks though, and if they use this from the console the docs is deployed differently, I believe. 

So I would tag @vandesai1 and @jkrick here to chime in.

index.md
```diff
@@ -4,27 +4,30 @@ NASA Astrophysics is developing the Fornax Initiative, a cloud-based system that
...
-caption: User Documentation
+caption: Tutorial Notebooks
 ---
 
-documentation/README
+forced_photometry/README
+light_curves/README
+```
+
+Note that we store these notebooks as markdown files.  In the Fornax Science Console, these should open automatically in the runnable jupyter notebook format.  If they do not, or you run them elsewhere, you may need to convert them explicitly with
```
bsipocz, line 31: > "runnable jupyer notebook format" 

the myst-markdown is a totally legit jupter notebook format, so this will need to be rephrased. And they don't open up, the solution should not be the conversion but to sort out dependencies, or right click and open them up that way, explicitly as notebook.

(We can have the command half hidden somewhere, but I would rather advocate for how to use these properly and stop sending ipynb formats around).

jkrick:
Since this is the GitHub page, I am assuming that people are coming to this website looking for code.  So, while I think we should link to the documentation here, I don't think it should be first. I guess it depends on how you `first` encounter Fornax.

When someone logs onto daskhub, they will see documentation there as well.

index.md
```diff
@@ -4,27 +4,30 @@ NASA Astrophysics is developing the Fornax Initiative, a cloud-based system that
 brings together data, software, and computing so that researchers can focus on science.
 
 
-## Tutorial Notebooks
+## User Documentation
```
bsipocz, line 7: I don't think these changes are correct, with this PR the rendered HTMLs will have the fornax documentation first and then the notebooks.

bsipocz:
We have added more docs about myst markdown as well as the jupyterlab extension is now included in the fornax image, so there is really no more need for ipynb workarounds. 

And as the review pointed out the swap over with the FSC documentation is not really in scope for this tutorial repo, thus I go ahead and close this without merge. 

Thank you for initiating the discussion, it helped us with the restructure and extension of the docs!

Issue #376, opened 2025-02-25
BUG: pan-starrs module in light curve generator giving TypeError


jkrick:
section 2.4 of light curve generator seems to have an unexpected keyword:

```
[/opt/conda/envs/notebook/lib/python3.11/site-packages/hipscat/io/file_io/file_io.py:292](https://daskhub-dev.fornax.smce.nasa.gov/opt/conda/envs/notebook/lib/python3.11/site-packages/hipscat/io/file_io/file_io.py#line=291): UserWarning: point_map.fits file written in RING ordering, due to https://github.com/astronomy-commons/hipscat/issues/271. Converting to NESTED.
  warnings.warn(
/opt/conda/envs/notebook/lib/python3.11/site-packages/hipscat/io/file_io/file_io.py:292: UserWarning: point_map.fits file written in RING ordering, due to https://github.com/astronomy-commons/hipscat/issues/271. Converting to NESTED.
  warnings.warn(
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[10], line 5
      3 panstarrs_search_radius = 1.0 # search radius = 1 arcsec
      4 # get panstarrs light curves
----> 5 df_lc_panstarrs = panstarrs_get_lightcurves(sample_table, radius=panstarrs_search_radius)
      7 # add the resulting dataframe to all other archives
      8 df_lc.append(df_lc_panstarrs)

File [~/fornax-demo-notebooks/light_curves/code_src/panstarrs_functions.py:59](https://daskhub-dev.fornax.smce.nasa.gov/user/jkrick/lab/tree/notebooks/fornax-demo-notebooks/light_curves/code_src/panstarrs_functions.py#line=58), in panstarrs_get_lightcurves(sample_table, radius)
     53 sample_df = pd.DataFrame({'objectid': sample_table['objectid'],
     54                       'ra_deg': sample_table['coord'].ra.deg,
     55                       'dec_deg':sample_table['coord'].dec.deg,
     56                       'label':sample_table['label']})
     58 #convert dataframe to hipscat
---> 59 sample_lsdb = lsdb.from_dataframe(
     60     sample_df, 
     61     ra_column="ra_deg", 
     62     dec_column="dec_deg", 
     63     margin_threshold=10,
     64     # Optimize partition size
     65     drop_empty_siblings=True
     66 )
     68 #plan to cross match panstarrs object with my sample 
     69 #only keep the best match
[...]

PR #379, opened 2025-02-26
removing drop_empty_siblings parameter from lsdb.from_dataframe

jkrick:
simple fix to issue #376 .  The lsdb documentation says that this parameter exists, so I have opened an issue over there to ask them to clarify their documentation, and commented out this line for now in our code. 

This is somewhat pointless, because making this fix then allows me to run the code and realize that the panstarrs module doesn't work at all anymore because hipscat is apparently not a thing anymore.  I will open a new issue for that.

### DIFF
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -62,7 +62,8 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
         dec_column="dec_deg", 
         margin_threshold=10,
         # Optimize partition size
-        drop_empty_siblings=True
+        #drop_empty_siblings=True
+        #this parameter appears to be deprecated 2/26/25
     )
 
     #plan to cross match panstarrs object with my sample 


light_curves/code_src/panstarrs_functions.py
```diff
@@ -62,7 +62,8 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
         dec_column="dec_deg", 
         margin_threshold=10,
         # Optimize partition size
-        drop_empty_siblings=True
+        #drop_empty_siblings=True
+        #this parameter appears to be deprecated 2/26/25
```
bsipocz, line 66: Delete unnecessary lines rather than comment them out. If needed we can always go back in the git history and check on previous versions if needed.

bsipocz:
> anymore because hipscat is apparently not a thing anymore. I will open a new issue for that.

We should have some issues for hipscat--> hats already, but by all means open a panstars specific bug report, too as I don't think the rename was meant to be done in a way to mess up existing usage.

jkrick:
closing this because it is not necessary with PR #382  which gets rid of read_hipscat anyway in favor of read_hats

PR #385, opened 2025-03-06
Point photometry notebook at new s3 bucket

troyraen:
The new s3 bucket 'irsa-fornax-testdata' has replaced the old bucket 'irsa-mast-tike-spitzer-data'. There is currently only one call to this bucket in all of the notebooks. This PR updates that call to point to the new bucket.

### DIFF
--- a/forced_photometry/multiband_photometry.md
+++ b/forced_photometry/multiband_photometry.md
@@ -199,11 +199,11 @@ spitzer = cosmos_results[cosmos_results['dataset'] == 'IRAC']
 
 ```{code-cell} ipython3
 # Temporarily add the cloud_access metadata to the Atlas response.
-# This dataset has limited acces, thus 'region' should be used instead of 'open'.
-# S3 access should be available from the daskhub and those who has their IRSA token set up.
+# This dataset has limited access, thus 'region' should be used instead of 'open'.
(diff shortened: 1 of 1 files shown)


bsipocz (approved):
Thanks!

troyraen:
When testing this, I found that this bucket is not actually being accessed. It seems that the SIA call reverts to downloading from "On-prem". Here is the relevant cell and the output:

```
>>> fornax_download(spitzer, access_url_column='sia_url', fname_filter='go2_sci',
                data_subdirectory='IRAC', verbose=False)
2025-03-06 00:57:05,315 | fornax | data region: us-east-1
2025-03-06 00:57:05,316 | fornax | data access mode: restricted
2025-03-06 00:57:05,529 | fornax | --- Downloading data from On-prem ---
...
```

I got the same result when I replaced the bucket name with an empty string, which I think verifies that it is not actually touching the bucket.

@bsipocz do you know any details about this code or its history? I don't even know who wrote it.

bsipocz:
I can dig up the history, but so much has changed since that I'm not surprised it actually falling back to the on prem version. 

I would say, merge the pr as is and there are other issues open for changing this dataset of the standard mosaics instead, so neither the bucket, nor the workarounds are needed.

PR #388, opened 2025-03-12
added AI generated docstrings

### LINKED ISSUE #263: Document private functions in helpers/scale_up.py
Caused by #195 

light_curves/code_src/helpers/scale_up.py

jkrick:
this will close #263

### DIFF
--- a/light_curves/code_src/helpers/scale_up.py
+++ b/light_curves/code_src/helpers/scale_up.py
@@ -34,19 +34,18 @@ DEFAULTS = {
 
 
 def run(*, build, **kwargs_dict):
-    """Run the light_curve_generator step indicated by `build`.
+    """
+    Run the light_curve_generator step indicated by `build`.
 
     Parameters
-    ==========
-
+    ----------
     build : str
         Which step to run. Generally either "sample" or "lightcurves". Can also be either "kwargs"
         or any of the `kwargs_dict` keys, and then the full set of kwargs is built from `kwargs_dict`
         and either the whole dictionary is returned (if "kwargs") or the value of this key is returned.
-
-    kwargs_dict
+    kwargs_dict : dict
         Key/value pairs for the build function. This can include any key in the dict
-        `helpers.scale_up.DEFAULTS` plus "archive". These are described below.
+        `helpers.scale_up.DEFAULTS` plus "archive".
 
         run_id : str
             ID for this run. This is used to name the output subdirectory ("base_dir") where the
@@ -55,7 +54,7 @@ def run(*, build, **kwargs_dict):
         get_samples : list or dict[dict]
             Names of get_<name>_sample functions from which to gather the object sample.
             To send keyword arguments for any of the named functions, use a dict with key=name
-             value=dict of keyword arguments for the named function). Defaults will be
+            value=dict of keyword arguments for the named function). Defaults will be
             used for any parameter not provided.
 
         consolidate_nearby_objects : bool
@@ -91,6 +90,11 @@ def run(*, build, **kwargs_dict):
         parquet_dataset_name : str
             Name of the directory to read/write the parquet dataset, relative to the base_dir.
             The dataset will contain one .parquet file for each archive that returned light curve data.
+
+    Returns
+    -------
+    Various
+        Depending on the `build` parameter, returns the result of the corresponding build function.
     """
     my_kwargs_dict = _construct_kwargs_dict(**kwargs_dict)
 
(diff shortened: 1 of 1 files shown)


bsipocz (commented):
Unfortunately this doesn't follow the docstring format that we tried to use normally (numpydoc), but will likely be an OK good first draft.

(CI doesn't (yet) checks for formatting, so it's not breaking anything, but still, I would think it would be good to stay consistent)

bsipocz:
Overall, I'm not sure where AI got its formatting, this is not a style I recall seeing anywhere.

jkrick:
Happy to change it @bsipocz if you tell me more specifically what parts are not conforming.  When I look at the docs style guide https://numpydoc.readthedocs.io/en/latest/format.html I see a section for a short summary, extended summary, parameters, returns, and notes (among other things).  Or is it more the spacing and placement of semicolons that are a problem?

bsipocz:
> Or is it more the spacing and placement of semicolons that are a problem?

yeap. The sections look OK, but the formatting around them is not exactly correct. I'll check on the actual content later, but with a brief first look, it looked good.

troyraen:
@jkrick you might try telling the AI to "use numpydoc style".

jkrick:
@bsipocz  let me know if this is better, and if it is, please merge.

light_curves/code_src/helpers/scale_up.py
```diff
@@ -107,7 +111,36 @@ def run(*, build, **kwargs_dict):
 
 
 def _build_sample(*, get_samples, consolidate_nearby_objects, sample_file, overwrite_existing_sample, **_):
```
bsipocz, line 113: Wow, what is this `**_`? We really should not swallow unused kwargs, especially when everything is nicely spelt out. 

(similarly, all the above should not have `**kwargs` when those kwargs are 1) few 2) well defined, e.g. the function is not a placeholder of other functions with different kwargs.

light_curves/code_src/helpers/scale_up.py
```diff
@@ -107,7 +111,36 @@ def run(*, build, **kwargs_dict):
...
+        If True, the existing sample file will be overwritten if it exists.
+
+    Returns
+    -------
+    astropy.table.Table
+        The resulting sample table containing the AGN objects.
+
+    Notes
+    -----
+    If the sample file already exists and `overwrite_existing_sample` is False, the function will read
```
bsipocz, line 139: We don't generate API docs from these, but overall sphinx is unhappy with unsolvable links like this, so double backticks is usually the way to go.
```suggestion
    If the sample file already exists and ``overwrite_existing_sample`` is False, the function will read 
```

light_curves/code_src/helpers/scale_up.py
```diff
@@ -201,8 +293,30 @@ def _build_other(keyword, **kwargs_dict):
...
+
+    Returns
+    -------
+    dict
+        A dictionary containing the combined and final keyword arguments for the run, sorted by key.
+
+    Notes
+    -----
+    Default values are defined in the `DEFAULTS` dictionary.
+    If `use_yaml` is True in `kwargs_dict` or in the defaults, the function will load additional keyword
```
bsipocz, line 314: Same, all of these should be double backticks. But 1) we don't build docs from these, 2) these are private functions so even if we made API docs, these would not be included. 3) this really is a numpydoc bug/known issue.

light_curves/code_src/helpers/scale_up.py
```diff
@@ -283,79 +459,22 @@ def _now():
...
-    yaml_path = run(build="yaml_file", **kwargs_dict)
-    with open(yaml_path, "w") as fout:
-        yaml.safe_dump(kwargs_dict, fout)
-    print(f"kwargs written to {yaml_path}")
-
-
-# ---- helpers for __name__ == "__main__" ---- #
-
-
-def _argparser():
```
bsipocz, line 302: Did you mean to remove these?

light_curves/code_src/helpers/scale_up.py
troyraen: Oh, I just realized this PR is against a file that I wrote. Thanks @jkrick for adding these docstrings! I made an effort to document the bash script and public python functions clearly but I completely skipped the private ones.

@bsipocz yes I agree, the `**_` and `**kwargs` are not good and I generally avoid them for the reasons you say and more 🙃. In this case, I recall that it was much more convenient to let the bash script (which is the main feature of this "notebook" and is the thing that calls this scale_up.py script) just pass in all the parameters and let this python script sort them out than it was to make the bash script sort them out. But that led to bad stuff here like needing to silently discard args that this function doesn't need using `**_`. The whole thing could certainly benefit from a redesign but it's not a simple fix because it involves how configs get passed around between files/processes. By the time I got to this part of the code I opted to leave it and just finish the pr. These scripts were useful --  this is how I was able to actually run the full-scale 500k sample size and get all the light curves -- but then we stashed the results for future use and I don't think anyone has needed or wanted to use this again. Going forward, I'm still not sure how we are "supposed" to launch large-scale jobs on Fornax and I'd like to hear some updated recommendations from the infrastructure folks before we try to really maintain this.

light_curves/code_src/helpers/scale_up.py
troyraen: Looks accidental. Deleting these will break it.

jkrick: WOW!  so sorry, that is scary accidental.  I'm glad you caught that.

jkrick:
ok, I added back in the accidentally deleted functions and their doc strings.  I did not worry about the double back tics as you say that is a known issue and has no measurable effects.  Should be good to merge when you are ready @bsipocz

light_curves/code_src/helpers/scale_up.py
bsipocz: no worries, this is why we do PR reviews :)

bsipocz (approved):
Thanks! Let's merge this now.

Issue #258, opened 2024-03-28
change ML_utils filename


jkrick:
This filename is confusing to me because really it houses functions for the ML_AGNzoo notebook.  Can we change the name to something more like "AGNzoo_ML_utils" or "AGNzoo_utils"?

And I will make a "classifier_utils" file

PR #389, opened 2025-03-12
changing name of ML_utils to be in line with other function files

jkrick:
will close issue #258

changed filename from 'utils' to 'functions' and made it more specific to the AGNzoo notebook.


bsipocz (approved):
Thank you!

PR #396, opened 2025-03-28
Run autopep8 and strip whitespace in light_curves/code_src/

troyraen:
No substantive changes, just formatting.


bsipocz (approved):
Thanks!

bsipocz:
Not sure why we ended up with a DeadKernel with the classifier, I'm restarting the build (it should have just having the scale-up issues, but the other notebooks should still work)

bsipocz:
FYI: the restart build didn't run into the same deadkernel, we're very much on the limit that sometimes fails and sometimes passes. I've opened up #398 to track it.

PR #382, opened 2025-02-27
converting panstarrs read_hipscat to read_hats

jkrick:
panstarrs module in light_curve_generator was broken maybe because read_hipscat is gone, or the hipscat files themselves were gone.  So, the switch to hats format is in this PR.

I mostly just changed lsdb.read_hipscat to lsdb.read_hats.  Some warnings pop up now in this panstarrs module, so I have added a line of text to warn people that the warnings are expected.

This closes #380 

This also closes #376

PR #397, opened 2025-03-28
Patch scale_up notebook for PanSTARRS

troyraen:
This should merge after #396.

- Bugfix multiprocessing. Followup to #387 and #382.
- Include panstarrs with calls that scale well. Followup to #382.

### DIFF
--- a/light_curves/code_src/helpers/scale_up.py
+++ b/light_curves/code_src/helpers/scale_up.py
@@ -18,7 +18,7 @@ sys.path.append(str(HELPERS_DIR.parent))  # put code_src dir on the path
 ARCHIVE_NAMES = {
     "all": ["Gaia", "HCV", "HEASARC", "IceCube", "PanSTARRS", "TESS_Kepler", "WISE", "ZTF"],
     # these are expected to run successfully at scale
-    "scaled": ["Gaia", "HEASARC", "IceCube", "WISE", "ZTF"],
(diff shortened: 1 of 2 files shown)


troyraen:
The notebook now runs to completion for me on Fornax but circleci is failing here. This runs a similar call to the light_curve_generator's parallel processing cell and requires just over 4G ram. I don't see anywhere in the configs that we're specifying how big of a machine these jobs are running on, so I'm wondering if we're getting some default machine size that is smaller than this requires?

bsipocz:
The circleci issue is new as of yesterday, the classifier notebook is the one that started running into the resources.

bsipocz:
Here are the resource usage stats: https://app.circleci.com/pipelines/github/nasa-fornax/fornax-demo-notebooks/374/workflows/6c8a8ab8-85fc-4710-80c0-4a31abbfcf54/jobs/310/resources

bsipocz (approved):
Changes here look good, the issue is with hitting resources on CircleCI.

GHA has better limits, so I go ahead and merge this and will pull over some configs we have at IRSA to enable GHA testing on PRs, etc.

bsipocz:
FYI, I don't think that should have worked, e.g. the other PR goes into main then this PR still has a head branch at the top of the other branch and not on main.

Will fix all of this in a terminal.

PR #405, opened 2025-04-24
possible workaround for astroquery.mast JWST timeout

jkrick:
following suggestion on this issue: https://github.com/astropy/astroquery/issues/2473

Neither the original JWST query, nor this updated code timeout for me on Fornax so i can't tell if I have actually solved the timeout problem.  The PR is faster (10min compared to 17min) so probably worth using regardless.


bsipocz (approved):
Thanks!

PR #409, opened 2025-05-05
hats light curves

### LINKED ISSUE #371: Bug: Light Curves notebook lost access to ZTF data
In the light_curve_generator.md notebook, the `ztf_get_lightcurves` function no longer works because it relies on DR18 which is no longer available through IRSA. To clarify: We put the actual ZTF DR18 Lightcurves files in an S3 bucket specifically for this purpose -- the call to IRSA's on-premise service was a necessary prerequisite to fetching the cloud data because of the way ZTF structures its Lightcurves dataset. IRSA only serves the 5 most recent ZTF DRs from on premise, so we lost access to DR18 when DR23 was released in Jan 2025.

Option 1:

Replace the DR18 files that are in the bucket with the same files for DR23. The notebook code should then only need a single line changed to point to the new DR and work again.

Option 2:

A HATS version of ZTF DR23 is currently in prep. Using this version would eliminate the need to call the on-premise service, which would be a good improvement for this notebook. It would require a refactoring of `ztf_get_lightcurves`, but this was always the intention for this function (#367).

jkrick:
This PR does 3 main things
1) makes a hats_get_lightcurves function which generalizes grabbing light curves from hats files.  This is with the hope that more files in the future will also be in this format.  For now, ZTF and Panstarrs are in similar enough formats to use this general function.

2) re-writes the panstarrs_get_lightcurves functions to be a wrapper around hats_get_lightcurves.  We were already using a hats catalog for panstarrs, but the change here is using the generic function written for 1) above.

3) re-writes ztf_get_lightcurves function to be a wrapper around hats_get_lightcurves.  This is a complete re-write of that function and is nothing like the original function.  Unfortunately I can not benchmark it against the old function because that is not working with the old data release due to issue #371.

This PR will close #371

### DIFF
--- a/light_curves/code_src/hats_functions.py
+++ b/light_curves/code_src/hats_functions.py
@@ -1,0 +1,133 @@
+import numpy as np
+import pandas as pd
+from astropy.table import Table
+import lsdb
+from dask.distributed import Client
+from data_structures import MultiIndexDFObject
+from upath import UPath
+
+def hats_get_lightcurves(
+    sample_table: Table,
+    object_catalog_path: str,
+    lightcurve_catalog_path: str,
+    object_columns: list[str],
+    light_curve_columns: list[str],
+    filter_id_to_name: dict[int, str],
+    id_col: str,
+    time_col: str,
+    flux_col: str,
+    err_col: str,
+    *,
+    radius: float = 1.0,
+    object_margin_cache=None,
+    lightcurve_margin_cache=None,
+) -> MultiIndexDFObject:
+    """
+    Generic LSDB/HATs light-curve fetcher by cross-matching an object catalog
+    to an input coord list, then joining to a light-curve catalog.
+
+    Parameters
+    ----------
+    sample_table : astropy.table.Table
+        Table with columns ['objectid', 'coord', 'label'] (SkyCoord in 'coord').
+    object_catalog_path : str
+        S3/HATs path to the **object** catalog (used for cross-matching).
+    lightcurve_catalog_path : str
+        S3/HATs path to the **light curve** catalog.
+    object_columns : list of str
+        Columns to read from the object catalog (must include `id_col`, ra/dec).
+    detect_columns : list of str
+        Columns to read from the light-curve catalog (must include `id_col`,
+        `time_col`, `flux_col`, `err_col`, plus the filter-ID column).
+    filter_id_to_name : dict
+        Mapping from integer filter IDs to human-readable band names.
+    id_col : str
+        Name of the object-ID column in both catalogs.
+    time_col : str
+        Name of the time/MJD column in the light-curve catalog.
+    flux_col : str
+        Name of the flux column in the light-curve catalog.
+    err_col : str
+        Name of the flux-uncertainty column in the light-curve catalog.
+    radius : float, optional
+        Cross-match radius in arcseconds (default 1.0).
+
+    Returns
+    -------
+    MultiIndexDFObject
+        A multi-indexed (objectid, label, band, time) container of the light curves.
+    """
+    # 1) read the two catalogs lazily
+    
+    # this table will be used for cross matching with our sample's ra and decs
+    # but does not have light curve information
+    obj_cat = lsdb.read_hats(object_catalog_path,
+                             columns=object_columns,
+                             margin_cache=object_margin_cache)
+    # this table houses the light curves
+    lc_cat  = lsdb.read_hats(lightcurve_catalog_path,
+                             columns=light_curve_columns,
+                            margin_cache=lightcurve_margin_cache)
+    
+    # 2) convert astropy sample_table into an LSDB catalog
+    sample_df = pd.DataFrame({
+        'objectid': sample_table['objectid'],
+        'ra_deg':    sample_table['coord'].ra.deg,
+        'dec_deg':   sample_table['coord'].dec.deg,
+        'label':     sample_table['label']
+    })
+    # convert dataframe to hats catalog
+    sample_lsdb = lsdb.from_dataframe(
+        sample_df,
+        ra_column='ra_deg',
+        dec_column='dec_deg',
+        margin_threshold=10,
+        drop_empty_siblings=True
+    )
+
+    # 3) cross-match to find nearest object
+    # only keep the best match
+    matched_obj = obj_cat.crossmatch(
+        sample_lsdb,
+        radius_arcsec=radius,
+        n_neighbors=1,
+        suffixes=("", "")     # ← prevent “objID” → “objID_1”
+    )
+
+    # 4) join to the light-curve catalog
+    matched_lc = matched_obj.join(
+        lc_cat,
+        left_on=id_col,
+        right_on=id_col,
+        output_catalog_name=f"{id_col}_lc",
+        suffixes=["", ""]
+    )
+
+    # 5) compute on Dask
+    # here is where the actual work gets done
+    with Client():
+        matched_df = matched_lc.compute()
+
+    # 6) handle no matches
+    if matched_df.empty:
+        return MultiIndexDFObject(data=pd.DataFrame())
+
+    # 7) map filter IDs to names
+    filter_col = next(c for c in light_curve_columns if 'filter' in c.lower())
+    band_names = np.vectorize(filter_id_to_name.get)(matched_df[filter_col])
+
+    # 8) build and index the DataFrame
+    df_lc = pd.DataFrame({
+        'flux':     pd.to_numeric(matched_df[flux_col], errors='coerce'),
+        'err':      pd.to_numeric(matched_df[err_col],  errors='coerce'),
+        'time':     pd.to_numeric(matched_df[time_col], errors='coerce'),
+        'objectid': matched_df['objectid'].astype(np.int64),
+        'band':     band_names,
+        'label':    matched_df['label'].astype(str)
+    })
+    
+    df = pd.DataFrame(df_lc).set_index(['objectid', 'label', 'band', 'time'])
+    
+    return MultiIndexDFObject(data=df)
+
+
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -5,117 +5,63 @@ import lsdb
 from dask.distributed import Client
 from data_structures import MultiIndexDFObject
 from upath import UPath
+from hats_functions import hats_get_lightcurves
 
-# panstarrs light curves from hipscat catalog in S3 using lsdb
 
-
-def panstarrs_get_lightcurves(sample_table, *, radius=1):
-    """Searches panstarrs hipscat files for light curves from a list of input coordinates.
-
+def panstarrs_hats_get_lightcurves(sample_table: Table, *, radius: float = 1.0):
+    """
+    Searches panstarrs hats files for light curves from a list of input coordinates.  
+    
     Parameters
     ----------
-    sample_table : `~astropy.table.Table`
-        Table with the coordinates and journal reference labels of the sources
-    radius : float
-        search radius, how far from the source should the archives return results
+    sample_table : astropy.table.Table
+        Input table of targets. Must have columns:
+        
+        - ``objectid`` : int or str  
+          Unique identifier for each source.  
+        - ``coord`` : astropy.coordinates.SkyCoord  
+          SkyCoord of each source for cross-matching.  
+        - ``label`` : str  
+          Label to propagate into the output light-curve index.  
+    radius : float, optional
+        Cross-match search radius in arcseconds (default is 1.0).
 
     Returns
     -------
-    df_lc : MultiIndexDFObject
-        the main data structure to store all light curves
+
+    data_structures.MultiIndexDFObject
+        A container whose `.data` is a pandas DataFrame with a MultiIndex
+        (``objectid``, ``label``, ``band``, ``time``) and columns:
+        
+        - ``flux`` : float  
+          PSF flux in mJy (``psfFlux * 1e3``).  
+        - ``err`` : float  
+          PSF flux uncertainty in mJy (``psfFluxErr * 1e3``).
+
+
     """
-
-    # read in the panstarrs object table to lsdb
-    # this table will be used for cross matching with our sample's ra and decs
-    # but does not have light curve information
-    panstarrs_object = lsdb.read_hats(
-        UPath('s3://stpubdata/panstarrs/ps1/public/hats/otmo', anon=True),
-        margin_cache=UPath('s3://stpubdata/panstarrs/ps1/public/hats/otmo_10arcs', anon=True),
-        columns=["objID",  # PS1 ID
-                 "raMean", "decMean",  # coordinates to use for cross-matching
-                 "nStackDetections",  # some other data to use
-                 ]
+    df_lc =  hats_get_lightcurves(
+        sample_table,
+        object_catalog_path='s3://stpubdata/panstarrs/ps1/public/hats/otmo',
+        object_margin_cache='s3://stpubdata/panstarrs/ps1/public/hats/otmo_10arcs',
+        lightcurve_catalog_path='s3://stpubdata/panstarrs/ps1/public/hats/detection',
+        lightcurve_margin_cache='s3://stpubdata/panstarrs/ps1/public/hats/detection_10arcs',
+        object_columns=["objID", "raMean", "decMean", "nStackDetections"],
+        light_curve_columns=["objID", "detectID", "obsTime", "filterID", "psfFlux", "psfFluxErr"],
+        id_col='objID',
+        time_col='obsTime',
+        flux_col='psfFlux',
+        err_col='psfFluxErr',
+        radius=radius,
+        filter_id_to_name={
+            1: 'Pan-STARRS g',
+            2: 'Pan-STARRS r',
+            3: 'Pan-STARRS i',
+            4: 'Pan-STARRS z',
+            5: 'Pan-STARRS y',
+        }
     )
-    # read in the panstarrs light curves to lsdb
-    # panstarrs recommendation is not to index into this table with ra and dec
-    # but to use object ids from the above object table
-    panstarrs_detect = lsdb.read_hats(
-        UPath('s3://stpubdata/panstarrs/ps1/public/hats/detection', anon=True),
-        margin_cache=UPath('s3://stpubdata/panstarrs/ps1/public/hats/detection_10arcs', anon=True),
-        columns=["objID",  # PS1 object ID
-                 "detectID",  # PS1 detection ID
-                 # light-curve stuff
-                 "obsTime", "filterID", "psfFlux", "psfFluxErr",
-                 ]
-    )
-    # convert astropy table to pandas dataframe
-    # special care for the SkyCoords in the table
-    sample_df = pd.DataFrame({'objectid': sample_table['objectid'],
-                              'ra_deg': sample_table['coord'].ra.deg,
-                              'dec_deg': sample_table['coord'].dec.deg,
-                              'label': sample_table['label']})
+    return df_lc
 
-    # convert dataframe to hipscat
-    sample_lsdb = lsdb.from_dataframe(
-        sample_df,
-        ra_column="ra_deg",
-        dec_column="dec_deg",
-        margin_threshold=10,
-        # Optimize partition size
-        drop_empty_siblings=True
-    )
-
-    # plan to cross match panstarrs object with my sample
-    # only keep the best match
-    matched_objects = panstarrs_object.crossmatch(
-        sample_lsdb,
-        radius_arcsec=radius,
-        n_neighbors=1,
-        suffixes=("", "")
-    )
-
-    # plan to join that cross match with detections to get light-curves
-    matched_lc = matched_objects.join(
-        panstarrs_detect,
-        left_on="objID",
-        right_on="objID",
-        output_catalog_name="yang_ps_lc",
-        suffixes=["", ""]
-    )
-
-    # Create default local cluster
-    # here is where the actual work gets done
-    with Client():
-        # compute the cross match with object table
-        # and the join with the detections table
-        matched_df = matched_lc.compute()
-
-    # handle the case where there are no matches and return empty df
-    if len(matched_df["filterID"]) == 0:
-        return MultiIndexDFObject(data=pd.DataFrame())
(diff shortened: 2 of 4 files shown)


bsipocz:
this needs a rebase to have a diff compared to the latest `main`

bsipocz:
I think there is minimal risk to damage, there are only 2 commits, so it should be an easy one. Shall we switch to slack?

But also, I'm happy to do it if you prefer.

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+    object_columns: list[str],
+    light_curve_columns: list[str],
+    filter_id_to_name: dict[int, str],
+    id_col: str,
+    time_col: str,
+    flux_col: str,
+    err_col: str,
+    *,
+    radius: float = 1.0,
+    object_margin_cache=None,
```
troyraen, line 22: Needs an entry in the docstring.

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+    lightcurve_margin_cache=None,
```
troyraen, line 23: I think this margin arg is unnecessary since this catalog is not being crossmatched, only joined on object ID.

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+    ----------
+    sample_table : astropy.table.Table
+        Table with columns ['objectid', 'coord', 'label'] (SkyCoord in 'coord').
+    object_catalog_path : str
+        S3/HATs path to the **object** catalog (used for cross-matching).
+    lightcurve_catalog_path : str
+        S3/HATs path to the **light curve** catalog.
+    object_columns : list of str
+        Columns to read from the object catalog (must include `id_col`, ra/dec).
+    detect_columns : list of str
```
troyraen, line 39:
```suggestion
    light_curve_columns : list of str
```

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+    object_catalog_path: str,
+    lightcurve_catalog_path: str,
+    object_columns: list[str],
+    light_curve_columns: list[str],
+    filter_id_to_name: dict[int, str],
+    id_col: str,
+    time_col: str,
+    flux_col: str,
+    err_col: str,
+    *,
```
troyraen, line 20:
```suggestion
    *,
    object_catalog_path: str,
    lightcurve_catalog_path: str,
    object_columns: list[str],
    light_curve_columns: list[str],
    filter_id_to_name: dict[int, str],
    id_col: str,
    time_col: str,
    flux_col: str,
    err_col: str,
```

I would make all of these keyword-only since it would be easy to mix them up.

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        'label':     sample_table['label']
+    })
+    # convert dataframe to hats catalog
+    sample_lsdb = lsdb.from_dataframe(
+        sample_df,
+        ra_column='ra_deg',
+        dec_column='dec_deg',
+        margin_threshold=10,
+        drop_empty_siblings=True
+    )
```
troyraen, line 86: It's not clear to me what this actually doing (what partitions it's making) since you're not supplying a kwarg for partitioning threshold. I'll check it out when I run the code.

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+
+    # 3) cross-match to find nearest object
+    # only keep the best match
+    matched_obj = obj_cat.crossmatch(
+        sample_lsdb,
+        radius_arcsec=radius,
+        n_neighbors=1,
+        suffixes=("", "")     # ← prevent “objID” → “objID_1”
```
troyraen, line 94: Pick unique suffixes for each catalog so that same-named columns don't get mixed up. I'm not sure what the problem is with objID but it probably needs to be handled a different way.

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+    )
+
+    # 4) join to the light-curve catalog
+    matched_lc = matched_obj.join(
+        lc_cat,
+        left_on=id_col,
+        right_on=id_col,
+        output_catalog_name=f"{id_col}_lc",
+        suffixes=["", ""]
```
troyraen, line 103: Same as above.

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        'err':      pd.to_numeric(matched_df[err_col],  errors='coerce'),
+        'time':     pd.to_numeric(matched_df[time_col], errors='coerce'),
+        'objectid': matched_df['objectid'].astype(np.int64),
+        'band':     band_names,
+        'label':    matched_df['label'].astype(str)
+    })
+    
+    df = pd.DataFrame(df_lc).set_index(['objectid', 'label', 'band', 'time'])
+    
+    return MultiIndexDFObject(data=df)
```
troyraen, line 131:
```suggestion
    df_lc = df_lc.set_index(['objectid', 'label', 'band', 'time'])
    
    return MultiIndexDFObject(data=df_lc)
```

Use the same name to avoid making an extra copy.

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+from astropy.table import Table
+import lsdb
+from dask.distributed import Client
+from data_structures import MultiIndexDFObject
+from upath import UPath
+
+def hats_get_lightcurves(
+    sample_table: Table,
+    object_catalog_path: str,
+    lightcurve_catalog_path: str,
```
troyraen, line 12:
```suggestion
    object_catalog_path: str | UPath,
    lightcurve_catalog_path: str | UPath,
```

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+    to an input coord list, then joining to a light-curve catalog.
+
+    Parameters
+    ----------
+    sample_table : astropy.table.Table
+        Table with columns ['objectid', 'coord', 'label'] (SkyCoord in 'coord').
+    object_catalog_path : str
+        S3/HATs path to the **object** catalog (used for cross-matching).
+    lightcurve_catalog_path : str
+        S3/HATs path to the **light curve** catalog.
```
troyraen, line 36:
```suggestion
    object_catalog_path : str or UPath
        S3 path to the **object** catalog in HATS format (used for cross-matching).
    lightcurve_catalog_path : str or UPath
        S3 path to the **light curve** catalog in HATS format.
```

light_curves/light_curve_generator.md
```diff
@@ -93,15 +93,15 @@ from gaia_functions import gaia_get_lightcurves
 from hcv_functions import hcv_get_lightcurves
 from heasarc_functions import heasarc_get_lightcurves
 from icecube_functions import icecube_get_lightcurves
-from panstarrs_functions import panstarrs_get_lightcurves
+from panstarrs_functions import panstarrs_hats_get_lightcurves
```
troyraen, line 96: Is there a reason for this change? None of the other `*_get_lightcurves` functions specify what format the data is in. Also, this change would require updating the scale_up notebook.

light_curves/code_src/ztf_functions.py
```diff
@@ -10,316 +11,91 @@
...
+        Input table of targets. Must have columns:
+        
+        - ``objectid`` : int or str  
+          Unique identifier for each source.  
+        - ``coord`` : astropy.coordinates.SkyCoord  
+          SkyCoord of each source for cross-matching.  
+        - ``label`` : str  
+          Label to propagate into the output light-curve index.  
+    radius : float, optional
+        Cross-match search radius in arcseconds (default is 1.0).
```
troyraen, line 37: There actually is no default set.

light_curves/code_src/ztf_functions.py
```diff
@@ -10,316 +11,91 @@
...
+        err_col='magerr',  # this becomes raw.data['err']
+        radius=radius,
+        filter_id_to_name=None,
+    )
+    # 2) turn the MultiIndex into columns
+    df = raw.data.reset_index()
+
+    # 3) rename columns
+    df = df.rename(columns={"hmjd": "time"})
+    #these are called flux and err in the generic hats_get_lightcurves, but are really magnitudes for ztf
```
troyraen, line 72: I would avoid having `hats_get_lightcurves` change the column names for this very reason.

light_curves/code_src/ztf_functions.py
```diff
@@ -10,316 +11,91 @@
...
+    df["flux"] = (mag * u.ABmag).to_value('mJy')
+    df["err"]  = (flux_up - flux_low) / 2
+
+    # 7) clean up intermediate columns (optional)
+    df = df.drop(columns=["mag", "magerr"])
+
+    # 8) re-index and return
+    df_lc = df.set_index(['objectid', 'label', 'band', 'time'])
+    
+    return MultiIndexDFObject(data=df_lc)
```
troyraen, line 100:
```suggestion
    df = df.set_index(['objectid', 'label', 'band', 'time'])
    
    return MultiIndexDFObject(data=df)
```

Use the same name to avoid making a copy.

light_curves/code_src/ztf_functions.py
```diff
@@ -1,5 +1,6 @@
 import multiprocessing as mp
```
troyraen, line 1: Remove unused libraries: multiprocessing, re, pandas, pyarrow, pyvo, tqdm, helpers.

light_curves/code_src/panstarrs_functions.py
```diff
@@ -5,117 +5,63 @@
 from dask.distributed import Client
```
troyraen, line 5: Remove unused libraries: numpy, pandas, lsdb, dask, upath.

light_curves/code_src/panstarrs_functions.py
```diff
@@ -5,117 +5,63 @@
...
-    df_lc = pd.DataFrame({
-        'flux': pd.to_numeric(matched_df['psfFlux'] * 1e3, errors='coerce').astype(np.float64),
-        'err': pd.to_numeric(matched_df['psfFluxErr'] * 1e3, errors='coerce').astype(np.float64),
-        'time': pd.to_numeric(matched_df['obsTime'], errors='coerce').astype(np.float64),
-        'objectid': matched_df['objectid'].astype(np.int64),
-        'band': filtername,
-        'label': matched_df['label'].astype(str)
-    }).set_index(["objectid", "label", "band", "time"])
 
     return MultiIndexDFObject(data=df_lc)
```
troyraen, line 67:
```suggestion
```

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+    # here is where the actual work gets done
+    with Client():
+        matched_df = matched_lc.compute()
+
+    # 6) handle no matches
+    if matched_df.empty:
+        return MultiIndexDFObject(data=pd.DataFrame())
+
+    # 7) map filter IDs to names
+    filter_col = next(c for c in light_curve_columns if 'filter' in c.lower())
```
troyraen, line 116: This is brittle since it implicitly relies on a specific naming syntax. I recommend having the user pass in the name of this column along with the others.

troyraen (commented):
Thanks for doing this! I left some comments based on reading through it. I haven't tried to run it yet.

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        left_on=id_col,
+        right_on=id_col,
+        output_catalog_name=f"{id_col}_lc",
+        suffixes=["", ""]
+    )
+
+    # 5) compute on Dask
+    # here is where the actual work gets done
+    with Client():
+        matched_df = matched_lc.compute()
```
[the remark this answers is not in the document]
troyraen: Another thing to experiment with in the future: I'm guessing that we'll get better results by tuning the `Client` kwargs `memory_limit`, `threads_per_worker`, and/or `n_workers`. [lsdb docs](https://docs.lsdb.io/en/stable/tutorials/dask-cluster-tips.html) suggest 1-2 threads per worker and 10G per worker per thread. 10G might not work so well on Fornax because of the different server sizes. `memory_limit=None` might work better here (I've had luck with that in the past, though it's mostly been on non-Fornax machines).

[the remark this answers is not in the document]
troyraen: And another for the future... I notice this in the panstarrs call:

```
INFO:distributed.core:Starting established connection to tcp://127.0.0.1:42698
/opt/conda/envs/notebook/lib/python3.12/site-packages/distributed/client.py:3370: UserWarning: Sending large graph of size 11.36 MiB.
This may cause some slowdown.
Consider loading the data with Dask directly
 or using futures or delayed objects to embed the data into the graph without repetition.
See also https://docs.dask.org/en/stable/best-practices.html#load-data-with-dask for more information.
```

My first guess is to call `.compute()` earlier to make the task graph simpler. But we should look into the link from the message.

troyraen (approved):
I'm having trouble getting it to run on Fornax. The ZTF call runs for about 4 minutes and then crashes, returning no results. The PANSTARRs call runs for about 70 seconds and then seems to crash (at least, it does not look like a normal dask shutdown) but it does return results. Weirdly though, it returns 2089 rows whereas this function prior to any hipscat/hats/lsdb (specifically, sha `df06e2ff148d56260c3eae13e557418a50750b56`) returns 2086 rows. But, the ZTF call doesn't currently work anyway and I don't think the PANSTARRs row issue was introduced in this PR. I left several suggestions below and above, but otherwise I think this is fine to merge if you want. We'll have to keep working on the lsdb + dask calls and figure out how to get them running more smoothly. Both of these functions ran a lot better (faster and used less memory) before switching to hats/lsdb.

Reminder to add a note and/or change the Runtime statement to reflect that this currently requires the server type: "Large - 64GB RAM/16 CPU".

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        radius_arcsec=radius,
+        n_neighbors=1,
+        suffixes=("", "")     # ← prevent “objID” → “objID_1”
+    )
+
+    # 4) join to the light-curve catalog
+    matched_lc = matched_obj.join(
+        lc_cat,
+        left_on=id_col,
+        right_on=id_col,
```
troyraen, line 101: The reason ZTF doesn't return data is that these columns have different names in Objects ("oid") vs Lightcurves ("objectid"). Currently the code is using `id_col="objectid"`, which ends up trying to join Lightcurves with our sample "objectid" rather than Objects "oid". To get this to work properly, I also had to change line 103 to assign unique suffixes to each table.

light_curves/code_src/ztf_functions.py
```diff
@@ -10,316 +11,91 @@
...
-    magerr = ztf_df["magerr"].to_numpy()
-    fluxupper = ((mag - magerr) * u.ABmag).to_value('mJy')
-    fluxlower = ((mag + magerr) * u.ABmag).to_value('mJy')
-    ztf_df["flux"] = (mag * u.ABmag).to_value('mJy')
-    ztf_df["err"] = (fluxupper - fluxlower) / 2
+    # 1) fetch the raw arrays 
+    raw = hats_get_lightcurves(
+        sample_table,
+        object_catalog_path=UPath('s3://irsa-fornax-testdata/ZTF/dr23/objects/hats'),
+        object_margin_cache=None,
```
troyraen, line 55: There is now a margin cache available for Objects, so you can do:

```suggestion
        object_margin_cache=UPath('s3://irsa-fornax-testdata/ZTF/dr23/objects/hats_margin_10arcsec'),
```

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+    with Client():
+        matched_df = matched_lc.compute()
+
+    # 6) handle no matches
+    if matched_df.empty:
+        return MultiIndexDFObject(data=pd.DataFrame())
+
+    # 7) map filter IDs to names
+    filter_col = next(c for c in light_curve_columns if 'filter' in c.lower())
+    band_names = np.vectorize(filter_id_to_name.get)(matched_df[filter_col])
```
troyraen, line 117: This crashes for ZTF because `filter_id_to_name=None`:

```
AttributeError: 'NoneType' object has no attribute 'get'
```

light_curves/code_src/hats_functions.py
```diff
@@ -0,0 +1,133 @@
...
+        dec_column='dec_deg',
+        margin_threshold=10,
+        drop_empty_siblings=True
+    )
+
+    # 3) cross-match to find nearest object
+    # only keep the best match
+    matched_obj = obj_cat.crossmatch(
+        sample_lsdb,
+        radius_arcsec=radius,
```
troyraen, line 92:
```suggestion
    matched_obj = sample_lsdb.crossmatch(
        obj_cat,
        radius_arcsec=radius,
```

I think these catalogs should be reversed. Otherwise this will return more than one row per target if there is more than one catalog object within `radius` of it. I think what we actually want is the single catalog object that is closest to each target.

jkrick: IThis PR isn't going to go through, but just to add that I tried swapping the order for the ztf functions and I got an error.  I do have "n_neighbors = 1" in the next line which is what tells it to only return the best match.  Full traceback if you want to go through it is below.  I will not bother with figuring this out for the new PR.

ValueError                                Traceback (most recent call last)
Cell In[4], line 5
      3 # get ZTF lightcurves
      4 ztf_search_radius = 1.0 #  arcsec
----> 5 df_lc_ZTF = ztf_get_lightcurves(sample_table, radius = ztf_search_radius)
      7 # add the resulting dataframe to all other archives
      8 df_lc.append(df_lc_ZTF)

File [~/fornax-demo-notebooks/light_curves/code_src/ztf_functions.py:86](https://science-console.fornax.smce.nasa.gov/jupyter/user/jkrick/lab/workspaces/fornax-demo-notebooks/light_curves/code_src/ztf_functions.py#line=85), in ztf_get_lightcurves(sample_table, radius)
     73 matched_lc = ztf_lc.crossmatch(
     74     ztf_lc,
     75     radius_arcsec=radius,
     76     n_neighbors=1,
     77 )
     78 #matched_lc = sample_lsdb.crossmatch(
     79 #    sample_lsdb,
     80 #    radius_arcsec=radius,
   (...)     84 # here is where the actual work gets done
     85 # compute the cross match with object table
---> 86 df = matched_lc.compute()
     88 #done with dask
     89 client.close() 

File [/opt/conda/envs/notebook/lib/python3.12/site-packages/lsdb/catalog/dataset/dataset.py:47](https://science-console.fornax.smce.nasa.gov/opt/conda/envs/notebook/lib/python3.12/site-packages/lsdb/catalog/dataset/dataset.py#line=46), in Dataset.compute(self)
     45 def compute(self) -> npd.NestedFrame:
     46     """Compute dask distributed dataframe to pandas dataframe"""
---> 47     return self._ddf.compute()

File [/opt/conda/envs/notebook/lib/python3.12/site-packages/lsdb/nested/core.py:437](https://science-console.fornax.smce.nasa.gov/opt/conda/envs/notebook/lib/python3.12/site-packages/lsdb/nested/core.py#line=436), in NestedFrame.compute(self, **kwargs)
    435 def compute(self, **kwargs):
    436     """Compute this Dask collection, returning the underlying dataframe or series."""
--> 437     return npd.NestedFrame(super().compute(**kwargs))

File [/opt/conda/envs/notebook/lib/python3.12/site-packages/dask/base.py:373](https://science-console.fornax.smce.nasa.gov/opt/conda/envs/notebook/lib/python3.12/site-packages/dask/base.py#line=372), in DaskMethodsMixin.compute(self, **kwargs)
    349 def compute(self, **kwargs):
    350     """Compute this dask collection
    351 
    352     This turns a lazy Dask collection into its in-memory equivalent.
   (...)    371     dask.compute
    372     """
--> 373     (result,) = compute(self, traverse=False, **kwargs)
    374     return result

File [/opt/conda/envs/notebook/lib/python3.12/site-packages/dask/base.py:681](https://science-console.fornax.smce.nasa.gov/opt/conda/envs/notebook/lib/python3.12/site-packages/dask/base.py#line=680), in compute(traverse, optimize_graph, scheduler, get, *args, **kwargs)
    678     expr = expr.optimize()
    679     keys = list(flatten(expr.__dask_keys__()))
--> 681     results = schedule(expr, keys, **kwargs)
    683 return repack(results)

File [/opt/conda/envs/notebook/lib/python3.12/site-packages/dask/dataframe/utils.py:357](https://science-console.fornax.smce.nasa.gov/opt/conda/envs/notebook/lib/python3.12/site-packages/dask/dataframe/utils.py#line=356), in check_meta()
    352     errmsg = "Expected partition of type `{}` but got `{}`".format(
    353         typename(type(meta)),
    354         typename(type(x)),
    355     )
    356 elif is_dataframe_like(meta):
--> 357     dtypes = pd.concat([x.dtypes, meta.dtypes], axis=1, sort=True)
    358     bad_dtypes = [
    359         (repr(col), a, b)
    360         for col, a, b in dtypes.fillna("-").itertuples()
    361         if not equal_dtypes(a, b)
    362     ]
    363     if bad_dtypes:

File [/opt/conda/envs/notebook/lib/python3.12/site-packages/pandas/core/indexes/base.py:4429](https://science-console.fornax.smce.nasa.gov/opt/conda/envs/notebook/lib/python3.12/site-packages/pandas/core/indexes/base.py#line=4428), in reindex()
   4426     raise ValueError("cannot handle a non-unique multi-index!")
   4427 elif not self.is_unique:
   4428     # GH#42568
-> 4429     raise ValueError("cannot reindex on an axis with duplicate labels")
   4430 else:
   4431     indexer, _ = self.get_indexer_non_unique(target)

ValueError: cannot reindex on an axis with duplicate labels

troyraen: I believe the error is happening because of the line below that passes `suffixes=("", "")` which is not advised because it cause this kind of confusion. (In this case, the library is confused. In other cases humans will be confused.)

troyraen: > I do have "n_neighbors = 1" in the next line which is what tells it to only return the best match.

But it will return the best match for every row in the catalog on the left, which means that rows in the catalog on the right might appear more than once in the results.

Issue #412, opened 2025-05-07
forced photometry notebook fails because the the downloaded irac data is not in the right place


zoghbi-a:
This is coming from the images testing CI.

The notebook does not run and stops with:
```python
FileNotFoundError: [Errno 2] No such file or directory: 'data/IRAC/irac_ch1_go2_sci_10.fits'
```

`handler.download` in `fornax_download` is expected to return file name but it never does in the aws case.

PR #413, opened 2025-05-07
Fix aws download in the forced photometry notebook; Fix-412

zoghbi-a:
Add return statements to the fornax download utility and make the notebook run.

### DIFF
--- a/forced_photometry/code_src/fornax.py
+++ b/forced_photometry/code_src/fornax.py
@@ -519,7 +519,7 @@ class AWSDataHandler(DataHandler):
 
             log.info('--- Downloading data from S3 ---')
             # proceed to actual download
-            self._download_file_s3(data_info, **kwargs)
+            return self._download_file_s3(data_info, **kwargs)
         else:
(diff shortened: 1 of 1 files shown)


forced_photometry/code_src/fornax.py
```diff
@@ -619,6 +619,7 @@ def progress_callback(numbytes):
                     pb.update(bytes_read)
 
             bkt.download_file(key, local_path, Callback=progress_callback)
+          return local_path
```
bsipocz, line 622: indent is off
```suggestion
            return local_path
```

bsipocz (changes requested):

zoghbi-a:
> Add return statements to the fornax download utility and make the notebook run. Fix #412

I used the web quick editor. It looks like I shouldn't be using it.

bsipocz (approved):
Thanks!

bsipocz:
> I used the web quick editor. It looks like I shouldn't be using it.

No worries, that's what review is for.

PR #415, opened 2025-05-07
Make plot labels with mathtext raw

### LINKED ISSUE #414: Use raw strings for mathtext when plotting
In the forced photometry notebook, there lines like:

```python
ax1.set(xlabel = 'COSMOS 2015 flux ($\mu$Jy)', ylabel = 'tractor flux ($\mu$Jy)', title = 'IRAC 3.6')
```
This prints warnings like:
```
SyntaxWarning: invalid escape sequence '\m'
```

They should be: `r'COSMOS 2015 flux ($\mu$Jy)'`. etc

I see this in the multi-band photometry notebook. Others appear to be ok, but may need to be checekd.

zoghbi-a:
Fix #414

### DIFF
--- a/forced_photometry/multiband_photometry.md
+++ b/forced_photometry/multiband_photometry.md
@@ -731,7 +731,7 @@ lims = [
 
 # now plot both limits against eachother
 ax1.plot(lims, lims, 'k-', alpha=0.75, zorder=0)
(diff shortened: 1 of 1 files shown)


bsipocz (commented):
Let me look into this, I can't recall using raw strings anywhere else in math mode, we should be able to opt-in in some other way.

bsipocz (approved):
It's a bit of a mix what we use, probably the whole label should be, if we use math mode, probably the whole label should be within the "$"s and not just the greek, but let's just go ahead with this fix now.

PR #417, opened 2025-05-09
Fix another mathtext occurance

zoghbi-a:
Fix another mathtext warning in forced photometry notebook similar to #415. The issue is reported in #414


bsipocz (approved):

bsipocz:
Thanks!

Issue #420, opened 2025-05-12
ENH: Experiment with how to optimize lsdb.from_dataframe in light curve notebooks


jkrick:
This is used to convert a sample that we made into lsdb in order to cross match it with a large catalog.  This happens in both code_src/ztf_functions and code_src/panstarrs_functions.

From Troy's code review:
This results in a hats catalog with about five partitions at order 0 and one partition at order 7 (for our default sample of 30). There are various choices we could make including min/max orders at which to partition and/or max number of rows in a partition. I assume these choices will affect the efficiency of the cross match, in addition to the sample size. It's probably worth doing some experimentation with this in the future.

jkrick:
the ztf call is now significantly slower than before conversion to hats and could use some attention to speed it up.

PR #421, opened 2025-05-12
Add ztf hats

### LINKED ISSUE #371: Bug: Light Curves notebook lost access to ZTF data

jkrick:
Making a generic hat_functions for both ztf and panstarrs is not possible at this point.  There are just too many individualized things that are different between the two different datasets (column names, data types).  Additionally, panstarrs recommends not accessing their light curves catalog directly but going through the object catalog.  For ZTF there is no such recommendation, so that search can happen with just a crossmatch.  All of this lead me to abandon that previous attempt and just make a ztf function to access the new hats files.

This PR will close #409.  I believe I have addressed all of the relevant comments from that PR in this one.  
This PR will close #371 

I did have to touch the wise function to make that work for testing the entire notebook.  There was a permissions issue which is fixed with this PR.

And this time I did check the entire flow of the notebook, and updated the runtime section to reflect that.  The entire runtime did increase from ~800s to ~1100s.  This is an unwelcome time increase when switching to hats file format.  Likely the culprit is that Troy tailored the previous ztf function to optimize runtime, but these hats functions are unfortunately not as speedy out of the box, and likely need to be optimized.  Suggestions from @troyraen on where to look for optimizations are documented in issue #420

### DIFF
--- a/light_curves/code_src/ztf_functions.py
+++ b/light_curves/code_src/ztf_functions.py
@@ -34,7 +34,7 @@ def ztf_get_lightcurves(sample_table, *, radius=1.0):
         ]
     )
 
-    # 2) Convert Astropy table → pandas → LSDB catalog, then persist sample only
+    # 2) Convert Astropy table → pandas → LSDB catalog
     sample_df = pd.DataFrame({
         'objectid': sample_table['objectid'],
         'ra_deg':   sample_table['coord'].ra.deg,
--- a/light_curves/code_src/ztf_functions.py
+++ b/light_curves/code_src/ztf_functions.py
@@ -11,6 +11,114 @@ import tqdm
 import helpers.scale_up
 from data_structures import MultiIndexDFObject
 
+import numpy as np
+import pandas as pd
+from astropy.table import Table
+import lsdb
+from dask.distributed import Client
+from data_structures import MultiIndexDFObject
+from upath import UPath
+
+# panstarrs light curves from hipscat catalog in S3 using lsdb
+
+
+def ztf_get_lightcurves(sample_table, *, radius=1):
+    """Searches ZTF hats files for light curves from a list of input coordinates.
+
+    Parameters
+    ----------
+    sample_table : `~astropy.table.Table`
+        Table with the coordinates and journal reference labels of the sources
+    radius : float
+        search radius, how far from the source should the archives return results
+
+    Returns
+    -------
+    df_lc : MultiIndexDFObject
+        the main data structure to store all light curves
+
+    note that other light curves in this notebook will have "time" as MJD instead of HMJD.
+    if your science depends on precise times, this will need to be corrected.
+
+    """
+
+    # read in the ZTF light curves to lsdb
+    ztf_lc = lsdb.read_hats(
+        UPath('s3://irsa-fornax-testdata/ZTF/dr23/lc/hats'),
+        #margin_cache=UPath('s3://stpubdata/panstarrs/ps1/public/hats/detection_10arcs', anon=True),
+        columns=["objectid","objra", "objdec", "hmjd", "filterid", "mag", "magerr", "catflags"]
+    )
+    # convert astropy table to pandas dataframe
+    # special care for the SkyCoords in the table
+    sample_df = pd.DataFrame({'objectid': sample_table['objectid'],
+                              'ra_deg': sample_table['coord'].ra.deg,
+                              'dec_deg': sample_table['coord'].dec.deg,
+                              'label': sample_table['label']})
+
+    client = Client()  #start a dask client for lsdb to use
+    
+    # convert dataframe to hipscat
+    sample_lsdb = lsdb.from_dataframe(
+        sample_df,
+        ra_column="ra_deg",
+        dec_column="dec_deg",
+        margin_threshold=10,
+        # Optimize partition size
+        drop_empty_siblings=True
+    )
+
+    # plan to cross match ZTF object with my sample
+    # only keep the best match
+    matched_lc = ztf_lc.crossmatch(
+        sample_lsdb,
+        radius_arcsec=radius,
+        n_neighbors=1,
+    )
+
+    # here is where the actual work gets done
+    # compute the cross match with object table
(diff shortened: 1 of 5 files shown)


troyraen (approved):
Thanks for working this out. A few comments below.

light_curves/code_src/ztf_functions.py
```diff
@@ -1,325 +1,110 @@
...
+        drop_empty_siblings=True
+    )
+
+    # plan to cross match ZTF object with my sample
+    # only keep the best match
+    matched_lc = ztf_lc.crossmatch(
+        sample_lsdb,
+        radius_arcsec=radius,
+        n_neighbors=1,
+    )
```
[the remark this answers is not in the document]
jkrick: I've re-worked the crossmatch to now have sample_table on the left and do the crossmatch 3 times, one with each band of the ztf light curve catalog with n_neighbors = 1.  This assures one match of each ZTF band with each row of sample_table , or less if there is no match within the supplied radius.  I chose this method instead of putting ztf on the left to avoid having many ztf matches per sample_table object.  There are many fine details which go into crossmatching.  This is one where we have to make a choice for the notebook, so I decided on the conservative, albeit slow choice.

jkrick:
Sorry @troyraen but I think it would probably be good for you to review the ztf function again since I changed it pretty significantly.

light_curves/light_curve_generator.md
```diff
@@ -256,7 +258,7 @@ print('WISE search took:', time.time() - WISEstarttime, 's')
 ```
 
 ### 2.4 MAST: Pan-STARRS
-The function to retrieve lightcurves from Pan-STARRS uses a version of both the object and light curve catalogs that are stored in the cloud and accessed using [lsdb](https://docs.lsdb.io/en/stable/).  This function is efficient at large scale (sample sizes > ~1000).
+The function to retrieve lightcurves from Pan-STARRS uses a version of both the object and light curve catalogs that are stored in the cloud and accessed using [LSDB](https://docs.lsdb.io/en/stable/).  This function is efficient at large scale (sample sizes > ~1000).
```
troyraen, line 261:
```suggestion
The function to retrieve lightcurves from Pan-STARRS uses [LSDB](https://docs.lsdb.io/) to access versions of the object and light curve catalogs that are stored in the cloud.  This function is efficient at large scale (sample sizes > ~1000).
```

Languaging and url suggestion similar to above.

troyraen (approved):
Thanks! Mostly minor suggestions below, but one that I'll call out is that I was able to reduce runtime by at least 1/3 by setting the `Client` kwarg `threads_per_worker=2`.

light_curves/requirements_light_curve_generator.txt
```diff
@@ -14,7 +14,7 @@ astroquery>=0.4.8.dev0
 acstools
 lightkurve
 alerce
-lsdb>=0.4.5
+lsdb>=0.5.1
```
troyraen, line 17:
```suggestion
lsdb>=0.5.2
```

Looks like using a HATS Collection requires 0.5.2 (for ZTF).

jkrick: done.

jkrick:
@bsipocz if you don't have any concerns, I think it is ready to merge.

bsipocz:
> @bsipocz if you don't have any concerns, I think it is ready to merge. 

let me have a look later today

light_curves/code_src/ztf_functions.py
```diff
@@ -1,325 +1,106 @@
...
+
+    # 2) Convert Astropy table → pandas → LSDB catalog
+    sample_df = pd.DataFrame({
+        'objectid': sample_table['objectid'],
+        'ra_deg':   sample_table['coord'].ra.deg,
+        'dec_deg':  sample_table['coord'].dec.deg,
+        'label':    sample_table['label'],
+    })
+    sample_lsdb = lsdb.from_dataframe(
```
bsipocz, line 44: would be a nice upstream feature to be able to ingest directly from an astropy table. Has that been raised?

light_curves/light_curve_generator.md
```diff
@@ -366,11 +368,11 @@ For sample sizes >~500 and/or improved logging and monitoring options, consider
 ```{code-cell} ipython3
 # number of workers to use in the parallel processing pool
 # this should equal the total number of archives called
```
bsipocz, line 370:
```suggestion
# this should equal the total number of archives called in the `pool` context below.
```

bsipocz (approved):
Overall looks good, and I don't have any substantial comments.

light_curves/code_src/ztf_functions.py
troyraen: That would be nice. I would be surprised if truly no one has requested it yet but I haven't heard anything about it. A quick look through their issues didn't turn up anything obviously related so probably worth opening an issue for.

jkrick: looks like @bsipocz did this.  Issue 780 over at lsdb

bsipocz: Yes, sorry, I forgot to xref it here: https://github.com/astronomy-commons/lsdb/issues/780

PR #423, opened 2025-05-19
Add rubin data to light curve generator

jkrick:
This code adds a cell to light curve generator to grab DP0 data from RSP by using a `pyvo` TAP query and `adql`.

It should not be run through CI because it requires a password and authentication token.  

I have added a lot of error raising around the authentication process to try to help people doing this to have readable errors.

Currently there is no data in DP0 for our Yang sample, so this PR should not have any effects on the notebook plotting.  It also only adds 6s to come to the conclusion that there is no data on our targets, so I have not changed the runtime statement.  

This code is not capable of handling this search at scale.  I have spoken with RSP to let them know that we would like to have table uploads enabled for TAP queries, and they know it is a priority.  I hope to see this when real data starts flowing to users.  Effectively this means this code is a placeholder for future data and capabilities.  

Do to all of the caveats above (authentication, no data, not able to run at scale), this code will definitely change in the future, but at least the process has begun.

Testing is possible by inserting the following code in a cell near the current Rubin cell to find sources which do have DP0 data:

```
#now find a sample which does have sources to test the code:
from astropy.coordinates import SkyCoord
import astropy.units as u
from rubin_functions import rubin_get_lightcurves, rubin_authenticate

# 1) Authenticate to DP0
rsp_tap = rubin_authenticate()

# 2) Grab a handful of known objects from the DP0 Object table
query = """
SELECT TOP 5 coord_ra, coord_dec, objectId
FROM dp02_dc2_catalogs.Object
WHERE detect_isPrimary = 1
"""
tbl = rsp_tap.run_sync(query).to_table()

# 3) Build  sample_table with the three required columns
coords = SkyCoord(
    ra=tbl['coord_ra'],
    dec=tbl['coord_dec'],
    unit=(u.deg, u.deg),   # only applied if your columns are plain floats
    frame='icrs'
)
sample_table = Table()
sample_table['coord']    = coords
sample_table['objectid'] = tbl['objectId']
sample_table['label']    = [f"obj{i}" for i in range(len(tbl))]

# 4) Now test  light-curve fetcher
lc = rubin_get_lightcurves(sample_table, search_radius=0.001)
print(lc.data)  # should be non‐empty if DP0 has data for these objects

```


light_curves/code_src/rubin_functions.py
```diff
@@ -0,0 +1,225 @@
...
+    #If there are no IDs, we don't need to search the light curve catalot
+    # and can just return an empty MultiIndexDFObject
+    if not objids:
+        # Create an empty MultiIndex 
+        mi = pd.MultiIndex.from_arrays(
+            [[], [], [], []],
+            names=['objectid', 'label', 'band', 'time']
+        )
+        empty_df = pd.DataFrame(index=mi)
+        return MultiIndexDFObject(empty_df)
```
troyraen, line 138:
```suggestion
    if not objids:
        return MultiIndexDFObject()
```

No need for anything else.

light_curves/code_src/rubin_functions.py
```diff
@@ -0,0 +1,225 @@
...
+        tbl['in_objid'] = in_objid
+        tbl['in_label'] = in_label
+        all_tables.append(tbl)
+
+    #in the case of an empty table, return an empty table
+    if not all_tables:
+        cols = ['coord_ra','coord_dec','objectId','in_objid','in_label']
+        # build a dict mapping each name to an empty list
+        empty_dict = {col: [] for col in cols}
+        return Table(empty_dict)
```
troyraen, line 103: Seems simpler to do:

```suggestion
    if not all_tables:
        return
```

Then after calling this in `rubin_get_lightcurves()`, do:

```python
if not obj_table:
    return MultiIndexDFObject()
```

and skip lines 131-138 in `rubin_access_lc_catalog()`.

troyraen (approved):
Thanks, fun to have Rubin data in here! I did a read-through and review. Comments below are mostly simplification suggestions; also a couple best practices and just questions out of curiosity. No blockers.

I have not actually tried to run it but I should / would like to at some point. LMK if you want me to do that before you merge, otherwise it won't happen this week.

jkrick (commented):
Thanks for the review and careful thought.  I think I have addressed everything. 

I am not in a hurry, so if you want to run it, please go ahead, I won't merge right away to give you that option.  I do want to say that there is almost no chance this function will be the final function that we will use once Rubin data starts flowing, so the future holds changes and therefore more chances for code review.  In my opinion, therefore, this code itself doesn't warrant a large amount of time investment.

bsipocz:
I would argue that anything rubin related should be not used/commented out from the default notebook, to ensure the notebook still runs for those who hasn't set up, or don't even have data access.

light_curves/code_src/rubin_functions.py
```diff
@@ -0,0 +1,225 @@
...
+        If the token file does not exist or is empty.
+    RuntimeError
+        If the TAP session cannot be created or the returned baseurl mismatches the expected URL.
+    """
+    # Read token from home directory
+    token_file = Path.home() / '.rsp-tap.token'    
+    if not os.path.exists(token_file):
+        raise FileNotFoundError(f"Token file not found: {token_file}")
+    with open(token_file, 'r') as f:
+        token = f.readline().strip()
```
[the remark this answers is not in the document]
troyraen: The one liner will close the file (https://docs.python.org/3/library/pathlib.html#pathlib.Path.read_text). But still, fine to leave as-is.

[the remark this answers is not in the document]
bsipocz: ahh, pathlib, thanks!

jkrick:
> I would argue that anything rubin related should be not used/commented out from the default notebook, to ensure the notebook still runs for those who hasn't set up, or don't even have data access.

yes, thank you, you are very right.  I commented out the lines which run the Rubin function and added text to say users need to uncomment once they have login/etc. setup.

light_curves/code_src/rubin_functions.py
```diff
@@ -0,0 +1,210 @@
...
+
+    Returns
+    -------
+    combined : astropy.table.Table
+        Combined table with columns ['coord_ra', 'coord_dec', 'objectId',
+        'in_objid', 'in_label'] tagging each DP0 Object match with the input sample info.
+    """
+    #could consider parallelizing this but for now we wait for TAP table upload and real data
+    all_tables = []
+    for row in sample_table:
```
bsipocz, line 73: do they support table upload? that way you could avoid manually looping over the coordinates and have just one query.

bsipocz: (this is for follow-up, easily point beyond the scope of this PR)

light_curves/code_src/rubin_functions.py
```diff
@@ -0,0 +1,210 @@
...
+        dec = row['coord'].dec.deg
+        in_objid = row['objectid']
+        in_label = row['label']
+        
+        query = (f"""
+        SELECT coord_ra, coord_dec, objectId
+        FROM dp02_dc2_catalogs.Object
+        WHERE CONTAINS(
+            POINT('ICRS', coord_ra, coord_dec),
+            CIRCLE('ICRS', {ra:.6f}, {dec:.6f}, {search_radius})
```
bsipocz, line 84: is it necessary to cut the precision here? (I mean it makes sense, but also why manually cut if not needed)

light_curves/light_curve_generator.md
```diff
@@ -349,6 +350,31 @@ df_lc_icecube = icecube_get_lightcurves(sample_table, icecube_select_topN=3)
...
+
+### 3.3 Rubin Data 
+Before running this section, you need to have both 1) a login to the Rubin Science Platform (RSP) and 2) an authenticating token setup in your Fornax home directory. 
+
+1) To see if you have Rubin data rights, and if you do, to get that login setup, follow these [instructions](https://rsp.lsst.io/guides/getting-started/get-an-account.html) 
+
+2) After logging in to RSP, follow these [instructions](code_src/rsp_token_instructions.txt) to get a token and store it in your home directory.
+
+
+This code will not work without the above information, so we have commented it out for now.  Uncomment when you are ready to proceed.
```
bsipocz, line 363: maybe uncomment the whole cell, or change the cell type? 
(I think that maybe easier than figuring out the few lines peppered through that needed to be uncommented. OTOH, if it's peppered through the user may actually get some insight which line is doing the actual work 😅 )

bsipocz (approved):
minor comments, otherwise this all looks good to me.

(and CI should be green after your latest commit that disabled the Rubin access)

light_curves/code_src/rubin_functions.py
jkrick: I have requested table upload with RSP, Gregory knows this and told me this was a priority for them.  For now, it doesn't exist.

light_curves/light_curve_generator.md
jkrick: Done. Commented out the whole cell and added explicit instructions in that cell.  I figured changing cell type was just going to annoy either you, or someone else who doesn't use jupyter to do their work.  Commented lines seem a more universal way.

bsipocz: nope, I'm OK with changing the type, but not sure which one is more user friendly.

Also, unrelated but this reminded me that we can hide cells or make them hidden (with an option to dropdown/extend) with metadata. I think at some point we can start using these features, if it makes sense to you.

PR #429, opened 2025-06-04
Jwst spectra switch to astroquery's MastMissions class

jkrick:
This PR switches the MAST JWST astroquery spectroscopy work from their class Observations to MastMissions.  I started out trying to make the JWST query faster because its speed is limiting my ability to use it on larger samples of even a few tens (which I want for the spherex notebook).  I am currently confused and stuck and am not seeing an improvement in speed, so pushing this now in hopes to get some feedback from someone who is not confused and stuck.

This notebook's original runtime estimate from 2024 December is 17 minutes for a sample of 10 galaxies, and a set of maybe 8 archives.  Right now, the jwst_get_spectra function is taking over an hour alone.    The switch to MastMissions is not improving runtime.  

Things I don't understand:

1. Why is this taking so much longer now when this function has not changed in a significant way?  I did make one change in Apr 2025 to switch one line from:
`data_products_list = Observations.get_product_list(query_results)`
to 
`data_products = [Observations.get_product_list(obs) for obs in query_results]
data_products_list = vstack(data_products)`
which at the time removed a timeout problem we were having.  I do not have Runtime estimates for that PR, but I am guessing the query was already slow or I wouldn't have been trying to remove the timeout.

2. Even stranger, this current PR using MastMissions returns different results than the main branch!!!!  The target COSMOS1 is my test galaxy.  The Observations based query doesn't find spectra for COSMOS 1 but the MastMissions based query does.  
3. FYI for MAST: column name FLUX_ERROR changed to ERROR in the switch between Observations and MastMissions

FWIW I didn't touch the HST search function, and that one is chugging along just fine, not lightning fast, but not over an hour (think it was ~4 min.)


jkrick:
@snbianco This PR is the result of your super useful responses to my helpdesk questions over the last week, but now, when putting this into practice, I  am having trouble figuring out the best way to efficiently access the JWST archive for a random list of targets.  I would very much appreciate your input.

spectroscopy/code_src/mast_functions.py
```diff
@@ -10,11 +10,130 @@
 import pandas as pd
 from astropy.table import Table
 from astropy.table import vstack
-from astroquery.mast import Observations
+from astroquery.mast import Observations, MastMissions
 from specutils import Spectrum1D
 
 from data_structures_spec import MultiIndexDFObject
 
+def JWST_get_spec_helper(sample_table, radius_arcsec, download_dir, verbose=False, delete_downloaded_data=True):
```
snbianco, line 18: How big do we expect `radius_arcsec` to be? A bigger search radius will return more observations and majorly increase the call to fetch products.

spectroscopy/code_src/mast_functions.py
```diff
@@ -10,11 +10,130 @@
...
+        objid = row['objectid']        # unique ID
+        if verbose:
+            print(f"MM Querying JWST around {label} ({coord.to_string('hmsdms')})")
+
+        # cone search for JWST datasets
+        datasets = m.query_region(
+            coord,
+            radius=(radius_arcsec * u.arcsec).to(u.deg).value, 
+            exp_type='MIR_LRS-FIXEDSLIT,MIR_LRS-SLITLESS,MIR_MRS,NRC_GRISM,NRC_WFSS,NIS_SOSS,NIS_WFSS,NRS_FIXEDSLIT,NRS_MSASPEC',  # Spectroscopy data
+            instrume='!FGS',  # Any instrument except FGS
```
snbianco, line 58: This filter is probably unneeded given that you're already filtering by exposure type, and none of the options correspond to FGS.

spectroscopy/code_src/mast_functions.py
```diff
@@ -10,11 +10,130 @@
...
+ 
+        products = Table()
+        for batch in batches:
+            # Get the products for each batch
+            print(f"Getting products for batch of {len(batch)} products")
+            batch_prods = m.get_product_list(batch)
+ 
+            # Append the results to the products table
+            products = vstack([products, batch_prods])
+
```
snbianco, line 79: Because the bottleneck for this code is in `get_product_list`, adding parallelization here may help to speed things up.

This should go outside the loop:
```
def fetch_products(batch):
    print(f"Fetching products for batch of {len(batch)} products")
    try:
        return m.get_product_list(batch)
    except Exception as e:
        print(f"Error fetching products for batch: {e}")
        return Table()
```
And this should replace lines 71-78:

```
with ThreadPoolExecutor(max_workers=None) as executor:
    futures = [executor.submit(fetch_products, batch) for batch in batches]
    batch_tables = []
    for future in as_completed(futures):
        result = future.result()
        if len(result) > 0:
            batch_tables.append(result)

products = vstack(batch_tables) if batch_tables else Table()
```

snbianco: With a radius of 5 arcseconds, this performs the query and fetches all the products in under 2 minutes on my machine.

jkrick (commented):
Thanks for the suggestions @snbianco!

Are there any other speedups you could recommend?  For my full sample of 10 galaxies, with downloads, the JW cell is still running after 37min(and not done yet), which is just painfully long.   I already dropped batch size to 100 to try to optimize the number of batches with the number of workers, which helped a bit.  

@afaisst if we can't get the time down lower when pulling all available JWST data, what do you think about truncating the list of products so we only download and plot the first 10 or 100 JWST products?  We are already median stacking them for the plot, so we could add a note about not downloading everything and direct users how to dig deeper if they want more data.

Also, @snbianco what do you make of the Observations query in the main branch returning different results for our targets?

snbianco:
I'm sorry for the delay! I just got back from AAS 246 and some PTO.
 
One question I have is how necessary is it to actually download files for this workflow? It's not supported for `MastMissions` yet, but `Observations` can return cloud URIs for data products which can then be accessed directly from S3. This won't be helpful for the calls to `get_product_list`, but will probably shave off a good amount of time spent downloading the files.
 
As far as why the two queries are returning different results, `Observations` and `MastMissions` use two different APIs that access different databases. The data for the supported missions in `MastMissions` should reflect what is returned in `Observations`, but there is a "lag" between them so that newly added data shows up in the `MastMissions` database before it shows up in the `Observations` database. Are you still finding that the results are conflicting?

jkrick:
closing this because it is superseded by #438

PR #442, opened 2025-07-10
ENH: adding Euclid spectra

xoubish:
This has a Euclid function consistent in style to the other ones. It queries IRSA's MER catalog (cone search) for RA/DEC of the object and given the object_id gets the 1d spectrum. Q1 MER and Spectra file are hardcoded for now and can be changed to later releases. 
I also added 1 galaxy to the sample with Euclid spectra so this can be tested.


spectroscopy/code_src/Euclid_functions.py
```diff
@@ -0,0 +1,123 @@
...
+from astroquery.ipac.irsa import Irsa
+
+from data_structures_spec import MultiIndexDFObject
+import warnings
+from astropy.utils.exceptions import AstropyWarning
+
+warnings.simplefilter('ignore', category=AstropyWarning)
+
+
+def get_coord_from_objectid(object_id, table_mer="euclid_q1_mer_catalogue"):
```
jkrick, line 16: Does this function get used?  This does an IRSA.query_tap but in the main function you have an IRSA.query_region.

spectroscopy/code_src/Euclid_functions.py
```diff
@@ -0,0 +1,123 @@
...
+    WHERE object_id = {object_id}
+    """
+    try:
+        result = Irsa.query_tap(adql_query).to_table()
+        if len(result) == 0:
+            print(f"No match found for object_id {object_id}")
+            return None
+        ra, dec = result[0]["ra"], result[0]["dec"]
+        return SkyCoord(ra=ra * u.deg, dec=dec * u.deg)
+    except Exception as e:
```
jkrick, line 28: I'll let @troyraen or @bsipocz  comment, but I know they don't like these very broad "except Exceptions" and would rather have a specific exception caught if it is an issue.   Here and below..

jkrick (approved):
Thanks for adding Euclid to this notebook!  This looks good to me, just a few comments.
- please add a text bullet for Euclid under section 2.1
- please update the runtime.  This just took ~18 minutes for me (mostly still JWST) on the same server listed already in the runtime comment.
- please add numpydocs style docstrings - and mention that this function is pulling data from the cloud.

spectroscopy/code_src/Euclid_functions.py
xoubish: This I used to find RA/DEC of the test object I showed on the notebook. Its not currently used somewhere but its useful to have, I think

jkrick:
> @jkrick I add ~2 minutes to the time, as I mentioned in another issue JWST took more than an hour for me so I can't time the whole thing.

when you pulled in the main branch you got the updated JWST function which now takes something like 10 min on the medium fornax server (with the setup listed in the runtime section).  The whole thing including Euclid took 18 minutes for me earlier this afternoon, so you can just use that estimate.

spectroscopy/code_src/Euclid_functions.py
bsipocz: yes, please be specific with exceptions. Also, I wonder if we even need a try/except here as you already did a check for result length --> propagate the issue to the user as it's a real one rather than quietly swallow it.

spectroscopy/code_src/Euclid_functions.py
```diff
@@ -0,0 +1,185 @@
...
+        print(f"Query failed for object_id {object_id}: {e}")
+        return None
+
+
+def Euclid_get_spec(sample_table, search_radius_arcsec):
```
bsipocz, line 60: snake case should be all lower case, no matter if it has a proper noun

```suggestion
def euclid_get_spec(sample_table, search_radius_arcsec):
```

spectroscopy/code_src/Euclid_functions.py
```diff
@@ -0,0 +1,185 @@
...
+    Retrieve Euclid 1D spectra for sources in a sample table using IRSA TAP cloud services.
+
+    This function performs a cone search on the Euclid MER catalog via IRSA (a cloud-based service),
+    retrieves associated 1D spectral file URIs from a TAP-accessible association table,
+    downloads the spectra FITS files, extracts flux and error, and packages them
+    into a MultiIndexDFObject.
+
+    Parameters
+    ----------
+    sample_table : list of dict
```
bsipocz, line 71: make it a table as you call it a table as well as it will be then consistent with the rest of the API

spectroscopy/code_src/Euclid_functions.py
```diff
@@ -0,0 +1,185 @@
...
+        try:
+            tab = Irsa.query_region(
+                coordinates=coord,
+                catalog=table_mer,
+                spatial="Cone",
+                radius=search_radius_arcsec * u.arcsec
+            )
+        except Exception as e:
+            print(f"Query failed for {stab['label']}: {e}")
+            continue
```
bsipocz, line 114: Don't swallow error, and practically never do an `except Exception`, it is just in very rare cases the appropriate option.

bsipocz: This comment stands for all the remaining try/excepts, try to remove them, and definitely change the exception type for the ones remaining

spectroscopy/spectra_generator.md
```diff
@@ -317,8 +329,12 @@ labelled by sample number.
...
-               bin_factor=5,
+               bin_factor=1,
                show_nbr_figures=10,
                save_output=False,
                )
 ```
+
+```{code-cell} ipython3
+
+```
```
bsipocz, line 340: remove the extra cell

spectroscopy/code_src/Euclid_functions.py
```diff
@@ -0,0 +1,185 @@
...
+from astropy.coordinates import SkyCoord
+from astropy.table import QTable
+from astropy.io import fits
+from astroquery.ipac.irsa import Irsa
+
+from data_structures_spec import MultiIndexDFObject
+import warnings
+from astropy.utils.exceptions import AstropyWarning
+
+warnings.simplefilter('ignore', category=AstropyWarning)
```
bsipocz, line 13: Please do not add this, we use warnings for cases where/when the user need to see them. If there are cases where they need to be dealt with in a code module, do deal with that at the specific spot, but let the rest of the warnings go  through the end user

bsipocz (changes requested):
the `except Exceptions` will need to cleaned up prior merging, also rule of thumb, avoid blanket warning catching

spectroscopy/code_src/Euclid_functions.py
xoubish: I made the errors more explicit let me know if it still is not ok. Or go ahead and fix if you prefer, thanks.

spectroscopy/code_src/Euclid_functions.py
bsipocz: OK, these are not something the user can do much about but the typical cases that need to be reported upstream to the data provider to fix. I'll add a specific filter to get these, but only these out rather than a blanket filter.

(also, these have been reported before we released euclid, I just need to find the relevant jira or confluence page and check in with Anastasia to see when a fix will be out).

.virtual_documents/spectroscopy/spectra_collector.md
bsipocz, line 1: Not sure what generates this directory, but the directory will need to be added to the `.gitignore` file.

spectroscopy/code_src/Euclid_functions.py
bsipocz: OK, so I would add these very explicit lines, so we know why and what to ignore and when this workaround can be removed.

```
# We filter these specific warnings out as the end user can do nothing to about them, they have to be fixed at the data providers level. 
# Remove the filters once they are not used in a newer data release
warnings.filterwarnings("ignore", "'Number' did not parse as fits unit", u.UnitsWarning)
warnings.filterwarnings("ignore", "'erg/s/cm2/Angstrom' contains multiple slashes", u.UnitsWarning)
warnings.filterwarnings("ignore", "'erg2/s2/cm4/Angstrom2' contains multiple slashes", u.UnitsWarning)
```

xoubish: thank you and done.

spectroscopy/spectra_collector.md
```diff
@@ -4,12 +4,11 @@ jupytext:
...
     format_version: 0.13
-    jupytext_version: 1.17.2
+    jupytext_version: 1.16.7
 kernelspec:
-  name: py-spectra_generator
-  display_name: py-spectra_generator
+  display_name: Python 3 (ipykernel)
   language: python
-
+  name: python3
```
bsipocz, line 11: Jessica has just updated these, please keep the updated values

spectroscopy/spectra_collector.md
```diff
@@ -317,8 +329,12 @@ labelled by sample number.
...
-               bin_factor=5,
+               bin_factor=1,
                show_nbr_figures=10,
                save_output=False,
                )
 ```
+
+```{code-cell} ipython3
+
+```
```
bsipocz, line 340: remove new cell

spectroscopy/code_src/Euclid_functions.py
```diff
@@ -0,0 +1,160 @@
...
+            if not (len(wave) == len(signal) == len(var) == len(mask)):
+                raise ValueError(f"Spectrum array length mismatch for object_id {object_id}")
+
+            valid = (mask % 2 == 0) & (mask < 64)
+
+            wave = wave[valid]
+            flux = signal[valid] * fscale * u.erg / u.s / u.cm**2 / u.angstrom
+            error = np.sqrt(var[valid]) * fscale * flux.unit
+        except (KeyError, ValueError) as e:
+            raise RuntimeError(f"Could not parse spectrum for object_id {object_id}: {e}")
```
bsipocz, line 145: Raising specific exceptions for different errors is a feature, I still fail to see the value all these try/excepts and reraising errors bring to the table (and e.g. this one even override the error in L137)

bsipocz (approved):
I think this will need a rebase, that would resolve e.g the notebook metadata updates. Or you can change those back manually.

Either case, once the fixes made, I don't need to rereview again, this can go ahead. Thanks!

spectroscopy/code_src/Euclid_functions.py
xoubish: you are totally right, I changed the raise errors to prints, so if any of the issues happen the user knows why the spectra is not there but still continues to the next.

PR #443, opened 2025-07-10
Add analytical data search notebook

adrianlucy:
This notebook from MAST for Fornax, "Beyond metadata: analytical data search in the cloud with JWST spectral cubes", answers a data-intensive archival search query ("_I want to find JWST spectral image cubes containing Fe II emission from jets launched by young stellar objects_"), with broader pedagogical goals related to large-scale target coordinate searches beyond the capabilities of archive services, parallelizing with dask, algorithmically searching for select data characteristics, and loading data from cloud URIs into memory.

The focus is on JWST spectral cubes from MAST, but I hope that the strategies taught will be of more general use for other missions and archives. I also touch briefly on SOFIA data from IRSA at the end, and hope to maybe switch that to SPHEREx in a future PR when more SPHEREx data is available.

The notebook has been reviewed internally by several staff internal to MAST, their feedback has been incorporated, and we are ready for review by the Fornax team!

A couple practical notes:
- You may need to pip install the requirements file. I've recently noticed that the Default Astrophysics kernel in Fornax no longer has boto3 and pyarrow, but it used to as recently as a couple weeks ago. I'm wondering if this might be accidental; I'll file an issue later today or tomorrow.
- In any case, you need a more recent astroquery version than the one currently installed in Fornax (last time I checked). This is discussed in the notebook.
- [An executed and rendered copy is temporarily available in a temporary repo](https://github.com/adrianlucy/notebook_drafts/blob/main/analytical_data_search/analytical_data_search.ipynb), in case it helps illuminate the issue if something goes wrong.

### DIFF
--- a/data_search/README.md
+++ b/data_search/README.md
@@ -1,0 +1,7 @@
+# Beyond metadata: analytical data search in the cloud with JWST spectral cubes
+
+Not all archival search queries can be answered with metadata alone. In this use case, we start from an intimidating user query: "I want to find JWST spectral image cubes containing Fe II emission from jets launched by young stellar objects." In the course of answering this query, we learn to think carefully about the order of operations for coordinate searches on large numbers of targets, to perform parts of this search locally when there are too many targets for archive services to handle, to parallelize our data processing, to load data into memory instead of storage, and to load data from the cloud instead of on-premise storage.
+
+Although this tutorial focuses on JWST spectral cubes, it is intended to teach general strategies that will be useful for answering other data-intensive search queries unrelated to spectral cubes, in any of the Fornax archives. We design some simple algorithms to search for extended emission in spectral cubes, but these algorithms are designed to be replaceable by more sophisticated tools tailored to your science case.
+
+Note: the notebook (analytical_data_search.md) required the files in the code_src/ directory in order to run.
--- a/data_search/analytical_data_search.md
+++ b/data_search/analytical_data_search.md
@@ -1,0 +1,1144 @@
+---
+jupytext:
+  text_representation:
+    extension: .md
+    format_name: myst
+    format_version: 0.13
+    jupytext_version: 1.17.2
+kernelspec:
+  display_name: python3
+  language: python
+  name: python3
+---
+
+# Beyond metadata: analytical data search in the cloud with JWST spectral cubes
+
++++
+
+For this notebook, we recommend using the [Fornax science platform](https://nasa-fornax.github.io/fornax-demo-notebooks/documentation/README.html) with Server Type = "Large - 64GB RAM/16 CPU" and Environment = "Default Astrophysics". As of June 2025, running this notebook once on that server will take about 10 minutes.
+
+We assume that you are running this notebook in Fornax, but it will also still work (in order of preference) in another AWS-based cloud science platform, a non-AWS cloud science platform, and even on your local computer — perhaps at slower speeds, depending on your computing resources and internet connection.
+
++++
+
+## Learning goals
+
+The overarching intent of this notebook is to learn about how to answer data-intensive archival search queries in astronomical data archives like [MAST](https://archive.stsci.edu/). We'll have to think about:
+* when to parallelize our data processing,
+* the order of operations for performing cone searches on large numbers of targets,
+* when to load data into memory versus downloading to storage,
+* the advantages and constraints of a cloud science platform,
+* and when to load data from cloud storage instead of on-premise storage.
+
+## Introduction
+
+As you read through a recent paper by [Assani et al. 2025](https://arxiv.org/pdf/2504.02136), imagine that you find yourself consumed by a single thought:
+
+> I want to find more JWST spectral image cubes containing Fe II emission from jets launched by young stellar objects (YSOs).
+
+You may now be experiencing some despair. "This observation contains a JWST spectral cube" is certainly something readily available from MAST's metadata tables, but "this file contains Fe II transitions" is certainly not. So instead, you need to dig beyond the metadata and into the data of thousands of files. That may sound difficult, but it can be done!
+
+## Table of Contents
+
+More granularly, this notebook is comprised of the following sections:
+
+- [Imports](#Imports)
+- [1. Finding JWST spectral cubes of YSOs](#1.-Finding-JWST-spectral-cubes-of-YSOs)
+    - [1.1 Querying SIMBAD for YSOs](#1.1-Querying-SIMBAD-for-YSOs): *Use `astroquery` to ask SIMBAD for all the YSOs it knows about.*
+    - [1.2 Querying MAST for all JWST spectral cube observations](#1.2-Querying-MAST-for-all-JWST-spectral-cube-observations): *Use `astroquery` to ask MAST for all JWST spectral cube observations.*
+    - [1.3 Crossmatching YSO coordinates to MAST footprints](#1.3-Crossmatching-YSO-coordinates-to-MAST-footprints): *Write custom functions to crossmatch the SIMBAD YSOs to the footprints of the MAST observations, bypassing an astroquery service call that would take to long for this number of targets.*
+    - [1.4 Refining and parallelizing the footprint crossmatch](#1.4-Parallelizing-the-footprint-crossmatch-with-dask): *Improve and parallelize the crossmatch technique, with gnomonic projection and dask.*
+- [2. Finding Fe II emission jets](#2.-Finding-Fe-II-emission-jets)
+    - [2.1 Retrieving sample data from AWS](#2.1-Retrieving-sample-data-from-AWS): *Retrieve MAST data from the AWS cloud into memory.*
+    - [2.2 Searching for emission lines in spectral cubes](#2.2-Searching-for-emission-lines-in-spectral-cubes): *Write custom functions to algorithmically analyze thousands of spectral cube data files, and assess which of these are likely to exhibit extended Fe II emission.*
+    - [2.3 Parallelizing the search for emission lines in spectral cubes](#2.3-Parallelizing-the-search-for-emission-lines-in-spectral-cubes): *Parallelize the spectral line search and apply it to thousands of data files.*
+- [3. Postscript: Transposing to IRSA missions](#3.-Postscript:-Transposing-to-IRSA-missions): *Exercise for transposing these techniques to SOFIA data.*
+- [About this notebook](#About-this-notebook)
+
++++
+
+## Imports
+
++++
+
+<span style="color: purple;">Temporary note: pending incorporation of a recent `astroquery` development version into the default Fornax kernel, we need to install the latest development version of astroquery. We can suppress the voluminous output here with `%%capture`, which will store the cell's output in a variable `captured` (accessed through `captured.stdout` and `captured.stderr`). Uncomment and run this cell:</span>
+
+```{code-cell} ipython3
+# %%capture captured
+# %pip install git+https://github.com/astropy/astroquery.git
+```
+
+**You may need to restart your kernel now for that to take effect.**
+
++++
+
+Otherwise, all dependencies are included in the Fornax "Default Astrophysics" environment. If you are not working in Fornax, you can uncomment and run the following cell to install these dependencies into your environment. You may need to restart your kernel after pip-installing new packages, before running the imports.
+
+```{code-cell} ipython3
+# !pip install -r requirements_analytical_data_search.txt
+```
+
+```{code-cell} ipython3
+# Query databases
+from astroquery.simbad import Simbad
+from astroquery.mast import Observations
+from astroquery.ipac.irsa import Irsa
+
+# Crossmatch coordinates to footprint polygons
+from matplotlib.path import Path
+from astropy.wcs import WCS
+
+# Create plots
+import matplotlib.pyplot as plt
+from matplotlib.gridspec import GridSpec
+from matplotlib.colors import PowerNorm
+
+# Wrangle data 
+from astropy.table import Table, Column, Row
+from astropy.io import fits
+import pandas as pd
+import numpy as np
+
+# Wrangle units
+from astropy.coordinates import SkyCoord
+import astropy.units as u
+
+# Detect extended emission in specific wavebins
+from scipy.signal import find_peaks
+from scipy.ndimage import label
+import bisect
+
+# Utility functions
+import os
+import sys
+import copy
+
+# Parallelize
+import dask
+import dask.dataframe as dd
+```
+
+We'll also need a couple local functions written specifically for this notebook. These are located in the `code_src` directory, and you can read them there. However, they're pretty long, and their detailed implementation isn't critical to understand. As long as you understand their inputs and outputs, which we'll discuss later, you can write better functions to replace them, tailored to your own science case.
+
+```{code-cell} ipython3
+# Local code imports
+sys.path.append('code_src/')
+
+from extension_vs_wavebin import extension_vs_wavebin
+from detect_spikes import detect_spikes
+```
+
+Let's also turn on `enable_cloud_dataset` for `astroquery.mast.Observations`. This will allow us to fetch the cloud locations for data products and access data directly from the cloud, instead of retrieving data from MAST's on-premise servers.
+
+```{code-cell} ipython3
+Observations.enable_cloud_dataset()
+```
+
+Later, we'll be evaluating boolean conditionals on values that could be `None`, the pandas equivalent `pandas.NA`, or a normal value, and we'll also need this little function. It's pretty uninteresting, so let's get it out of the way now!
+
+```{code-cell} ipython3
+# Check that x is not None and is not pd.NA
+def exists(x):
+    try:
+        return x is not None and not pd.isna(x)
+    except TypeError:
+        # For types that can't be checked with pd.isna
+        return x is not None
+```
+
+## 1. Finding JWST spectral cubes of YSOs
+
++++
+
+Let's ignore the emission line issue and tackle the easiest parts of our driving question first: I want to find more JWST spectral cubes of young stellar objects.
+
++++
+
+### 1.1 Querying SIMBAD for YSOs
+
++++
+
+At this time, MAST doesn't support reliable object classification search. So let's go to the experts in astronomical object cataloging, SIMBAD, which we can access through `astroquery.simbad.Simbad.query_tap`.
+
+In the cell below, we search for all SIMBAD-catalogued objects labeled as YSOs (`otype='Y*O'`) or as any of the descendant sub-concepts of YSOs, like T Tauri stars (`otype='Y*O..'` to retrieve both explicitly labeled YSOs and their subtypes).
+
+```{code-cell} ipython3
+# This will take a minute or two.
+yso_table = Simbad.query_tap("SELECT * FROM basic WHERE otype='Y*O..'", maxrec=1000000)
+```
+
+```{code-cell} ipython3
+print(f'We found {len(yso_table)} YSOs.')
+```
+
+Let's make a list of their sky coordinates:
+
+```{code-cell} ipython3
+yso_coords = []  # List of tuples, each tuple a coordinate.
+
+for row in yso_table:
+    # Check if the coordinates exist
+    if isinstance(row['ra'], float) and isinstance(row['dec'], float):
+        # Append the coordinates to the list
+        yso_coords.append((row['ra'], row['dec']))
+    
+print(f'{len(yso_coords)} of these YSOs have good coordinates.')
+```
+
+### 1.2 Querying MAST for all JWST spectral cube observations
+
++++
+
+Now we want to get all the JWST spectral cube observations in MAST whose sky footprints overlap with any of these YSO coordinates. The natural tool is the `astroquery.mast.Observations` class, which gives programmatic access to MAST's multi-mission archive.
+
+However, `astroquery.mast` is not currently set up to quickly run a multi-target query for tens of thousands of sky coordinates. So in order to get things done in a reasonable amount of time, we need to think about the order of operations here.
+
+It turns out that, at the time of writing, there are only a few tens of thousands of JWST spectral cubes across the whole sky. So let's retrieve *all* of those first, excluding only those with a `calib_level` of `-1` (planned observations that haven't yet executed) and non-public datasets.
+
+```{code-cell} ipython3
+# This will typically take anywhere from a few seconds to a minute or so,
+# depending on how many people are using MAST right now.
+# Rarely, if MAST is overloaded, it may take several minutes or time out.
+
+jwst_obstable = Observations.query_criteria(dataproduct_type='cube',
+                                                 obs_collection='JWST',
+                                                 dataRights='PUBLIC',  # Limit to public data
+                                                 calib_level = [0, 1, 2, 3, 4])  # Exclude planned observations
+```
+
+```{code-cell} ipython3
+print(f'We found {len(jwst_obstable)} JWST spectral cube observations.')
+```
+
+`Astroquery` has retrieved a wide variety of metadata for each of these observations. Let's look at the first couple rows:
+
+```{code-cell} ipython3
+jwst_obstable[0:2]
+```
+
+### 1.3 Crossmatching YSO coordinates to MAST footprints
+
++++
+
+Now, with that table as our starting point, let's match our YSO coordinates to the sky footprints of all these JWST observations. We already have everything from MAST that we need to do this, so we don't need to run another query. The footprint of an observation is in the `s_region` column of the observations table output by `astroquery`:
+
+```{code-cell} ipython3
+# For example...
+print(jwst_obstable[0]['s_region'])
+```
+
+This `s_region` denotes the vertices of a polygon in right ascension and declination. We can make this string more parsable with this function:
+
+```{code-cell} ipython3
+def parse_polygon(s_region):
+    """
+    Extract coordinates from an s_region string.
+
+    Input parameters
+    ----------
+    s_region : string - s_region from cross-mission archive
+               As written, this s_region must be a single polygon,
+               which is the case for all JWST and SOFIA observations,
+               but not all missions in the Fornax archives.
+
+    Returns
+    ----------
+    coord_array : numpy array - parsed representation of the s_region
+    """
+
+    if 'POLYGON ICRS' in s_region:
+        coords = list(map(float, s_region.replace("POLYGON ICRS", "").strip().split()))
+    elif 'POLYGON' in s_region:
+        coords = list(map(float, s_region.replace("POLYGON", "").strip().split()))
+
+    # Check if the polygon is closed, with the last vertex identical to the first vertex.
+    # If not, append a copy of the first vertex to the end of the list.
+    if (coords[-2], coords[-1]) != (coords[0], coords[1]):
+        coords.append(coords[0])
+        coords.append(coords[1])
+
+    # Create a numpy array listing a coordinate tuple for each vertex
+    coord_array = np.array([(coords[i], coords[i + 1]) for i in range(0, len(coords), 2)])
+
+    return(coord_array)
+```
+
+```{code-cell} ipython3
+# For example...
+print(parse_polygon(jwst_obstable[0]['s_region']))
+```
+
+Now, we'll parse our polygons and use `Matplotlib`'s `Path` class to test whether any of our YSO coordinates are in each observation's sky footprint.
+
+Doing this without parallelizing, as in the following cell, would take about 6 or 7 minutes. On Fornax, if you uncomment this cell and open the "Kernel usage" tab to your right while it runs, you'll notice that you're using only a small percentage of the total CPU capacity that you have access to (e.g., less than 10% if you selected the 16-CPU server option when you launched Fornax). Feel free to click the `Interrupt the kernel` stop button on your notebook, because there's a better way!
+
+```{code-cell} ipython3
+# Since this cell is an example of what not to do,
+# we'll also omit the nuance of gnomonic projection: see the next section.
+
+# yso_jwst_obs = []  # Instantiate a list of observations whose footprints overlap with any of our YSO coordinates.
+
+# for row in jwst_obstable:  # Looping through all JWST spectral cube observations...
+#     polygon = parse_polygon(row['s_region'])  # Parse the polygon
+
+#     polygon_path = Path(polygon, closed=True)  # Convert to matplotlib Path
+#     is_inside = polygon_path.contains_points(yso_coords).any()  # Test whether any coordinate from yso_coord is inside this observation's polygon
+
+#     if is_inside:
+#         yso_jwst_obs.append(row)
+
+# yso_jwst_obstable = Table(rows=yso_jwst_obs, names=jwst_obstable.colnames)  # Convert back to an astropy table.
+
+# print(f'We found {len(yso_jwst_obstable)} JWST spectral cube observations whose footprints overlap with YSO coordinates.')
+```
+
+### 1.4 Refining and parallelizing the footprint crossmatch
+
++++
+
+That cell would take a long time to run, so let's parallelize this code to save time! First we'll need to define a function that does what we did above, for each observation, so that we can pass this function into our parallelization tool.
+
+While we're at it, we'll also add in a subtle nuance; *RA* and *dec* are spherical coordinates, so the segments of a footprint polygon should be great circles on the celestial sphere, not straight lines. However, they can be treated as straight lines if we project into a 2D cartesian coordinate system, to an excellent approximation in the vicinity of the center of the projection. An appropriate choice is the gnomonic (`TAN`) projection, which is the projection used in the JWST FITS images from which `s_region` values were derived. Before checking which YSO coordinates are inside a particular observation's footprint, we'll convert both the footprint's polygon vertices and the YSO coordinates to a gnomonic projection centered on that specific observation's coordinates.
+
+We mentioned that the projection issue is a subtle nuance; in fact, at the time of writing, you could omit the projection step and still get the same number of observations matched. YSO coordinates tend not to be at the very edges of the observation footprints, after all. Nevertheless, to avoid any future problems, let's do things properly! On the Fornax server recommended above, this will only add about 30 seconds to our parallelized processing.
+
+Note that we are not accounting for proper motions—and neither would `astroquery.mast` cone search, if we were using it. See [High PM Stars in MAST](https://outerspace.stsci.edu/display/MASTDOCS/High+PM+Stars+in+MAST) if this is a concern for your science case.
+
+```{code-cell} ipython3
+def check_points_in_polygon(row, coordinates_to_test):
+    """
+    Check if coordinates_to_test are in the s_region polygon for an observation.
+
+    Input parameters
+    ----------
+    row : a row of an astropy table of observations from astroquery
+    coordinates_to_test: python list of celestial coordinate tuples
+
+    Returns
+    ----------
+    Either `row` (same as input) if a match was found,
+    or an empty row of NaNs if a match was not found.
+    """
+
+    # Make WCS for a gnomonic (TAN) projection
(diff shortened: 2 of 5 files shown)


adrianlucy:
> I would suggest a different title, as I had no idea what to expect from the notebook when I first saw its title. Maybe something like: "detect_emission_spectral_cubes" or "detect_FE_YSOs" or "search_YSO_cubes" or "FE_emission_YSOs" or ??
Thank you @jkrick! Just to check, when you ask for a different title, are you only referring to the filename, or do you also suggest changing the in-notebook title ("_Beyond metadata: analytical data search in the cloud with JWST spectral cubes_")?

Also, do you want the directory holding the notebook to be named the same as the file? Currently, the directory name is even more generic - I guess I was imagining that maybe you'd want to group this notebook with ones sharing similar goals in the future, or something like that.

data_search/analytical_data_search.md
```diff
@@ -0,0 +1,1144 @@
...
+
+## Introduction
+
+As you read through a recent paper by [Assani et al. 2025](https://arxiv.org/pdf/2504.02136), imagine that you find yourself consumed by a single thought:
+
+> I want to find more JWST spectral image cubes containing Fe II emission from jets launched by young stellar objects (YSOs).
+
+You may now be experiencing some despair. "This observation contains a JWST spectral cube" is certainly something readily available from MAST's metadata tables, but "this file contains Fe II transitions" is certainly not. So instead, you need to dig beyond the metadata and into the data of thousands of files. That may sound difficult, but it can be done!
+
+## Table of Contents
```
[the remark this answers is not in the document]
adrianlucy: Maybe in this case a good approach would be the suggested prose overview + hyperlinks to sections in that prose, e.g.: "First, we [blah blah blah (Section 1)]. Then, we [blah blah blah (Section 2)]." with hyperlinks on the text in brackets. Like the last paragraph of a paper's introduction, but including meaningful text in the hyperlinks for improved accessibility.

(Unfortunately links currently aren't stylized at all when you open a notebook in the Fornax platform, but hopefully that might change in the future...)

(FWIW in case you're interested: I'm told the reason MAST tends to recommend ToCs in our own notebook repository is 1. users running the notebook in-browser or on a platform probably won't find the tab to open the auto-ToC, and 2. concern about possible accessibility issues w.r.t. the auto-ToC.)

data_search/analytical_data_search.md
```diff
@@ -0,0 +1,1144 @@
...
+Note that we are not accounting for proper motions—and neither would `astroquery.mast` cone search, if we were using it. See [High PM Stars in MAST](https://outerspace.stsci.edu/display/MASTDOCS/High+PM+Stars+in+MAST) if this is a concern for your science case.
+
+```{code-cell} ipython3
+def check_points_in_polygon(row, coordinates_to_test):
+    """
+    Check if coordinates_to_test are in the s_region polygon for an observation.
+
+    Input parameters
+    ----------
+    row : a row of an astropy table of observations from astroquery
```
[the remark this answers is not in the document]
adrianlucy: I like `observation`! Will make this change here and elsewhere.

data_search/analytical_data_search.md
```diff
@@ -0,0 +1,1144 @@
...
+  display_name: python3
+  language: python
+  name: python3
+---
+
+# Beyond metadata: analytical data search in the cloud with JWST spectral cubes
+
++++
+
+For this notebook, we recommend using the [Fornax science platform](https://nasa-fornax.github.io/fornax-demo-notebooks/documentation/README.html) with Server Type = "Large - 64GB RAM/16 CPU" and Environment = "Default Astrophysics". As of June 2025, running this notebook once on that server will take about 10 minutes.
```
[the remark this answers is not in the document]
troyraen: You might want to consider using python's builtin `multiprocessing` for the parallelization instead of dask. After a quick look at your code, I'm guessing the calls with dask are using the same number of workers as there are CPU, so 4 on Medium server and 16 on Large. While the 4x speedup on Large is much better than the 1.5x on Medium, both are poor relative to the number of workers (and CPU, and probably memory as well) being used. If the parallelization were going well, I would expect a 16x speedup from 16 workers. As you guessed above, dask comes with a lot of overhead. The way the code is written also affects the efficiency of the parallelization but, while I didn't look at your code in enough detail to comment on it, I've had enough experience with dask that is similar to what you're reporting here that I suspect dask just isn't doing a very good job running it efficiently. Perhaps dask can be made to do better, but unfortunately I haven't seen much documentation instructing users about how to do that (though I haven't checked in quite awhile). Also, I'm hearing reports of significant issues with dask development right now with things in broken states and few developers available to fix it. So `multiprocessing` seems the better option all around. But that being said, I would certainly understand if you don't want to put the effort into switching your code at this point.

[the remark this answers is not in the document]
adrianlucy: @troyraen Interesting, thank you! Yeah, `dask` wasn't a deeply considered choice on my part, I was just following guidance that it's the strategic direction we wanted to focus on. But we shouldn't recommend it in cases where it's the wrong choice.

Indeed, I am seeing much better results with `multiprocessing` (with `num_workers = num_cores`). Relative to no explicit parallelization at all, now a 10x speed-up on Large (was 4x with `dask`), 3x speed-up on Medium (was 1.5x with `dask`). CPU usage is now ~100% (was ~70-80% with `dask`).

The choice between `dask` and `multiprocessing` may also have to do with the function being parallelized being very brief, in this case ~30--90 milliseconds per usage.

The speed up is still short of [number of cores]x. The reason may be that my code needs improvement, but my best guess is that it's at least partially because some python libraries are implicitly managing to make use of multiple cores even without explicit parallelization. On medium, CPU usage without explicit parallelization was 30--40%, so a 3x speed-up from parallelizing seems about right. On large, CPU usage without explicit parallelization was 7%, so we're a bit short of the 14x speed-up I might expect—but in this case 14x vs. 10x is a difference of 15 seconds so I could see start-up/wrap-up overhead being an issue. No improvement from increasing `chunksize`.

Anyway, I think I will indeed switch this to `multiprocessing`. And I guess I'll try to see if changing the parallelized function to apply to batches of rows instead of individual rows may yield better performance on either or both of `dask` and/or `multiprocessing`.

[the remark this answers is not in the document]
troyraen: Oh good, ya those are obviously better speedups. And I should have added a few more words to this sentence since the overhead may not be completely surmountable... "If the parallelization were going well, I would expect _closer to_ a 16x speedup from 16 workers." There are probably things you can tweak to get more efficiency, but also fine to leave as is. 

I should also mention, `multiprocessing` is not without issues. One in particular is that the behavior is a bit different on Linux, Mac, and Windows and it can be tricky to write code that works out of the box on all of them. In this repo we assume Linux, since that's what Fornax Console runs, and then include some text with info/instructions for Mac and Windows if we can. We've been hoping that dask would eventually be a better solution but so far haven't had enough success to fully adopt it.

> And I guess I'll try to see if changing the parallelized function to apply to batches of rows instead of individual rows

One simple way to do this is with the `chunksize` argument. The default is usually 1 which means that a new task is generated for every item in the input iterable, whereas a chunksize of (eg) 100 only generates a new task for every 100 items and so reduces the overhead. My rule of thumb is to generate at least 2-3 chunks per worker using something like `chunksize = min(100, len(my_iterable) // (3 * nworkers))`. Making chunks too big can result in some workers finishing a lot earlier than others and then just sitting around idle because all tasks have already been assigned out.

Another thing you can look into is the different functions that are available and what they're recommended for. For example, if the iterable is very long and the order in which results are returned is not important, `Pool.imap_unordered()` is recommended. But again, the 3x and 10x speedups are already reasonably good.

data_search/analytical_data_search.md
```diff
@@ -0,0 +1,1144 @@
...
+# Note: instead of using fsspec inside an astropy.fits.open() to handle the cloud connection,
+# you could equivalently use the s3fs module explicitly, as follows.
+# Results are typically similar.
+
+# import s3fs
+# fs = s3fs.S3FileSystem(anon=True)
+# with fs.open(cloud_uri, 'rb') as cloud_file:
+#     with fits.open(cloud_file, 'readonly') as hdul:
+#         header = hdul['SCI'].header
+#         data = hdul['SCI'].data
```
bsipocz, line 504: Please remove, no need for this commentary, we do not go one layer deep and explain all astropy functionality.

bsipocz (commented):
first round (dry-run -- I haven't run the notebook) reviews. The biggest issue is with the very right requirements. 

And the manual TOC will certainly be an issue as this further evolves.


Also, I wonder if we can try to add this to the rendering as part of this PR, or would you prefer to do that in a follow-up (e.g. TOC cleanup can be postponed to that one).

adrianlucy:
> Also, I wonder if we can try to add this to the rendering as part of this PR, or would you prefer to do that in a follow-up
I'd be happy to do it all in one PR. I'll work on incorporating the feedback thus far soon, almost certainly finishing sometime next week.

data_search/requirements_analytical_data_search.txt
```diff
@@ -0,0 +1,10 @@
+astropy>=7.0.1
+astroquery>=0.4.11.dev10199
+dask>=2025.4.1
+matplotlib>=3.10.3
+numpy>=2.3.0
```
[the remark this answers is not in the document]
adrianlucy: Probably the only tight requirement is on the astroquery version, I'll test out SPEC0 on everything else and loosen the requirements.

data_search/analytical_data_search.md
```diff
@@ -0,0 +1,1144 @@
...
+
++++
+
+At this time, MAST doesn't support reliable object classification search. So let's go to the experts in astronomical object cataloging, SIMBAD, which we can access through `astroquery.simbad.Simbad.query_tap`.
+
+In the cell below, we search for all SIMBAD-catalogued objects labeled as YSOs (`otype='Y*O'`) or as any of the descendant sub-concepts of YSOs, like T Tauri stars (`otype='Y*O..'` to retrieve both explicitly labeled YSOs and their subtypes).
+
+```{code-cell} ipython3
+# This will take a minute or two.
+yso_table = Simbad.query_tap("SELECT * FROM basic WHERE otype='Y*O..'", maxrec=1000000)
```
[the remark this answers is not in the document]
adrianlucy: This one I'd really, really like to push back on. The goal of the notebook is showing how to do large-scale data exploration, and SIMBAD object type search is an important workflow to demonstrate; there are a bunch of help desk tickets on our end on that subject that I could imagine pointing to this notebook. Notebook results will change anyway as new JWST observations are ingested, no way around that. MAST will help maintain the notebook.

data_search/README.md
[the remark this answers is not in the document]
jkrick: ok, then how about 'spectral_cubes' for the directory name since the other directories are named roughly by their data types?

data_search/analytical_data_search.md
```diff
@@ -0,0 +1,1144 @@
...
+axes[2].imshow(data[pos+10])  # +10 slices away for comparison
+axes[2].set_xlabel('y')
+axes[2].set_ylabel('x')
+axes[2].set_title('+10 wavebins away')
+
+plt.tight_layout()
+plt.show()
+```
+
+> Figure 4 image description: three astronomical images in pixel coordinates, side by side, display three different wavebin slices of a spectral cube. The left and right images are labeled "-10 wavebins away" and "+10 wavebins away", respectively, and feature a bright central source with faint hints of adjacent extention to the lower right. The middle image is labeled "5.34 micron line". In the middle image, the central source is much dimmer, emission extends all the way to the lower-right corner, and a blob in the lower-right corner is the brightest feature, in a region where the other two images exhibit no emission.
```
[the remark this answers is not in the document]
bsipocz: I know that there has been some recent work on it in the wider ecosystem, and also asked around.
What's certain atm, is that myst supports alt text for figures, but you may need to wizard around to have it added to the mpl generated figure. 

https://myst-parser.readthedocs.io/en/latest/syntax/images_and_figures.html

data_search/analytical_data_search.md
```diff
@@ -0,0 +1,1144 @@
...
+
+```{code-cell} ipython3
+# For example...
+print(jwst_obstable[0]['s_region'])
+```
+
+This `s_region` denotes the vertices of a polygon in right ascension and declination. We can make this string more parsable with this function:
+
+```{code-cell} ipython3
+def parse_polygon(s_region):
```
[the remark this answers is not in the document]
adrianlucy: > could this be bumped into code_src?

Sure.

> Also I wonder if all this polygon checkings (here and below) could be done with regions? If not, then I would recommend opening an issue.

On replacing `parse_polygon` with upstream functionality, it's possible I might have missed something in astropy for parsing s_region strings, I'll take another look. I agree that would be nice for `astropy/regions` to have, I can open an issue if I don't find anything. Note that they would probably want to cover a variety of cases we didn't need to consider here, e.g. multiple polygons in an s_region string, etc., so not quite trivial.

On replacing `check_points_in_polygon` with upstream functionality, [the relevant astropy/regions issue is 276](https://github.com/astropy/regions/issues/276) (and [this comment in 295](https://github.com/astropy/regions/issues/295#issuecomment-523597772) is also informative). I think we definitely need this function to create a WCS for the projection anyway, no way around that, but it's possible that the 4 lines involving `wcs_world2pix`, `Path`, and `Path.contains_points` could maybe be replaced with a `PolygonSkyRegion.contains` call. I'll see if that's right + test performance.

data_search/analytical_data_search.md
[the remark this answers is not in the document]
adrianlucy: Changed `row` to `obs` here and in some other functions.

data_search/analytical_data_search.md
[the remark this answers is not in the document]
adrianlucy: Runtimes for both medium and large server now included, removed advice to use large server.

data_search/requirements_analytical_data_search.txt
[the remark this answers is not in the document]
adrianlucy: Loosened the requirements to around spec0, except `astroquery` which needs version 0.4.11 (we added `return_uri_map` setting to `get_cloud_uris` for this notebook).

data_search/README.md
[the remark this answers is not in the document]
adrianlucy: Done

data_search/analytical_data_search.md
```diff
@@ -0,0 +1,1144 @@
...
+
+```{code-cell} ipython3
+Observations.enable_cloud_dataset()
+```
+
+Later, we'll be evaluating boolean conditionals on values that could be `None`, the pandas equivalent `pandas.NA`, or a normal value, and we'll also need this little function. It's pretty uninteresting, so let's get it out of the way now!
+
+```{code-cell} ipython3
+# Check that x is not None and is not pd.NA
+def exists(x):
```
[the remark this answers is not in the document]
adrianlucy: Done

data_search/analytical_data_search.md
```diff
@@ -0,0 +1,1144 @@
...
+```{code-cell} ipython3
+# This will take a minute or two.
+yso_table = Simbad.query_tap("SELECT * FROM basic WHERE otype='Y*O..'", maxrec=1000000)
+```
+
+```{code-cell} ipython3
+print(f'We found {len(yso_table)} YSOs.')
+```
+
+Let's make a list of their sky coordinates:
```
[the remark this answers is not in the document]
adrianlucy: Switched to the more elegant `yso_coords = np.transpose(np.asarray((yso_table_clean['ra'], yso_table_clean['dec'])))` (later on)

data_search/analytical_data_search.md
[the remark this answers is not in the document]
adrianlucy: (Also I did try `Pool.imap_unordered`, changing chunksize, etc., but didn't find anything that improved performance except as noted above.)

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+
+# As before...
+for obs in yso_jwst_obstable:
+    polygon = parse_polygon(obs['s_region'])
+    polygon_path = Path(polygon, closed=True)
+    # We'll use matplotlib.Path.contains_point (singular) this time,
+    # instead of .contains_points (plural)
+    is_inside = polygon_path.contains_point((tmc1a_ra, tmc1a_dec))
+    if is_inside:
+        tmc1a_jwst.append(obs)
```
bsipocz: This should be possible natively with astropy and regions; please use that rather than  brute force mpl.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+
+```{code-cell} ipython3
+# Convert back to a pandas dataframe
+results = pd.DataFrame(results)
+
+# Remove empty rows, where check_points_in_polygon reached the "else" part of the return statement.
+results = results.dropna(how='all')
+
+# Convert back to an astropy table.
+yso_jwst_obstable = Table.from_pandas(results)
```
bsipocz: Why convert back and forth between pandas and astropy? If you want to work with astropy Tables, then stick with those all the way.

spectral_cubes/code_src/detect_spikes.py
bsipocz: Please consolidate these functions into preferably one file, having one function per file is not exactly practical.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+* the advantages and constraints of a cloud science platform,
+* and when to load data from cloud storage instead of on-premise storage.
+
+## Introduction
+
+As you read through a recent paper by [Assani et al. 2025](https://arxiv.org/pdf/2504.02136), imagine that you find yourself consumed by a single thought:
+
+> I want to find more JWST spectral image cubes containing Fe II emission from jets launched by young stellar objects (YSOs).
+
+You may now be experiencing some despair. "This observation contains a JWST spectral cube" is certainly something readily available from MAST's metadata tables, but "this file contains Fe II transitions" is certainly not. So instead, you need to dig beyond the metadata and into the data of thousands of files. That may sound difficult, but it can be done!
```
bsipocz: Please make the language more factual and less story book like. Do not assume the user is in despair and don't know how to approach a science question, etc.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+## Imports
+
++++
+
+:::{tip}
+Temporary note: we'll need `boto3` to retrieve data from the AWS cloud with `astroquery`, and it's not currently installed in the default astrophysics Fornax kernel. Uncomment and run the following cell to install all dependencies, then restart your notebook's kernel.
+:::
+
+```{code-cell} ipython3
+# !pip install -r requirements_emission_search_cubes.txt
```
bsipocz:
```suggestion
# Uncomment the next line to install dependencies if needed.
# %pip install -r requirements_emission_search_cubes.txt
```

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+```
+
+**You need to restart your kernel for that to take effect.**
+
+```{code-cell} ipython3
```
bsipocz: Not a big deal, but this import section is the total opposite of the standard import orders, e.g. standard lib first, then core packages, then local stuff.

Also please tell AI to comment less, these are not very helpful -- standard lib is not utility, etc.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+
+## 1. Finding JWST spectral cubes of YSOs
+
++++
+
+Let's ignore the emission line issue and tackle the easiest parts of our driving question first: *I want to find more JWST spectral cubes of young stellar objects.*
+
++++
+
+### 1.1 Querying SIMBAD for YSOs
```
bsipocz: picking an input object list from simbad is not exactly lead to reproducible results. Would you consider basing this on an existing and citable YSO catalogue?

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+condition_yso = ~np.isnan(yso_table['ra']) & ~np.isnan(yso_table['dec'])
+yso_table_clean = yso_table[condition_yso]
+print(f'{len(yso_table_clean)} of these YSOs have good coordinates.')
+```
+
+### 1.2 Querying MAST for all JWST spectral cube observations
+
++++
+
+Now we want to get all the JWST spectral cube observations in MAST whose sky footprints overlap with any of these YSO coordinates. The natural tool is the `astroquery.mast.Observations` class, which gives programmatic access to MAST's multi-mission archive.
```
bsipocz: I would think this is a good place to link out to the astroquery MAST docs

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+
+However, `astroquery.mast` is not currently set up to quickly run a multi-target query for tens of thousands of sky coordinates. So in order to get things done in a reasonable amount of time, we need to think about the order of operations here.
+
+It turns out that, at the time of writing, there are only a few tens of thousands of JWST spectral cubes across the whole sky. So let's retrieve *all* of those first, excluding only those with a `calib_level` of `-1` (planned observations that haven't yet executed) and non-public datasets.
+
+```{code-cell} ipython3
+# This will typically take anywhere from a few seconds to a minute or so,
+# depending on how many people are using MAST right now.
+# Rarely, if MAST is overloaded, it may take several minutes or time out.
```
bsipocz: Pull this out into a note admonition. 

Overall pull out these type of comments from the code cells, we have narrative text to convey this info, no need to hide them in code.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+```{code-cell} ipython3
+yso_coords = np.transpose(np.asarray((yso_table_clean['ra'], yso_table_clean['dec'])))
+print(yso_coords.shape)
+```
+
+Performing the test on all observation sky footprints without parallelizing, as in the following cell, would take quite a while (e.g., six or seven minutes on the Fornax 16-CPU server). On Fornax, if you uncomment this cell and open the "Kernel usage" tab to your right while it runs, you'll notice that you're using only a small percentage of the total CPU capacity that you have access to (e.g., less than 10% if you selected the 16-CPU server option when you launched Fornax). Feel free to click the `Interrupt the kernel` stop button on your notebook, because there's a better way!
+
+```{code-cell} ipython3
+# Since this cell is an example of what not to do,
+# we'll also omit the nuance of gnomonic projection: see the next section.
```
bsipocz: pedagogically it's not the right thing to show what not to do, show what to do only and drop this code cell, it's perfectly adequate to narratively describe non-optimal ways, no need to show code for them.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+    # and used astropy PolygonSkyRegion.contains(coordinates_to_test, wcs),
+    # but the following technique yields >2x faster performance:
+
+    # Convert coordinates_to_test to the gnomonic projection
+    proj_coordinates_to_test = wcs.wcs_world2pix(coordinates_to_test, 0)
+
+    # Make footprint polygon matplotlib Path for this observation
+    polygon = parse_polygon(obs['s_region'])  # Parse s_region value
+    proj_polygon = wcs.wcs_world2pix(polygon, 0)  # Convert to gnomonic projection
+    proj_polygon_path = Path(proj_polygon, closed=True) # Convert to matplotlib Path
```
bsipocz: why not using regions instead?

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+
++++
+
+:::{tip}
+Temporary note: we'll need `boto3` to retrieve data from the AWS cloud with `astroquery`, and it's not currently installed in the default astrophysics Fornax kernel. Uncomment and run the following cell to install all dependencies, then restart your notebook's kernel.
+:::
+
+```{code-cell} ipython3
+# !pip install -r requirements_emission_search_cubes.txt
+```
```
troyraen: The notebook will get it's own environment in the Console and everything will be pre-installed. For those who aren't running this in the Console, we typically do this:

```suggestion
```{code-cell} ipython3
# %pip install -r requirements_emission_search_cubes.txt
```
```

File: spectral_cubes/emission_search_cubes.md [hunk 22]
```diff
@@ -0,0 +1,1101 @@
...
+
+```{code-cell} ipython3
+def check_points_in_polygon(obs, coordinates_to_test):
+    """
+    Check if coordinates_to_test are in the s_region polygon for an observation.
+
+    Parameters
+    ----------
+    obs : pandas.Series
+        A row of a pandas dataframe, converted from an astroquery-flavored astropy Table
```
troyraen: Include a description of the columns that are expected to be in here.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+
+```{code-cell} ipython3
+def line_search(obs, cloud_uri_map, transitions, plot=False):
+    """
+    Search a spectral cube observation for emission lines.
+
+    Parameters
+    ----------
+    obs : astropy.table.Row, pandas.Series
+        Row of an astropy or pandas table of observations
```
troyraen: Include a description of the columns that are expected to be in here.

spectral_cubes/code_src/parse_polygon.py
```diff
@@ -0,0 +1,35 @@
...
+def parse_polygon(s_region):
+    """
+    Extract coordinates from an s_region string.
+
+    Parameters
+    ----------
+    s_region : string - s_region from cross-mission archive
+               As written, this s_region must be a single polygon,
+               which is the case for all JWST and SOFIA observations,
+               but not all missions in the Fornax archives.
```
troyraen: Maybe give an example of what this string should look like.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+    (tmc1a_jwst_obstable['instrument_name']=='MIRI/IFU') &
+    (tmc1a_jwst_obstable['filters']=='CH1-SHORT') &
+    (tmc1a_jwst_obstable['proposal_id']=='1290')
+]
+
+# Get the first valid cloud URI found
+for obs in temp_observations:
+    cloud_uri = cloud_uri_map[obs['dataURL']]
+    if cloud_uri:
+        break  # When a valid URI is found, exit the loop
```
bsipocz: I don't have an alternative suggestion, but this loop feels a bit hacky

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+```{code-cell} ipython3
+with fits.open(cloud_uri, cache=False, use_fsspec=True, fsspec_kwargs={"anon": True}) as hdul:
+    header = hdul['SCI'].header
+    data = hdul['SCI'].data
+```
+
+Let's turn this into a convenience function that we can use later on:
+
+```{code-cell} ipython3
+def load_cloud(cloud_uri, extension='SCI'):
```
bsipocz: please rename this to something more descriptive, `load_cloud_data` or `load_cloud_fits_image`

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+```
+
+```{code-cell} ipython3
+plt.plot(test_blobs, color='k')
+plt.xlabel('Wavelength bin index')
+plt.ylabel('Fraction of non-NaN pixels above threshold')
+plt.show()
+```
+
+> Figure 1 alt text: a line plot of "Fraction of non-NaN pixels above threshold" versus "Wavelength bin index", the latter ranging from around 0 to 1000. The plotted values have substantial scatter and features, but one feature stands out by a factor of two above all others: a narrow, abrupt spike near wavelength bin index 548.
```
bsipocz: It's great to have alt-text for the figure, but I have serious concerns with its accuracy over time. Also, for actual alt-text I strongly suggest to use some mystmd native approach for embedding the output into a figure directive. 
Or, as this code cell is nothing but the plotting, use the `code-cell` caption?

What to do with this is more of a question to @troyraen -- I don't think we can guarantee that any of these type of alt-text would stay accurate, unless we somehow auto-generate it from the data/with matplotlib.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+print(detect_spikes(test_blobs, 548, 3.0))
+```
+
+And False if we give it an unremarkable wavebin index:
+
+```{code-cell} ipython3
+print(detect_spikes(test_blobs, 540, 3.0))
+```
+
+Our third function is simple, and retrieves the wavelength array from a JWST spectral cube file's FITS header:
```
bsipocz: I very much like that the narrative and code are nicely cut up here, thanks!

troyraen:
> Also, please note that the review was done on the code without running the notebook, execution of it is currently stuck on L332 with some traceback coming from the cell, but it's still running.

I believe that `multiprocessing` on OSX or Windows requires that the functions be in .py files and imported, so I would recommend moving them out of the notebook. (When functions are defined directly in the notebook, they only exist in-memory in the main process and do not get copied into child processes.)

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+- MAST internal review: Thomas Dutkiewicz, Sam Bianco, Zach Claytor, Brian McLean, Jonathan Hargis
+- Fornax review: Troy Raen, Jessica Krick, Brigitta Sipőcz
+
+### References
+
+This notebook relies on the following papers:
+- Assani et al. 2025 ([2025arXiv250402136A](https://ui.adsabs.harvard.edu/abs/2025arXiv250402136A/abstract))
+- Karska et al. 2025 ([2025A&A...697A.186K](https://ui.adsabs.harvard.edu/abs/2025A%26A...697A.186K/abstract))
+
+And the following software:
```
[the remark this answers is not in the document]
adrianlucy: Done for numpy and pandas

spectral_cubes/emission_search_cubes.md
adrianlucy: I looked at this a while back. Unfortunately, looping through an equivalent iterable of an astropy table (the output/input of astroquery) drastically hurts multiprocessing performance relative to iterating through a pandas dataframe, in this case. But I have cleaned up the code further up to define an iterable straight away rather than bothering to define a pandas table.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+    format_name: myst
+    format_version: 0.13
+    jupytext_version: 1.18.1
+kernelspec:
+  display_name: python3
+  language: python
+  name: python3
+---
+
+# Beyond metadata: analytical data search in the cloud with JWST spectral cubes
```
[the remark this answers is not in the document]
adrianlucy: Title revised to something more concrete!

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+
+As you read through a recent paper by [Assani et al. 2025](https://arxiv.org/pdf/2504.02136), imagine that you find yourself consumed by a single thought:
+
+> I want to find more JWST spectral image cubes containing Fe II emission from jets launched by young stellar objects (YSOs).
+
+You may now be experiencing some despair. "This observation contains a JWST spectral cube" is certainly something readily available from MAST's metadata tables, but "this file contains Fe II transitions" is certainly not. So instead, you need to dig beyond the metadata and into the data of thousands of files. That may sound difficult, but it can be done!
+
+### Runtime
+
+We assume that you are running this notebook on the [Fornax science platform](https://nasa-fornax.github.io/fornax-demo-notebooks/documentation/README.html) in `Environment` = `Default Astrophysics`, but it will also still work (in order of preference) in another AWS-based cloud science platform, a non-AWS cloud science platform, and even on your local computer — perhaps at slower speeds, depending on your computing resources and internet connection.
```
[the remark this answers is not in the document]
adrianlucy: Done

spectral_cubes/emission_search_cubes.md
adrianlucy: Please see existing thread https://github.com/nasa-fornax/fornax-demo-notebooks/pull/443#discussion_r2226359426

spectral_cubes/emission_search_cubes.md
adrianlucy: Done, here and elsewhere

spectral_cubes/emission_search_cubes.md
adrianlucy: Please see https://github.com/nasa-fornax/fornax-demo-notebooks/pull/443#discussion_r2583081790

spectral_cubes/emission_search_cubes.md
adrianlucy: This notebook was not written with AI, @bsipocz. But sure, I've dropped the "utility" comment and put the imports-formerly-known-as-utilities at the top.

spectral_cubes/emission_search_cubes.md
```diff
@@ -0,0 +1,1101 @@
...
+            plt.ylabel('Fraction of non-NaN pixels above threshold')
+            plt.show()
+
+    # Return the input row with a new column holding the detected_feii_lines list
+    obs['detected_feii_lines'] = detected_feii_lines
+
+    return obs
+```
+
+Phew! Let's try this out on our TMC1A observations.
```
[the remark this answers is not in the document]
adrianlucy: Sure, revised. FWIW it might be useful to write a style guide that explains the tone you are aiming for more concretely. While the content you want removed is clear in this case, "professional" is a moving target; the MAST Notebook Repository, for our part, generally does not have the same position on conversational tone.

spectral_cubes/emission_search_cubes.md
adrianlucy: I think the problem of tracking/flagging temporarily-missing cloud URIs is something that ought to be improved on the MAST services end, I can ask some people.

spectral_cubes/emission_search_cubes.md
adrianlucy: Oo, very interesting, thank you! I'll try that today.

adrianlucy: @bsipocz, I feel foolish but I can't figure out how this helps include alt-text or a caption in a notebook. If you think it can do that, could you perhaps show the change to the markdown that you think would work? I can't get it to work. e.g. for something easy to test I tried this in the markdown:
```
```{code-cell} ipython3
:caption: TESTING CAPTIONS

plt.plot(test_blobs, color='k')
plt.xlabel('Wavelength bin index')
plt.ylabel('Fraction of non-NaN pixels above threshold')
plt.show()
```
"TESTING CAPTIONS" does not appear in the executed notebook, and in my reading nothing in the `mystmd` documentation suggests it would. But maybe I'm doing something wrong.

Perhaps you mean that the alt-text or caption would manifest in the render. But isn't it better for alt-text to be available when interacting with the notebook itself in the Fornax platform, even if imperfectly?

Thank you!

troyraen:
Thanks for your patience and persistence @adrianlucy. This is a lot of work! I haven't tracked all the recent conversations yet, but after a quick look, I see that you've certainly been responsive to feedback. And also have two approving reviews. I'm headed out for a long weekend but will follow up next week.

%%writefile magic hadn't been on my radar - looking forward to seeing if it's a solution for problems with a couple notebooks I wrote for IRSA!

adrianlucy:
@troyraen That all sounds good to me—thank you!

Don't hesitate to let me know if you need anything from me during this process. I'm out Wedn and Thurs this week, but back Friday. (PS, if you happen to have an approximate guess on when it will be added to the website rendering, there are folks asking me for updates. No worries if not!)

Many thanks for the work that you all have and will put into this, and for your extremely useful feedback throughout this review.

PR #444, opened 2025-07-11
SDSS bugfix

### LINKED ISSUE #437: Bug: SDSS function in spectra_generator.md
WARNING: Field info are not available for this data release [astroquery.sdss.field_names]
WARNING: Field info are not available for this data release [astroquery.sdss.field_names]
KeyError                                  Traceback (most recent call last)
File [/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astropy/table/row.py:54](https://science-console.fornax.smce.nasa.gov/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astropy/table/row.py#line=53), in Row.__getitem__(self, item)
     50 try:
     51     # Try the most common use case of accessing a single column in the Row.
     52     # Bypass the TableColumns __getitem__ since that does more testing
     53     # and allows a list of tuple or str, which is not the right thing here.
---> 54     out = OrderedDict.__getitem__(self._table.columns, item)[self._index]
     55 except (KeyError, TypeError):

KeyError: 'run2d'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
File <timed exec>:2

File [~/fornax-demo-notebooks/spectroscopy/code_src/sdss_functions.py:44](https://science-console.fornax.smce.nasa.gov/jupyter/user/jkrick/fornax-demo-notebooks/spectroscopy/code_src/sdss_functions.py#line=43), in SDSS_get_spec(sample_table, search_radius_arcsec, data_release)
     41     print("Source {} could not be found".format(stab["label"]))
     42     continue
---> 44 sp = SDSS.get_spectra(matches=xid, show_progress=True, data_release=data_release)
     46 # Get data
     47 # only one entry because we only search for one xid at a time. Could change that?
     48 wave = 10**sp[0]["COADD"].data.loglam * u.angstrom

File [/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astroquery/sdss/core.py:798](https://science-console.fornax.smce.nasa.gov/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astroquery/sdss/core.py#line=797), in SDSSClass.get_spectra(self, coordinates, radius, matches, plate, fiberID, mjd, timeout, get_query_payload, data_release, cache, show_progress)
    785 @prepend_docstr_nosections(get_spectra_async.__doc__)
    786 def get_spectra(self, *, coordinates=None, radius=2. * u.arcsec,
    787                 matches=None, plate=None, fiberID=None, mjd=None,
    788                 timeout=TIMEOUT, get_query_payload=False,
    789                 data_release=conf.default_release, cache=True,
    790                 show_progress=True):
    791     """
    792     Returns
    793     -------
    794     list : List of `~astropy.io.fits.HDUList` objects.
    795 
    796     """
--> 798     readable_objs = self.get_spectra_async(coordinates=coordinates,
    799                                            radius=radius, matches=matches,
    800                                            plate=plate, fiberID=fiberID,
    801                                            mjd=mjd, timeout=timeout,
    802                                            get_query_payload=get_query_payload,
    803                                            data_release=data_release,
    804                                            cache=cache,
    805                                            show_progress=show_progress)
    807     if get_query_payload:
    808         return readable_objs

File [/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astroquery/sdss/core.py:738](https://science-console.fornax.smce.nasa.gov/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astroquery/sdss/core.py#line=737), in SDSSClass.get_spectra_async(self, coordinates, radius, matches, plate, fiberID, mjd, timeout, get_query_payload, data_release, cache, show_progress)
    734 linkstr = self.SPECTRA_URL_SUFFIX
    735 # _parse_result returns bytes (requiring a decode) for
    736 # - instruments
    737 # - run2d sometimes (#739)
--> 738 if isinstance(row['run2d'], bytes):
    739     run2d = row['run2d'].decode()
    740 elif isinstance(row['run2d'], (np.integer, int)):

File [/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astropy/table/row.py:64](https://science-console.fornax.smce.nasa.gov/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astropy/table/row.py#line=63), in Row.__getitem__(self, item)
     61         out = tuple(self.values())[item]
     62     else:
     63         # This is only to raise an exception
---> 64         out = self._table.columns[item][self._index]
     65 return out

File [/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astropy/table/table.py:264](https://science-console.fornax.smce.nasa.gov/opt/envs/py-spectra_generator/lib/python3.12/site-packages/astropy/table/table.py#line=263), in TableColumns.__getitem__(self, item)
    253 """Get items from a TableColumns object.
    254 
    255 ::
   (...)    261   tc[1:3] # <TableColumns names=('b', 'c')>
    262 """
    263 if isinstance(item, str):
--> 264     return OrderedDict.__getitem__(self, item)
    265 elif isinstance(item, (int, np.integer)):
    266     return list(self.values())[item]

KeyError: 'run2d'

jkrick:
Short PR to fix the SDSS bug in #437 
This changes the checks for what is returned from SDSS.query_region to capture the noted behavior.  
The SDSS module now works as expected and adds minimal time to the overall notebook so I have left the runtime estimate as is.

This will close #437

### DIFF
--- a/spectroscopy/code_src/sdss_functions.py
+++ b/spectroscopy/code_src/sdss_functions.py
@@ -2,6 +2,7 @@ import astropy.units as u
 import numpy as np
 import pandas as pd
 from astroquery.sdss import SDSS
+from astropy.table import Table
 
 from data_structures_spec import MultiIndexDFObject
 
@@ -37,8 +38,15 @@ def SDSS_get_spec(sample_table, search_radius_arcsec, data_release):
         xid = SDSS.query_region(search_coords, radius=search_radius_arcsec *
                                 u.arcsec, spectro=True, data_release=data_release)
 
-        if str(type(xid)) == "<class 'NoneType'>":
-            print("Source {} could not be found".format(stab["label"]))
+        #empty result means no match
+        if not isinstance(xid, Table) or len(xid) == 0:
+            print(f"Source {stab['label']} returned no results.")
+            continue
+            
+        #sometimes the query returns an unexpected result of an astropy table with no real columns
+        # and HTTP errors, make sure to catch this by checking if the colnames that we require exist.
+        if {"plate", "mjd", "fiberID"}.difference(xid.colnames):
+            print(f"Source {stab['label']} missing required SDSS columns: {xid.colnames}")
             continue
 
         sp = SDSS.get_spectra(matches=xid, show_progress=True, data_release=data_release)


bsipocz:
Yeap, it looks like it's on top of the unrelated changes. If you want to practice you can interactive rebase and comment out those 2-3 commits, or just let me know and I'll do it myself.

jkrick:
@bsipocz can you please do the rebase?  I looked at it, but it doesn't make sense to me.  There are commits in the log from the command line that don't show in the above website PR list.  I'm afraid I will do something unintentionally bad because I don't understand it.

bsipocz:
SDSS is usually pretty reliable, so I would really not worry too much about this one coming back.

jkrick:
That may be true, but it happened at least for a few days last week, and past performance is not a predictor of the the future.  Also, since its just a simple change of if statement, its not a large change, I think it would be better to go ahead with the PR, then the code is ready if it happens again.

troyraen:
I'm going to rebase this and then merge it.

bsipocz:
> That may be true, but it happened at least for a few days last week, and past performance is not a predictor of the the future. Also, since its just a simple change of if statement, its not a large change, I think it would be better to go ahead with the PR, then the code is ready if it happens again.

I still think that we should not swallow such issues in the user facing code. If there were errors it should not be silenced otherwise the results are non reproducible. --> returning null results is only acceptable when the DB returns no results due to no coverage or similar, not when there are errors along the way.

bsipocz:
Cases we had at https://github.com/nasa-fornax/fornax-demo-notebooks/issues/437 should not result a passing notebook. It should definitely be a user friendly/useful error message, but IMO should be an exception rather than silenced, as in practice anything printed will be taken in two ways: completely ignored or considered noise.

bsipocz:
Yes, I think we should keep raising errors when those are unexpected/something that has to be worked around all the time. E.g. in this case a temporary (and otherwise unusual) server side glitch.

troyraen:
I agree with @bsipocz about error handling. I didn't look at this closely enough before merging. I'll open a new PR to fix it.

jkrick:
Ok, I don't think I am understanding what is going on with the SDSS bugfix PR for light curve generator.  in #437  the code was crashing because SDSS returned junk.  This was problematic for two reasons, 1) the error message was unintelligable, as most are, and 2) it crashed the code.  If a source doesn't get light curve info out of a specific archive, I don't want the whole thing to crash, I want it to keep going on to the next source, the next archive, etc.  

So, my PR #444  did solve, I think, both of those problems.  It printed out something useful, and it kept the code continuing on through the flow.

I thought that Brigitta's comment above was to not have it print a statement and continue, but instead to raise an exception, and then catch the exception, so the code can move on.  This is probably where I made an assumption that might not be true.  Catching exceptions isn't currently in my repertoire, but I thought that was being requested as a more serious way of letting the user know something strange had happened by printing a "warning" instead of a lowly print statement that no one would read.

I'm mostly trying to understand what the best approach is going forward because we think this won't happen again with SDSS, but I feel like this is a common enough occurrence (an archive hiccups), that I should know best practice for "catching" that.  

Thank you for your patience on this one @bsipocz @troyraen

troyraen:
Thanks for following up @jkrick. I definitely agree that this is an important thing that comes up fairly often and it would be really helpful to converge on a clear strategy for error handling in our tutorials repos. I'm sure I've contributed plenty of miscommunication about it, so I'll try to clear that up. When I first got here and made such a mess of our error handling I ended up saying that I didn't care what we did. That was because I feel that the "right" answer depends strongly on the context and goals and I didn't know what ours were yet. At this point, I've had more exposure to what we're doing and thought through it more and now I have a pretty strong opinion. I'm not sure how I'm coming across in writing here, so I'll just say upfront that I do want to keep talking through it until we converge and I don't mean to be lecture-y or overwhelming. Just because I have a strong opinion doesn't mean that it's either right or communicated clearly.

In a nutshell, I WANT these kinds of errors to crash our notebooks, though not indefinitely. I do think that if we had the same goals as typical scientists then we might decide that we'd really prefer that the notebook keep running rather than to crash if there's anyway we can prevent that. But in my view, we need to play a very different role here, and in order to do that, we have to know that the error occurred.

For us, the _error_ is the problem that we need to fix (or help fix). I see this as an important part of our jobs working for archives. If we let the error persist and it prevents our end users from getting data, they'll miss out on the data regardless of whether their code full on crashes or just quietly passes over the error. Even if the error doesn't prevent data from getting to them, we still need to help remove the hurdles for everyone using these services rather than just hiding them in this one place (our notebooks). Preemptively hiding errors would ultimately be a disservice to end-user scientists.

The _crash_ is the single most useful symptom that alerts us to the problem and we do ourselves a big disservice by preventing it prematurely. The CI will never (in my understanding) notice a problem if the notebook doesn't crash, and humans will only do slightly better. Even though these particular errors were transitory and didn't require a fix this time, if they keep happening then they really will need a fix. If we alter the code to prevent the notebook from crashing if/when an error does come back then it's highly unlikely that we'll even notice the error at all for quite some time. When errors/crashes do occur, I agree that we don't want to just let the notebook keep crashing for a long time while we try to fix the underlying problem. But we should decide what do then, on a case-by-case basis, not now.

> This was problematic for two reasons, 1) the error message was unintelligable, as most are, and 2) it crashed the code.

I think I covered my opinion about the crash. For the error message, especially for transitory errors like these that we don't expect to come back, I don't think we should spend any time trying to help the end user understand the message. If/when an error does come back, assuming we haven't prevented the crashing then we have a good chance of catching it before any of our end users. We can look through our repo issues for the history and we have developers who have experience reading tracebacks. If the fix isn't quick, we'll decide how to handle the crashing then. If an end user sees it first, they can report it to us but they can't do anything to fix it themselves, so a preemptive explanation isn't super helpful. On the contrary, explaining problems that _might_ happen again but in reality never do gives the user the wrong impression. Also, putting it in code can leave us with a lot of unnecessary code to maintain that we don't even realize is unnecessary. 

If we think an error like this has a good chance of coming back (but isn't happening now) and we really want to warn the user, I think my preference is to add a comment in the notebook itself right above where the error will appear, similar to what we do for the dask and lightkurve "errors" (in quotes because those are fundamentally different than the errors we're talking about here). Another option is to catch the error, append our own error message that we think is more clear, and re-raise the same error so that the full traceback goes with it, but I really don't think we should be messing with that in these notebooks. In any case, what I definitely don't want to do is to catch (or bypass) the error and just drop it, even if we raise something else. The full tracebacks contain a lot of information that is important and helpful for the people who need to track down the error and fix it. Preemptively dropping means that not only do we have to do work to drop it, we have to go back and undo that work to find it again.

> I thought that Brigitta's comment above was to not have it print a statement and continue, but instead to raise an exception, and then catch the exception, so the code can move on.

I'm not quite following that. Either way, the result is the same -- no crash. If that's the point, there's no need for us to raise an error (= exception). We can just catch the original one that causes the crash, print something (or not), and move on. Raising a warning (!= exception) still just results in a printed statement. We might want that in some cases but we shouldn't expect our users to really notice that much more than a regular print statement. I can believe that we'll run into cases (and probably have) where we do want to catch an error and just move on. But I definitely don't think we should do it for this "just in case" kind of situation. It should be for errors that are actually occurring in the moment and we should talk/think thru other options as well. 

And actually, this reminds me that I added a custom and extremely broad exception to the forced_photometry notebook to catch a tractor error. That was a mistake. I was trying to put in a patch that would be both temporary and satisfy conflicting opinions about it, but it ended up being neither. That might be a case where we really do want to catch a specific error and just move on but, if so, we should do it differently than I set it up. I'll look into that.

PR #445, opened 2025-07-11
changed to specutils.Spectrum instead of Spectrum1D

### LINKED ISSUE #436: AstropyDeprecationWarning for spectra_generator.md
In the HST cell:
WARNING: AstropyDeprecationWarning: The Spectrum1D class is deprecated and may be removed in a future version.
        Use Spectrum instead. [astropy.io.registry.core]

jkrick:
fixing #436 deprecation warning by switching from Spectrum to Spectrum1D.  Appears to have no downstream effects and doesn't effect the runtime. 

This will close #436

### DIFF
--- a/spectroscopy/code_src/mast_functions.py
+++ b/spectroscopy/code_src/mast_functions.py
@@ -13,7 +13,7 @@ import pandas as pd
 from astropy.table import Table
 from astropy.table import vstack
 from astroquery.mast import Observations
-from specutils import Spectrum1D
+from specutils import Spectrum
 
 from data_structures_spec import MultiIndexDFObject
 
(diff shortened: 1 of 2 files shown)


spectroscopy/code_src/mast_functions.py
```diff
@@ -13,7 +13,7 @@
 from astropy.table import Table
 from astropy.table import vstack
 from astroquery.mast import Observations
-from specutils import Spectrum1D
+from specutils import Spectrum
```
bsipocz, line 16: This bumps the minimum spectutils requirement to be `specutils>=1.20` which got released 2025 May, so we may consider doing a conditional import (IMO it's fine in this specific case to require this almost latest version as specutils doesn't limit its dependencies that much, but overall this approach may not work with other packages).

bsipocz (approved):
Thanks. I've added the version restriction, so this should be good to go.

PR #453, opened 2025-07-28
Move ZTF to non-NASA section

troyraen:
Fixes #435

Move the ZTF call to the non-NASA section.

Issue #383, opened 2025-03-05
MAINT: cleanup direct upath usage once new HATS released


bsipocz:
See the discussion in this thread: https://github.com/nasa-fornax/fornax-demo-notebooks/pull/382#discussion_r1981948998

We can cleanup the notebook once the upstream https://github.com/astronomy-commons/hats/pull/466 is in a release.

PR #455, opened 2025-07-28
Switch ZTF call to use ODR dataset

### LINKED ISSUE #367: Update Light curve notebook to use HATS ZTF dataset from AWS ODR bucket
I think this is maybe part of #298, but this issue is specific to ZTF.
Once the ZTF catalog data is in HATS format and once it is in the ODR, then we need to update the notebook functions to be able to use that dataset.

troyraen:
Merge AFTER #453 

Closes #367 
Partial fix for #383 

Included changes:
- Switch ZTF call to use data in the public ODR bucket.
- Remove light_curve_generator.md from conf.py `nb_execution_excludepatterns` so that it will be tested again.

I tried to run the entire notebook to be able to update the runtime but I ran into a `DALQueryError` in `heasarc_get_lightcurves()`.


conf.py
```diff
@@ -56,12 +56,9 @@
 # Don't execute the forced photometry notebook until we base the CI on
 # the actual fornax image instead of the fresh installs
 # (aka tractor install pain).
+# Also blocked by https://github.com/nasa-fornax/fornax-demo-notebooks/issues/311 (Spitzer COSMOS).
 nb_execution_excludepatterns += ['multiband_photometry.md',]
 
-# We use the non-public IRSA bucket for ZTF data, cannot execute the generator
-# notebook until https://github.com/nasa-fornax/fornax-demo-notebooks/issues/311 is addressed
-nb_execution_excludepatterns += ['light_curve_generator.md',]
-
```
troyraen, line 64: @bsipocz is this the right thing to do here?

bsipocz: yes, this looks good to me, once we move to the ODR bucket the notebook is not blocked any more with the bucket issue. (I may have other, unrelated issues though, so whether you can do this in the same PR or not will depend)

light_curves/light_curve_generator.md
```diff
@@ -302,7 +300,7 @@ print('HCV search took:', time.time() - HCVstarttime, 's')
 ### 3.1 IRSA: ZTF
 The function to retrieve ZTF light curves accesses a [HATS](https://hats.readthedocs.io/en/stable/) parquet version of the ZTF catalog stored in the cloud using [LSDB](https://docs.lsdb.io/en/stable/). This is the simplest way to access this dataset at scale.  The ZTF [API](https://irsa.ipac.caltech.edu/docs/program_interface/ztf_lightcurve_api.html) is available for small sample searches.  One unique thing about this function is that it has parallelization built in to the function itself because lsdb uses dask under the hood.
 
-Traceback CommClosedErrors are expected and are just a dask houskeeping issue, the function is still running to completion and returning light curves.
+Traceback CommClosedErrors are expected and are just a dask housekeeping issue, the function is still running to completion and returning light curves.
```
bsipocz, line 303: added https://github.com/nasa-fornax/fornax-demo-notebooks/issues/456

jkrick (approved):
This code works for me.  Thanks for setting up this PR.  2 super short comments:

1. Dask is quite verbose.  Is there some way to quiet that?  or maybe at least mention in the text of light_curve_generator.md that all of these messages which start with "INFO" are expected.  It spits out around 50 lines of information that looks like a warning, and some of it looks like an error, but apparently isn't.  We should explicitly mention that one:
```
INFO:distributed.batched:Batched Comm Closed <TCP (closed) Scheduler connection to worker local=tcp://127.0.0.1:35617 remote=tcp://127.0.0.1:51736>
Traceback (most recent call last):
  File "[/opt/envs/py-light_curve_generator/lib/python3.12/site-packages/distributed/batched.py", line 115](https://science-console.fornax.smce.nasa.gov/opt/envs/py-light_curve_generator/lib/python3.12/site-packages/distributed/batched.py#line=114), in _background_send
    nbytes = yield coro
             ^^^^^^^^^^
  File "[/opt/envs/py-light_curve_generator/lib/python3.12/site-packages/tornado/gen.py", line 769](https://science-console.fornax.smce.nasa.gov/opt/envs/py-light_curve_generator/lib/python3.12/site-packages/tornado/gen.py#line=768), in run
    value = future.result()
            ^^^^^^^^^^^^^^^
  File "[/opt/envs/py-light_curve_generator/lib/python3.12/site-packages/distributed/comm/tcp.py", line 263](https://science-console.fornax.smce.nasa.gov/opt/envs/py-light_curve_generator/lib/python3.12/site-packages/distributed/comm/tcp.py#line=262), in write
    raise CommClosedError()
distributed.comm.core.CommClosedError"

```
Chatgpt tells me this is dask closing down and sending messages after it has already started closing down, and is not a real error.

2. One of the dask comments is that it would like to have jupyter-server-proxy installed.  I pip installed this, but am not sure what I am gaining from it.  Which is really a comment that I don't use any of the dask "watch your cpus work" features.  I think you have some issue open about teaching people how to use dask on Fornax, and one day I will follow whatever instructions get figured out.  In the meantime, I wonder if we should add a comment to this cell which says, go ahead and pip install if you want to "blah".  I just need help to know what "blah" is.  Alternatively, this falls under my above comment  about including some verbiage around expecting all the messages that come out.

jkrick:
well, I am at least consistent.  Sorry for not seeing the comment that already existed.  

I agree, fix bigger dask problem, then worry about smaller issue in this notebook.

conf.py
troyraen: Latest testing still failed with a dead kernel. Need to investigate why that is happening.

I suspect this is unrelated, but we probably need to exclude this from Windows testing due to the use of `multiprocessing`. I added a block to conf.py for that. I think this notebook should still work on Mac but need to check.

bsipocz:
Maybe pinning the dask+extension versions is a good workaround for now? Right now it's not clear to me if we should spend any time figuring out dask, I would say sit out until the situation either resolves or their pathway is cleaned up.

troyraen:
I can't find any useful info on CircleCI about why the dead kernel is happening. My first guess was that it's running out of memory but I tried to run the notebook on Fornax and it stops before it even gets to the memory-intensive calls because it runs into an error (#457). I think (but not sure) that CircleCI would hit that error, stop, and just move on to the next notebook, so I don't think that's the cause of the dead kernel either.

Let's work out the issues with the CircleCI run in a separate PR. For this one, I will remove the related commits and merge.

Issue #457, opened 2025-07-30
BUG: HEASARC call throws `DALQueryError`


troyraen:
The `heasarc_get_lightcurves()` function in light_curve_generator.md is throwing `DALQueryError`. The pyvo version is 1.7.

Full traceback:

```
---------------------------------------------------------------------------
DALQueryError                             Traceback (most recent call last)
Cell In[8], line 13
     10 heasarc_catalogs = {"FERMIGTRIG": max_fermi_error_radius, "SAXGRBMGRB": max_sax_error_radius}
     12 # get heasarc light curves in the above curated list of catalogs
---> 13 df_lc_HEASARC = heasarc_get_lightcurves(sample_table, catalog_error_radii=heasarc_catalogs)
     15 # add the resulting dataframe to all other archives
     16 df_lc.append(df_lc_HEASARC)

File [~/fornax/fornax-demo-notebooks/light_curves/code_src/heasarc_functions.py:114](https://science-console.fornax.smce.nasa.gov/jupyter/user/raen/lab/tree/fornax/fornax-demo-notebooks/light_curves/fornax/fornax-demo-notebooks/light_curves/code_src/heasarc_functions.py#line=113), in heasarc_get_lightcurves(sample_table, catalog_error_radii)
    110 upload_tables = [upload_table[idd] for idd in ids]
    112 # hresult: is a list of query results corresponding to upload_tables.
    113 # hresulttable: is the stacked table of all the results from individual calls.
--> 114 hresult = [heasarc_tap.service.run_sync(hquery, uploads={'mytable': upload_table})
    115            for upload_table in upload_tables]
    116 hresulttable = vstack([hr.to_table() for hr in hresult])
    118 # add results to multiindex_df
    119 # really just need to mark this spot with a vertical line in the plot, it's not actually a light curve
    120 # so making up a flux and an error, but the time stamp and mission are the real variables we want to keep

File [/opt/envs/py-light_curve_generator/lib/python3.12/site-packages/pyvo/dal/tap.py:282](https://science-console.fornax.smce.nasa.gov/opt/envs/py-light_curve_generator/lib/python3.12/site-packages/pyvo/dal/tap.py#line=281), in TAPService.run_sync(self, query, language, maxrec, uploads, **keywords)
    253 def run_sync(
    254         self, query, *, language="ADQL", maxrec=None, uploads=None,
    255         **keywords):
    256     """
    257     runs sync query and returns its result
    258 
[...]

PR #460, opened 2025-07-30
Rename generators to collectors

### LINKED ISSUE #441: Rename spectra_generator.md
https://jira.ipac.caltech.edu/browse/IRSA-7089

Name of the notebook is misleading, no spectra is being generated here. It can be called something like:
- SpectraCollector 
- SpecQuery
- CrossArchiveSpecQuery
- ...

troyraen:
Closes #441 

- Rename spectra_generator -> spectra_collector
- Rename light_curve_generator -> light_curve_collector
- Bugfix: Rename Euclid_functions.py to match the notebook import statement


bsipocz (changes requested):
You will need to the same renames for any references in the toctrees in the markdown files as well as if any of these are included in the config and ignore files.

troyraen:
I reverted the kernel names in the yaml headers so that the rest of these changes don't need to wait for coordination with https://github.com/nasa-fornax/fornax-images/issues/49. 

I tested both "collector" notebooks on fornax. The kernels are picked up automatically and the requirements*.txt files work, so I don't see any execution problems w.r.t. this PR. I did run into #457 as well as three timeout errors (1 in lightcurves, 2 in spectra) when calling MAST services. I'll check the MAST calls again in the morning and open issues if the problems persist.

The circleci failure is a dead kernel while executing light_curve_classifier, so I assume that's #475.

@bsipocz do you see any more work that needs to be done here or is this ready to merge?

troyraen:
The three calls to MAST services are all working again.

PR #472, opened 2025-08-15
Move notebook review

jkrick:
This was notebook_review_process.md in the fornax-documentation repo, but this might be a better place to put this set of checklists for code review in this fornax-demo-notebooks repo.

### DIFF
--- a/notebook_review_checklists.md
+++ b/notebook_review_checklists.md
@@ -1,0 +1,54 @@
+# How Does Fornax Review Notebooks?
+***
+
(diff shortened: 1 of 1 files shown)


bsipocz:
I would say maybe we want to put this into a docs directory, but also, we can just leave where it is right now and move if/when it gets chopped up or if/when we add more docs.

troyraen:
I think we should put it in the same directory as the template once that gets added.

PR #481, opened 2025-08-21
Comment out heasarc_get_lightcurves() to temporarily patch #457

troyraen:
This notebook has been crashing for at least three weeks due to #457. This PR adds a note to the HEASARC section stating that the data is temporarily unavailable and comments out the `heasarc_get_lightcurves()` function so that the rest of the notebook will run. If this PR goes through, I'll add a sub-issue to #457 to remind us to fully revert it once the bug is fixed (the serial cell will be obvious but the parallel cell will be easy to forget).


troyraen:
Sorry about the copilot review, I definitely was not trying to click on that. I'd love to just disable that so it stops being in my way when requesting other reviewers.

bsipocz:
Not sure how to disable it, I agree that it's most annoying. Have you tried turning of its visibility in https://github.com/settings/copilot/features? (I have also tried to look into the repo settings and disable everything I could; but I suspect it won't be enough. MS and Github is pushing it very very agressively)

troyraen:
Thanks for the reminder about those personal settings. I've disabled what I can, though I don't see an overall "visibility" option. After a brief search last night, I suspect the only way to do that would be at the Org level. I did see people recommending the browser extension https://github.com/gorhill/uBlock to block all kinds of things. Looks like something I would love but unfortunately Chrome recently stopped supporting it. Another entry in my growing frustration with Google.

PR #482, opened 2025-08-21
Update docs link and cloud languaging

troyraen:
- Update the link to rendered fornax-documentation since the docs and notebooks were separated.
- Change "in" -> "on" to comply with the [AWS Onboarding Handbook](https://assets.opendata.aws/aws-onboarding-handbook-for-data-providers-en-US.pdf) where page 18 says:

> Lastly, make sure it does not sound like your data is available “in” AWS. It is more appropriate
to reference that it is available “on” AWS.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@ This repo houses tutorial notebooks of fully worked science use cases for all us
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


bsipocz (approved):
Thanks!

Issue #489, opened 2025-08-27
MAINT: files in data/ directory?


bsipocz:
The force_photometry/data directory contains a bunch of files. Do we still need them in the git repo?

Assigning to Troy as git log point to him adding/reshuffling them.

PR #490, opened 2025-08-27
ENH: Adding manifest file and deploy notebooks branch

bsipocz:
The deployed content should be tracked in the manifest file, and then it will be deployed on the orphaned `deployed_notebooks` branch. This branch then be obtained for fornax deployment.

The logic should work (it does locally), but there isn't really a good way to test it besides merging and then fix any unforeseen permission issues (the force push may not work out of the box).

cc @troyraen 

(I have not added the data files from #489 to the manifest for now, we can add them if they are still necessary for the notebook)


bsipocz:
Foreseen problems: maintenance of the manifest file to ensure it's up-to-date with actual repo content. I would say cross that bridge when we get there, and certainly as a follow-up rather than blocking this PR.

PR #494, opened 2025-08-29
Deploy ztf_ps1_crossmatch.md

troyraen:
Cleanup:
- Simplify dependencies
- Make except catch KeyError only
- Remove whitespace
- Update comment about why light_curve_collector is excluded (unrelated to ztf_ps1_crossmatch)

Deploy:
- Add ztf_ps1_crossmatch to deployed_notebooks_manifest.in and tox.ini

### DIFF
--- a/crossmatch/ztf_ps1_crossmatch.md
+++ b/crossmatch/ztf_ps1_crossmatch.md
@@ -11,7 +11,7 @@ kernelspec:
   language: python
 ---
 
-# Cross-Matching ZTF and Pan-STARRS using LSDB
+# Cross-Match ZTF and Pan-STARRS using LSDB
 
 +++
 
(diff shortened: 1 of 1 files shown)


crossmatch/ztf_ps1_crossmatch.md
```diff
@@ -233,22 +233,22 @@ nworkers = [1, 2, 4, 8, 16, 32, 64, 128, 256, "default", "scale"]
...
+                                             categories=[str(x) for x in nworkers],
                                              ordered=True)
             b_N = b_N.sort_values("Nworkers")
 
             # only plot if there is more than 1 data point
             if len(b_N) > 1:
                 ax.plot("Nworkers", "time", marker="s", linestyle="-", data=b_N, label=f"{n} rows")
-        except:
+        except KeyError:
             pass
```
bsipocz, line 252: when does this happen, and when that happens, is that expected and normal?

bsipocz (commented):
You actually need to add this to a toctree, right now it will still be just an orphan page. 

I would just push one quick commit to here, and we can see what happens.

troyraen:
Thanks! I'm going to push one more to make the readme text generic for cross-match use cases.

crossmatch/ztf_ps1_crossmatch.md
troyraen: He is looping over dataframe indexes but not all of them have the data he's getting. I think it's "expected" and "normal" in that sense. But actually, I can move all but one of the code lines outside of the try/except. Pushing momentarily. It could probably be refactored to get rid of the try/except altogether but I'm not seeing exactly how to do that right now and not able to spend much more time on it today.

PR #495, opened 2025-08-30
Add Fornax contact info and standardize headings

### LINKED ISSUE #480: ENH: Add discourse link to each notebook so people know where to get help
Add a statement like the following to each notebook:

**Contact**: For help with this notebook, please open a topic in the [Fornax Community Forum](https://discourse.fornax.sciencecloud.nasa.gov/) "Support" category.

troyraen:
Closes #480 

Add Fornax contact info and standardize headings of all notebooks. Exception is ztf_ps1_crossmatch.md which is being done in #494 .


troyraen:
The rendering looks good to me. The build failure is just 2 warnings about crossmatch/README.md and crossmatch/ztf_ps1_crossmatch.md not being in a toctree. Those are being added in #494. I think this is ready.

light_curves/ML_AGNzoo.md
```diff
@@ -786,21 +778,16 @@ plt.legend(fontsize=12)
 plt.axis('off')
 ```
 
-## About this Notebook
-This notebook is created by the IPAC science platform team as a usecase of ML for time domain astrophysics. For questions contact: shemmati@caltech.edu
-
-**Author:** Shoubaneh Hemmati, Research scientist
-
-**Updated On:** 2025-03-31
```
bsipocz, line 794: Do we drop the last updated dates?

light_curves/ML_AGNzoo.md
```diff
@@ -786,21 +778,16 @@ plt.legend(fontsize=12)
...
+## About this notebook
 
+- **Authors:** Shoubaneh Hemmati (IRSA Research Scientist) and the Fornax team
```
bsipocz, line 783: somewhere maybe we should have a `fornax team` page? (definitely not on the notebook website though)

bsipocz (approved):
I don't really like the numbered headings in the right hand side "Contents" TOC, but can live with it. At least now things seems to be uniformed all over the notebooks.

troyraen:
> I don't really like the numbered headings in the right hand side "Contents" TOC, but can live with it. At least now things seems to be uniformed all over the notebooks.

I don't like the way we number these sections for a couple of reasons but I got pushback when requesting a change. So yes, this PR just makes them all uniform with syntax (eg, "1." instead of "1)"), etc.

light_curves/ML_AGNzoo.md
troyraen: Notebooks in this repo haven't historically included a "last updated" date. This one actually had two (one each at the top and bottom) and the scale_up notebook had one but none of the others did. So, not knowing what to put for the others, I opted to drop these ones. But I'm not opposed to adding them back in if we want.

light_curves/ML_AGNzoo.md
troyraen: That seems like a good idea but I'm not sure how to implement it in practice. Things like who all should be included and how do we keep it updated while still giving credit to people who leave for work they did while here.

light_curves/ML_AGNzoo.md
bsipocz: Let's wait until we switch to JB2, then this all can be header metadata, along with the authors.

light_curves/ML_AGNzoo.md
bsipocz: (we could definitely have a notebooks only docs, this could live there)

PR #498, opened 2025-08-30
Bump lsdb version

troyraen:
`lsdb` made significant improvements to their S3 read speed, so it would be good to upgrade if we can. This is not strictly necessary. Fine to close if problematic (though I think light_curve_collector will need to be bumped to 0.6.0 at least).


bsipocz (commented):
Does it come with any new nondirect minimum version change? If not then it should be ok to have these as minimum, otherwise I would not bump the minimum if changes are all internal and our code works with both versions.

PR #503, opened 2025-09-05
Cleanup data directory under forced photometry

### LINKED ISSUE #489: MAINT: files in data/ directory?

jkrick:
This will close #489 

It would be great if someone could double check that multiband_photometry runs for them, I am not entirely confident that my work on this is perfect.


troyraen (commented):
Can Chandra/chandra_multiband.fits also be removed?

jkrick:
> Can Chandra/chandra_multiband.fits also be removed?

yes, I removed it.  Also changed the runtime to reflect what it takes on small instance with no previously downloaded data (13 min.).  

@troyraen please approve and merge when ready.

troyraen (approved):
Thanks! I tested it as well and it ran without error and in about the same amount of time as what's now in the runtime statement.

PR #447, opened 2025-07-15
fixing ztf plotting bug

jkrick:
This PR addresses two key issues in the light curve plotting pipeline:
✅ Fixed: ZTF Zoom-In Panel Was Blank
- The zoom panel was not rendering because ZTF bands ("ztf_g", "ztf_r", "ztf_i") were being incorrectly filtered:
- Also fixed a logic error where band names were not matching due to incorrect unpacking from the groupby:
- this will close #419 

✅ Fixed: StopIteration from Exhausting Color Generator
- The previous color generator (next(COLORS)) raised StopIteration after plotting a few objects.  Replaced with itertools.cycle(...) to safely loop through the full color map indefinitely.

PR #506, opened 2025-09-13
Fix ZTF time min/max for plotting

troyraen:
This PR reverts part of #447 which is causing the plotting function to exit early, so only one figure gets displayed (instead of five) and its axes are not formatted correctly and the legend is missing. The problem is caused by `ztf_time_min_max` having an unexpected type, which itself is caused by the dataframes in `band_groups` having an unexpected index because the `groupby` was changed. The cell is actually throwing an error which probably went unnoticed because it's the last code cell and a figure still gets displayed.

I made a fresh clone of this repo on the Fornax Science Console today and ran all of the notebooks and checked them carefully for any errors. The scale_up notebook hit an error (I believe it was on an archive call) but it was transient. I ran scale_up ~~two~~ three more times later in the day and it's running fine. No other notebooks had problems.

### DIFF
--- a/light_curves/code_src/plot_functions.py
+++ b/light_curves/code_src/plot_functions.py
@@ -67,17 +67,12 @@ def create_figures(df_lc, show_nbr_figures, save_output):
 
         # Iterate over bands and plot light curves.
         # IceCube needs to be done last so that we know the y-axis limits.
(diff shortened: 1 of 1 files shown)


jkrick (approved):
I can confirm that 5 plots get made and look like what I remember them looking like.

PR #524, opened 2025-10-07
Fix lsdb ConeSearch import

### LINKED ISSUE #523: ConeSearch import broken in crossmatch notbeook after lsdb update
```python
from lsdb.core.search import ConeSearch
```
should change to

```python
from lsdb import ConeSearch
```

showed up in the[ image tests](https://github.com/nasa-fornax/fornax-images/actions/runs/18295466809/job/52093314325).

troyraen:
Closes #523

### DIFF
--- a/crossmatch/requirements_ztf_ps1_crossmatch.txt
+++ b/crossmatch/requirements_ztf_ps1_crossmatch.txt
@@ -1,6 +1,7 @@
 astropy
 pandas
 matplotlib
-# [full] needed for s3fs. >=0.6.2 for improved s3 read speed.
-lsdb[full]>=0.6.2
+# [full] needed for s3fs.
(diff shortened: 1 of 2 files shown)


troyraen:
This notebook appears to have successfully executed on CircleCI. As far as I can tell, the job failed due to spectra_collector exceeding a time limit. I doubt it was due to the SDSS TimeoutError that #525 will address because that would have thrown an actual error which would have been visible in the logs. The calls to get spectra from HST and JWST can take quite awhile so I'm guessing they took long enough this time that CircleCI's job time limit was hit. We should look into that and figure out if there's something we can do about it. But that's unrelated to changes made in this PR. So I think this PR is ready to go, but let me know @bsipocz .

bsipocz (approved):

PR #518, opened 2025-09-30
ENH: switching to JB2

bsipocz:
This switches the HTML build to the JB2/mystmd stack.

I didn't really change much on the content with the exception of adding cross references to the docs on the landing page -- you can all review how nicely the two can integrate.

We can similarly cross reference pieces from the IRSA tutorials, e.g. I expect linking to some parquet/HATS tutorials can be super beneficial, etc.

(and reverse, the same process can be used to add cross references or even sharing cells in the documentation page --> we don't need to maintain code there, but can cross reuse cells from other JB2 pages)

PR #531, opened 2025-10-10
CI: adding strict build option to fail on errors

bsipocz:
This builds on top of #518 and serves the purpose of helping mystmd debugging.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,21 +4,10 @@ jobs:
 
   build-docs:
     docker:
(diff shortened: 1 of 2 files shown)


bsipocz:
Note: we need to cleanup the TMP and its revert commit once this is all clean. Right now I leave them in so we don't loose the links to the respective CI logs (e.g. upstream https://github.com/jupyter-book/mystmd/issues/2549 issue was reported based on the build that was expected to fail due to these TMP commits).

PR #559, opened 2025-11-18
updating/clarifying docstrings

### LINKED ISSUE #544: Improve documentation of provided functions so they can be lifted out of the context of the notebook for which they were developed.
This ESRR item comes from Aurora's review. 

> it would be helpful to have more documentation about what the input and output for the pre-written functions are so that people will be able to more easily implement the pre-written code into their own work flows.
>
> For example, with the light curve search functions… it took me a lot of time to get these function to work on my own custom built table… I tried to look at the in-code comments to understand what the necessary format was for the table, but in wise_get_lightcurves for example, the input says “Table with the coordinates and journal reference labels of the sources”, which does not tell you what kind of coordinates and also seems to require these other entries like an “objectid” and “label” which really don’t serve much purpose for other table datasets

Action items:
- [ ] Update all *_get_lightcurves() function docstrings to describe each column that is expected to be present in the sample_table input. Include column name, data type, brief description of what it’s used for.
- [ ] All of our functions should already have docstrings containing this info. Make sure that’s true.
- [ ] Read all docstrings from a new user’s perspective and clarify them where appropriate. This could be a major task — given everything else that needs to be done, make this a quick once-over.

jkrick:
I have never touched so many docstrings in a morning.

This PR fully overhauls the documentation across all light-curve collection modules (ZTF, WISE, Gaia, Pan-STARRS, HCV, HEASARC, TESS/Kepler, IceCube, and Rubin), as well as the shared MultiIndexDFObject and sample_selection utilities. All functions now include clear, consistent, and user-focused docstrings that explicitly define required sample_table columns, parameter units, expected data types, and return formats. In particular, every *_get_lightcurves() function now documents the structure of the returned MultiIndexDFObject—including flux units, uncertainty definitions, time systems, and band naming conventions—ensuring uniform interpretation across heterogeneous archives. Ambiguous parameters (e.g., cone-search radii, mission-specific flux units, HEALPix pixel behavior, and Rubin TAP query requirements) were rewritten for clarity and scientific correctness. The PR also standardizes descriptions of the internal pandas DataFrame schema used throughout the pipeline, improving readability, reducing onboarding time for new users, and ensuring cross-archive consistency in downstream plotting and analysis workflows.

No code was touched in this PR

This will close #544 and covers all action items in that issue.

### DIFF
--- a/light_curves/code_src/data_structures.py
+++ b/light_curves/code_src/data_structures.py
@@ -8,8 +8,9 @@ class MultiIndexDFObject:
     """
     Container for multiband light-curve data stored as a Pandas MultiIndex DataFrame.
 
-    This class standardizes how all archives (Gaia, ZTF, WISE, Pan-STARRS,
-    TESS/Kepler, IceCube, Rubin, HEASARC, HCV) return light-curve data.
+    This class is intended to be used by the `*_get_lightcurves()` functions
+    that fetch data from the archives(Gaia, ZTF, WISE, Pan-STARRS,
+    TESS/Kepler, IceCube, Rubin, HEASARC, HCV) and return light-curve data.
     Each contributing function produces a pandas.DataFrame indexed by:
 
         objectid : int


light_curves/code_src/gaia_functions.py
```diff
@@ -133,12 +173,36 @@ def gaia_retrieve_epoch_photometry(gaia_table):
...
     gaia_table: Astropy Table
-        catalog of gaia source ids as well as the coords, objectid, and labels of our targets
+        Table returned by gaia_retrieve_catalog. Must include:
+        source_id : int
+            Gaia DR3 source identifier for the photometry request.
+
+        objectid : int
+            Used to label the MultiIndex rows later.
+
+        label : str
```
troyraen:
```suggestion
            Gaia DR3 source identifier for the photometry request.
        objectid : int
            Used to label the MultiIndex rows later.
        label : str
```

light_curves/code_src/gaia_functions.py
```diff
@@ -205,13 +269,48 @@ def gaia_clean_dataframe(gaia_df):
...
+                Uncertainty on the transit flux (electrons per second).
+
+            g_transit_time : float
+                Gaia mission time, which will be converted to MJD.
+
+            objectid : int
+                Identifier mapped from the input sample.
+
+            label : str
+                Literature or provenance label for the target.
```
troyraen:
```suggestion
            g_transit_flux : float
                G-band transit flux (electrons per second).
            g_transit_flux_error : float
                Uncertainty on the transit flux (electrons per second).
            g_transit_time : float
                Gaia mission time, which will be converted to MJD.
            objectid : int
                Identifier mapped from the input sample.
            label : str
                Literature or provenance label for the target.
```

light_curves/code_src/gaia_functions.py
```diff
@@ -205,13 +269,48 @@ def gaia_clean_dataframe(gaia_df):
...
+                Observation time converted from Gaia mission time to MJD.
+
+            objectid : int
+                Input sample object identifier.
+
+            band : str
+                Always "G" for Gaia epoch photometry.
 
+            label : str
+                Literature label associated with each source.
```
troyraen:
```suggestion
            flux : float
                Gaia G-band flux in electrons per second (e-/s).
            err : float
                Uncertainty on the Gaia flux (e-/s).
            time : float
                Observation time converted from Gaia mission time to MJD.
            objectid : int
                Input sample object identifier.
            band : str
                Always "G" for Gaia epoch photometry.
            label : str
                Literature label associated with each source.
```

light_curves/code_src/heasarc_functions.py
```diff
@@ -56,22 +70,75 @@ def make_hist_error_radii(missioncat):
...
+        Dictionary specifying which HEASARC mission catalogs to search and the
+        **maximum allowed error radius (degrees)** for each.
+
+        Keys must be valid HEASARC table names, for example:
+            {"FERMIGTRIG": 1.0, "SAXGRBMGRB": 3.0}
+
+        Values:
+            float (degrees)
+                Maximum localization error radius to accept from that catalog.
+                This filters out events with very large positional uncertainties.
```
troyraen: I'm not sure what the standard way to document dict key and value requirements is, but currently it reads a little strange with the Keys example being a complete dictionary. Also, is it helpful to keep the tip that gets deleted above about not using a large error radius because it can match many objects? Here's a try...

```suggestion
        Key : str
            A valid HEASARC table name, for example: "FERMIGTRIG" or "SAXGRBMGRB".
        Value : float (degrees)
            Maximum localization error radius to accept from that catalog.
            Recommended values are about 1.0 to 3.0.
            Very large values have the potential to match many target objects.
```

light_curves/code_src/heasarc_functions.py
```diff
@@ -23,8 +23,22 @@ def make_hist_error_radii(missioncat):
...
+
+            name : str
+                Mission source identifier.
+            ra : float (deg)
+                Right Ascension (ICRS).
+            dec : float (deg)
+                Declination (ICRS).
+            error_radius : float (deg)
+                Positional localization radius provided by the mission.
+    Notes
```
troyraen:
```suggestion
                Positional localization radius provided by the mission.

    Notes
```

light_curves/code_src/data_structures.py
troyraen:
```suggestion
   df_lc.data
   """
```

troyraen (approved):
Thanks for taking the time, this is a great improvement.

I read through for understandability and sanity checks. I didn't have time to check all the details like making sure that variable descriptions are accurate, but I can trust that those are right.

I'm much more used to the description (paragraphs before the Parameters section) containing all the info rather than having a separate "Notes" section stuck in at the bottom. But probably not worth changing now.

Most of my suggestions are just for consistent spacing, which was easy for me to both spot and suggest. I know some people don't care about spacing, but for people like me it  can make the reading experience much better.

light_curves/code_src/gaia_functions.py
```diff
@@ -14,16 +14,38 @@ def gaia_get_lightcurves(sample_table, *, search_radius=1 / 3600, verbose=0):
...
-     verbose : int
+        Cone-search radius in degrees for matching Gaia DR3 sources.
+        Default is 1/3600 (1 arcsec).
+    verbose : int
         How much to talk. 0 = None, 1 = a little bit , 2 = more, 3 = full
 
     Returns
     --------
-    MultiIndexDFObject with Gaia light curve photometry
+      df_lc : MultiIndexDFObject
```
bsipocz: indent is off
```suggestion
    df_lc : MultiIndexDFObject
```

light_curves/code_src/wise_functions.py
```diff
@@ -51,19 +84,44 @@ def wise_get_lightcurves(sample_table, *, radius=1.0, bandlist=["W1", "W2"]):
...
+            objectid : int  
+            label : str  
+            coord.ra : float (deg)  
+            coord.dec : float (deg)  
+            pixel : int
+                HEALPix pixel number at K=5 that overlaps the cone.
+
+    Notes
+    -----
+    * Each source may correspond to multiple HEALPix pixels.
```
bsipocz: These bullet points may not render as intended, but again, we don't do rendering so it doesn't matter for now.

bsipocz (approved):
Thank you @jkrick, this is a great improvement!

I'm not sure a couple of small things; if they are valid sphinx syntax, but I said in the individual comments that since we don't render these docstrings into API docs it doesn't really matter.

OTOH, we're started experimenting with docstring linting for pyvo (it's not yet figured out), if there is an interest I can look into bringing it into this repo, too. But honestly, I feel that would really be an overkill and unnecessary nitpicking here.

light_curves/code_src/data_structures.py
```diff
@@ -68,17 +106,25 @@ def append(self, x):
...
-            The MultiIndexDFObject instance to concatenate with.
+            Another instance whose light-curve data will be concatenated.
 
         Returns
         -------
         MultiIndexDFObject
-            A new MultiIndexDFObject instance with concatenated data.
+            A new container with combined data. The objectid values in
+            `other_df` are shifted upward so that there are no collisions
+            between the two samples.
```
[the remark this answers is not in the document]
jkrick: agreed, this PR isn't the right place to handle that.

jkrick (commented):
@bsipocz and @troyraen thanks for both of your reviews.  I certainly don't see all the same things you do.  

I'm certain the docstings will never be perfect, but we can agree they are better so I will merge.

PR #570, opened 2025-12-02
cleaning up build your own sample and bugfix wise function

### LINKED ISSUE #512: Make light_curve_collector more robust
[IRSA-7302](https://jira.ipac.caltech.edu/browse/IRSA-7302)

One of the Fornax ESRR reviewers ran the light_curve_collector notebook with a different list of targets and encountered problems. We should troubleshoot and try to update the notebook to handle it. She was using a modified version of our notebook and she sent me an executed copy. Contact me if you want it. I looked briefly and see two issues:

1. Warnings and TypeErrors from `tess_kepler_get_lightcurves()`.
2. A ValueError from `wise_get_lightcurves()`.

jkrick:
This will close #512 

This update adds a new internal validation utility for user-supplied sample_table objects and fixes a column-collision bug in the WISE light-curve loader.
Changes:
1. Added validate_sample_table() and surrounding explanatory text
- Introduces a structural validator ensuring that any user-constructed sample table is an astropy.table.Table with the required coord, objectid, and label columns and correct data types.
- Included inline documentation to guide users building their own samples.
2. Fixed column-overlap bug in load_lightcurves()
- The join between WISE detections and object metadata produced overlapping ra/dec columns and raised a ValueError.
Added rsuffix="_obj" to the join operation to avoid collision and preserve both coordinate sets safely.

### DIFF
--- a/light_curves/code_src/wise_functions.py
+++ b/light_curves/code_src/wise_functions.py
@@ -227,7 +227,7 @@ def load_lightcurves(locations, radius, bandlist):
         match_df["object_ilocs"] = object_ilocs
         match_df = match_df.set_index("object_ilocs").join(
             locs_df.reset_index(drop=True),
-            rsuffix="_obj"             # prevents ra/dec collisions
+            rsuffix="_obj"             # prevent name collisions if sample_table has extra columns
         )
 
         wise_df_list.append(match_df)
--- a/light_curves/light_curve_collector.md
+++ b/light_curves/light_curve_collector.md
@@ -155,74 +155,7 @@ or
 You can use [astropy's read](https://docs.astropy.org/en/stable/io/ascii/read.html) function to read in an input table
 to an [astropy table](https://docs.astropy.org/en/stable/table/)
 
-If you want to build your own sample_table, your table must contain one row per source and include exactly three required pieces of information: a coord column holding an astropy.coordinates.SkyCoord position for each object, a unique integer objectid, and a label column giving a short string describing that source’s origin (e.g., literature reference, sample name, or provenance tag). 
-
-```{code-cell} ipython3
----
-jupyter:
-  source_hidden: true
----
-def validate_sample_table(tbl):
-    """
-    Validate the structure of a user-supplied ``sample_table``.
-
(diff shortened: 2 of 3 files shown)


light_curves/code_src/wise_functions.py
```diff
@@ -225,7 +225,10 @@ def load_lightcurves(locations, radius, bandlist):
         match_df = pixel_tbl.take(pixel_ilocs).to_pandas()
         # attach identifying info like the objectid by joining with locs_df
         match_df["object_ilocs"] = object_ilocs
-        match_df = match_df.set_index("object_ilocs").join(locs_df.reset_index(drop=True))
+        match_df = match_df.set_index("object_ilocs").join(
+            locs_df.reset_index(drop=True),
+            rsuffix="_obj"             # prevents ra/dec collisions
```
troyraen:
```suggestion
            rsuffix="_obj"  # prevent name collisions if sample_table has extra columns
```

A collision can only happen if the user added an extra column(s) to `sample_table`, and then it could be with any column we load from WISE and not just ra or dec.

light_curves/light_curve_collector.md
```diff
@@ -155,7 +155,83 @@ or
...
+
+    if len(set(obj)) != len(obj):
+        raise ValueError("Column 'objectid' must contain unique values.")
+
+    if sorted(obj) != list(range(1, len(obj) + 1)):
+        raise ValueError("objectid must start at 1 and increment without gaps.")
+
+    # label check
+    if not all(isinstance(x, str) for x in tbl["label"]):
+        raise TypeError("Column 'label' must contain strings.")
```
troyraen:
```suggestion
    # objectid checks
    obj = tbl["objectid"]
    if len(set(obj)) != len(obj):
        raise ValueError("Column 'objectid' must contain unique values.")
```

Having this function is great but I think it's too strict. I don't find anywhere in our code that requires certain types for objectid or label or that requires objectid to be sequential starting at 1. Docstring would need to be updated as well.

troyraen (approved):
Thanks for looking into #512.

PR #579, opened 2025-12-05
addressing comments from #293"

### LINKED ISSUE #293: Forced Photometry Notebook cleanup
- [x] Add some information about what server to choose and how long this notebook takes to run if you are on the Fornax Science Console.
- [ ] Can any of the cloud subsetting work done in astropy be integrated with this notebook? 
- [x] "Section 3c. Calculate Forced Photometry with Straightforward but Slow Method" should be removed, as it's empty.
- [x] In general, it's a bit confusing to understand what information is provided in bullets versus comments versus markdowns. The bullets sometimes feel like leftovers from an outline.
- [ ] Everywhere in the markdown, use consistent and correct capitalization for GALEX, RA, Dec, Fornax, FITS.
- [ ] Everywhere in the markdown, use consistent capitalization and punctuation for sentences.
- [ ] In the Introduction, this statement appears: "The code will run on 2 different science platforms and makes full use of multiple processors to optimize run time on large datasets." That should probably be removed.
- [ ] One of the inputs is "RA and DEC within COSMOS catalog". We need to add a central RA and DEC and approximate size to make this user-friendly.
- [ ] One of the inputs is "mosaics of that region for IRAC and Galex". Is this really an input, or are the mosaics grabbed by the code?
- [ ] One of the introductory bullets for Section 1 says "Data exploration". I don't think this is very clear. It needs to be either expanded or deleted.
- [ ] In Section 2, should probably delete or modify "Use the fornax cloud access API to obtain the IRAC data from the IRSA S3 bucket" and "Details here may change as the prototype code is being added to the appropriate libraries, as well as the data holding to the appropriate NGAP storage as opposed to IRSA resources" and "Temporary solution, remove when the fornax API is added to the image" This is too much inside information that doesn't help the average user.
- [ ] In Section 2, some markdown says "# Temporarily add the cloud_access metadata to the Atlas response." I don't think users know what Atlas is.
- [ ] In Section 2, some markdown says "S3 access should be available from the daskhub and those who has their IRSA token set up." Does IRSA give out tokens? How does one get one?
- [ ] In Section 2, some markdown says "Adding function to download multiple files using the fornax API." What is the fornax API?
- [ ] The section titled "Use IVOA image search and Fornax download to obtain Galex from the MAST archive" has a cell with markdown "76k with 15arcmin diameter IRAC images". What does that mean? In that same section, some markdown says "# The column containing the on-prem access to fall back on if cloud is unavailable." Do we really need to provide that? The same section says "Download filtered products from AWS". Clarify what "Download" means in the context of computing in Fornax, and that this notebook should be able to run outside of Fornax.
- [ ] Section 3d says "Parallelization: we can either interate over the rows of the dataframe and run the four bands in parallel; or we could zip together the row index, band, ra, dec," Which one was actually done?
- [ ] Section 3d reword "#proove we can do this for one object"
- [ ] Section 3d missplelling "daraframe"

jkrick:
This will close #293 

Sorry, this issue is old, but I just found it!  Most of these items have been taken care of in previous PRs, but this one addresses whatever remained.

### DIFF
--- a/forced_photometry/multiband_photometry.md
+++ b/forced_photometry/multiband_photometry.md
@@ -24,7 +24,7 @@ By the end of this tutorial, you will be able to:
 
 ## Introduction
 
-This code performs photometry in an automated fashion at all locations in an input catalog on 4 bands of IRAC data from IRSA and 2 bands of Galex data from MAST.  The resulting catalog is then cross-matched with a Chandra catalog from HEASARC to generate a multiband catalog to facilitate galaxy evolution studies.
+This code performs photometry in an automated fashion at all locations in an input catalog on 4 bands of IRAC data from IRSA and 2 bands of GALEX data from MAST.  The resulting catalog is then cross-matched with a Chandra catalog from HEASARC to generate a multiband catalog to facilitate galaxy evolution studies.
 
 If you run this code as is, it will only look at a small region of the COSMOS survey, and as a result, the plots near the end will not be so very informative.  We do this for faster runtimes and to show proof of concept as to how to do this work.  Please change the radius in section 1. to be something larger if you want to work with more data, but expect longer runtimes associated with larger areas.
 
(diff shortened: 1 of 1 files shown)


vandesai1:
Looks good to me!

vandesai1 (approved):

Issue #587, opened 2025-12-16
Update notebooks for lsdb v0.8 `crossmatch()` changes


troyraen:
lsdb v0.8 will be released soon. The `crossmatch()` function will include a breaking change. The default suffix behavior will change from applying suffixes to all columns to only applying suffixes to overlapping columns. The following notebooks use that function:

- light_curve_collector.md (ztf_functions.py and panstarrs_functions.py)
- ztf_ps1_crossmatch.md 

Update any column names that will break as a result of this change. I took a brief look and the only two I see are in ztf_functions.py: `"objectid_sample"` and `"label_sample"`.

troyraen:
Ya, dual support probably makes the most sense.

PR #588, opened 2025-12-16
MAINT: limit lsdb version as we know API breakage is coming

bsipocz:
This is to add an instant workaround for the incompatibilities coming in the next release. For context and long term solution, see https://github.com/nasa-fornax/fornax-demo-notebooks/issues/587


troyraen (approved):
Thanks!

PR #608, opened 2026-01-16
Optimize lsdb

### LINKED ISSUE #420: ENH: Experiment with how to optimize lsdb.from_dataframe in light curve notebooks

jkrick:
This will close #420 

- explicitly set suffix_method="all_columns" for all LSDB crossmatch() and join() calls to preserve the current column-naming behavior and avoid breakage when the LSDB default changes.
- changed client parameters based on recommendations as a result of benchmarking.
- increased verbiage in the main notebook about warnings being present that can be ignored.

### DIFF
--- a/light_curves/code_src/ztf_functions.py
+++ b/light_curves/code_src/ztf_functions.py
@@ -54,7 +54,7 @@ def ztf_get_lightcurves(sample_table, *, radius=1.0):
     
     # 1) Start Dask client & read full ZTF light-curve catalog
     # Use multiple workers with a single thread per worker for better performance on Fornax
-    client = Client(n_workers=2, threads_per_worker=1, memory_limit=None)
+    client = Client(threads_per_worker=1, memory_limit=None)
     ztf_lc = lsdb.read_hats(
         's3://ipac-irsa-ztf/contributed/dr23/lc/hats/',
         columns=[


light_curves/code_src/panstarrs_functions.py
```diff
@@ -99,12 +100,14 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
...
-        suffixes=["", ""]
+        suffixes=["", ""],
+        suffix_method="all_columns",
     )
 
     # Create default local cluster
     # here is where the actual work gets done
-    with Client():
+    # Use multiple workers with a single thread per worker for better performance on Fornax
+    with Client(n_workers=2, threads_per_worker=1, memory_limit=None):
```
troyraen: Is there a reason to limit to 2 workers? It will mean that these functions can't take advantage of larger server sizes (more cpu).

troyraen (commented):
Thanks for investigating a bunch of stuff. I can't test it right now because I can't get onto fornax. Leaving a couple comments based on reading.

In the ZTF function, can you try moving the "filterid" query into the catalog read? It's an example of this from #420 that looks like low-hanging fruit for us:

> One thing I learned recently is that dask will do best if filter criteria is provided directly in the open_catalog() command.

I tried something similar when investigating #609 and it looked like a good improvement, though I didn't do a proper comparison. The code looks like this:

```python
# Remove the current `ztf_lc = lsdb.read_hats()` call.
for fid, band_name in band_map.items():
    # Remove `ztf_lc.query()`. Put that filter in `open_catalog()`.
    ztf_band = lsdb.open_catalog(
        "s3://ipac-irsa-ztf/contributed/dr23/lc/hats/",
        filters=[("filterid", "==", fid)],
        columns=[...],  # Same as before
    )

    # Now same as before
    matched = sample_lsdb.crossmatch(...)
    df = matched.compute()
```

troyraen:
I opened a new issue for most of my comments. Let's merge this one. But how about the number of workers? Is there a reason to limit to 2, @jkrick? It usually seems better to leave the default so that it uses one worker per cpu.

troyraen:
The build error is #619 in the crossmatch notebook, which has already been fixed in the main branch. I tested this manually on Fornax and it was fine. Fornax image CI test is also passing.

PR #629, opened 2026-02-03
MAINT: remove upper limit on fsspec

bsipocz:
Upstream has been fixed, so we can update our pin. Most lsdb/hats/nested-pandas will have additional pins for it, but at this point we can trust that the solver will take care of it.


troyraen (approved):
LGTM

Issue #651, opened 2026-02-26
Merge spectral cubes into spectroscopy


troyraen:
- [ ] Move the spectral cubes files into the existing spectroscopy directory.
- [ ] `spectroscopy/code_src/` already contains code files for spectra_collector. We should probably differentiate the two sets of code files. Need to decide if we want them in separate directories or handle it in some other way.
- [ ] Merge the content of the current `spectral_cubes/README.md` with `spectroscopy/spectroscopy.md`. We should wait until we actually release the notebook before pushing these changes to `main` since any changes to `spectroscopy/spectroscopy.md` will automatically be released. But the changes can be prepped as part of this issue since `spectral_cubes/README.md` will be deleted.

jkrick:
The last topic of merging readme's has an open PR but is waiting for other things to happen first.

PR #658, opened 2026-03-04
Adding spectral_cubes to the deployed notebooks

### LINKED ISSUE #654: Add the spectral cubes notebook to CI testing
Add the spectral cubes notebook to CI testing. The number of sources in the input sample may need to be reduced in order for the notebook to run on the available server size in a reasonable amount of time. If so, do similar to what we do in other notebooks where we keep the full sample available for people who want to use it, but the notebook processes a smaller sample by default.

### LINKED ISSUE #655: Deploy the spectral cubes notebook
Deploy the spectral cubes notebook to the website and Fornax.

- [ ] Add an entry to `toc.yml`
- [ ] Add the notebook, code files, and requirements file to `deployed_notebooks_manifest.in`
- [ ] The text for `spectroscopy/spectroscopy.md` will be prepped (but not released) in #651. Release it as part of this deployment.

This is blocked by #651, #652, and probably #654.

bsipocz:
This resolves #655 as a prerequisite to close #654 

(Also note, that the content here will change as the moving around of the notebook is done as part of #651; merging this PR will have to be done after that restructure is merged)

The testing on GHA is expected to fail as the kernel is not yet available on Fornax; but also some other issues are expected as we likely need to scale back the resources this notebook requires.

### DIFF
--- a/deployed_notebooks_manifest.in
+++ b/deployed_notebooks_manifest.in
@@ -81,3 +81,7 @@ spectroscopy/data/README.md
 spectroscopy/output/README.md
 spectroscopy/requirements_spectra_collector.txt
 spectroscopy/spectra_collector.md
+spectral_cubes/emission_search_cubes.md
+spectral_cubes/requirements_emission_search_cubes.txt
+spectral_cubes/code_src/parse_polygon.py
+spectral_cubes/code_src/find_extended_emission.py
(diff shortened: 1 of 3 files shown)


bsipocz:
OK, so this is running unsustainably long as is (45m and counting when writing this), we definitely need some cut backs.

bsipocz:
@adrianlucy - since you asked about the timeline of deploying the notebook: currently it's timing out, and running just too long. I suspect we can get this through the finish line faster if you could help with a meaningful scale downs that would allow this to run in a reasonably timeframe and still give enough learning context to the users.

PR #668, opened 2026-04-08
Restrict Simbad to Taurus Molecular Cloud

troyraen:
Trying to reduce the runtime of the spectral cubes notebook for #654.

- Restrict the initial Simbad query to the Taurus Molecular Cloud region.
- Demo object PER-EMB-33 isn't in the Taurus Molecular Cloud, so switch that to TAU042021.

I ran this several times on the fornax medium instance with good results. It reduced the runtime of most of the long-running cells:
- First Simbad query 1m -> 1s
- First `pool.map()` call 4m -> 20s
- Last `Observations.get_cloud_uris()` call 2m -> 20s-50s
- Second `pool.map()` call 11m -> 1m30s

Biggest remaining potential issue that I see is the `Observations.query_criteria()` call. Changing the Simbad query doesn't affect this cell, and I've seen it take up to 4m on fornax.

### DIFF
--- a/deployed_notebooks_manifest.in
+++ b/deployed_notebooks_manifest.in
@@ -81,3 +81,7 @@ spectroscopy/data/README.md
 spectroscopy/output/README.md
 spectroscopy/requirements_spectra_collector.txt
 spectroscopy/spectra_collector.md
+spectral_cubes/emission_search_cubes.md
+spectral_cubes/requirements_emission_search_cubes.txt
+spectral_cubes/code_src/parse_polygon.py
+spectral_cubes/code_src/find_extended_emission.py
--- a/spectral_cubes/emission_search_cubes.md
+++ b/spectral_cubes/emission_search_cubes.md
@@ -883,8 +883,9 @@ else:
 
 # Execute
 with Pool(num_workers) as pool:
-    results = pool.map(line_search,
-                       [obs for _, obs in copy_yso_jwst_obstable_iterable])
+    results = list(pool.imap_unordered(line_search,
+        (obs for _, obs in copy_yso_jwst_obstable_iterable),
+        chunksize=15))
 ```
 
 ```{code-cell} ipython3
--- a/spectral_cubes/emission_search_cubes.md
+++ b/spectral_cubes/emission_search_cubes.md
@@ -875,11 +875,7 @@ We will make one minor modification to the parallelization approach that we used
 # Use pandas to make the iterable
 copy_yso_jwst_obstable_iterable = copy_yso_jwst_obstable.to_pandas().iterrows()
 
-# Increase the number of workers to compensate for I/O bound tasks
-if num_cores<=4:
-    num_workers = round(num_cores*2)
-else:
-    num_workers = round(num_cores*1.4)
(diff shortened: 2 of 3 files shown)


troyraen:
I ran the CircleCI job three times:

1. Succeeded in 11m
2. Failed in 1h
3. Failed in 1h

Both failures resulted from `context deadline exceeded` with no other relevant info in the logs. The Resources figures show 100% RAM usage at about 5m and CPU usage goes to 0% about 30s later. The job that succeeded hit a max of 98% RAM usage at about 10m.

The CircleCI machine has 4CPU and 8G RAM.

troyraen:
Changing the second `pool.map()` -> `pool.imap_unordered()` brought the RAM usage down just enough for it to complete successfully, though it still hit a max of 98%.

Adding the change to `num_workers = num_cores` (down from `2*num_cores`) in that same call brought the max RAM usage down to 76%. Testing on fornax, this increases the runtime of that cell by about 20s (1m30s -> 1m50s).

troyraen:
@adrianlucy I've tried several things attempting to get the spectral cubes notebook to run successfully on CircleCI. You can read the details in this PR and in #669. In summary:

The blocking problem is that it runs out of memory on the cell that does the second `pool.map()` call. The machine has 4CPU and 8G RAM. The runtime is somewhat high but that's not what's making the job fail.

Restricting Simbad to Taurus Molecular Cloud (c4e75755ddaca954eca7ebb666eb717a1ef8bce7) + reducing num_workers to num_cores in the second pool.map() call (4d4f3be7cedd66f9a89683b1771f293f970dbea1) looks like it will do the trick. Only potential issue I see is if/when `Observations.query_criteria()` takes an especially long time. I saw it timeout once in the ~20 times I've run it (hit astroquery's 10 min limit). We could probably live with that unless it starts happening more often.

562e7d597d14840d0fd0865cdb0aed421f7607f6 cuts the time range of JWST observations to only what's needed to include the two demo objects (my guess about how to implement your preference expressed in https://github.com/nasa-fornax/fornax-demo-notebooks/pull/658#issuecomment-4006536375). That would probably prevent the `Observations.query_criteria()` timeout but it's not enough to bring the job under the RAM limit. I'm not sure how much it needs to be reduced, but we could try reducing num_workers to num_cores.

How do you want to proceed? We could do some combination of what I've tried or something else if you have other ideas.

adrianlucy:
Thank you so much for your work on this, Troy! I'll look at this today and give you an answer hopefully today (worst case end of the week).

adrianlucy:
Sorry for the delay. Looking at this today.

troyraen:
And idea what you want to do here @adrianlucy? No particular rush on our end, just want to make sure all your hard work doesn't fall by the wayside!

adrianlucy:
Sorry about that, juggling too many things. Will get you an answer within a week. (Hopefully but not necessarily tomorrow.) Thank you for the reminder and for the work you put into this as well!

troyraen:
I totally get it! And just to let you know about schedules on our end, most of us will be at IVOA and then AAS starting end of next week, so we'll be more unavailable than usual.

PR #669, opened 2026-04-09
Restrict MAST to obs time range

troyraen:
Trying to reduce the runtime of the spectral cubes notebook for #654.

- Restrict the `Observations.query_criteria()` call to `t_min` in range [59824.6, 60007.4], which includes the two demo objects TMC1A and PER-EMB-33.

I ran this several times on the fornax medium instance with reasonably good results. It reduced the runtime of most of the long-running cells:
- `Observations.query_criteria()` call 30s-4m -> 40s-1m30s
- First `pool.map()` call 4m -> 30s
- Last `Observations.get_cloud_uris()` call 2m -> 20s
- Second `pool.map()` call 11m -> 2m30s

#668 is slightly better overall, but the gain here for the `Observations.query_criteria()` might be more important for the CI since the runtime for that cell can vary widely.

### DIFF
--- a/deployed_notebooks_manifest.in
+++ b/deployed_notebooks_manifest.in
@@ -81,3 +81,7 @@ spectroscopy/data/README.md
 spectroscopy/output/README.md
 spectroscopy/requirements_spectra_collector.txt
 spectroscopy/spectra_collector.md
+spectral_cubes/emission_search_cubes.md
+spectral_cubes/requirements_emission_search_cubes.txt
+spectral_cubes/code_src/parse_polygon.py
+spectral_cubes/code_src/find_extended_emission.py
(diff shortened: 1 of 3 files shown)


bsipocz:
I would think we could use #658 as a base for these PRs, so the mandatory or agreed upon stuff can end up in that PR straight away. And then chop away parts until things fit into the runners.

(currently we hit the memory limit ~4mins in. It's possible that we don't run into it with GHA, but nevertheless it would be better if we could cut down resourse usage and have the previews and such in PRs.)

troyraen:
I ran CircleCI once without restricting `t_min` to get a baseline. It failed on the `Observations.query_criteria()` call with the error:

```
TimeoutError: Timeout limit of 600 exceeded.
```

Then I ran CircleCI twice with `t_min` restricted:

1. Failed in 1h
2. Failed in 1h

Both failures resulted from `context deadline exceeded` with no other relevant info in the logs. The Resources figures show 100% RAM usage at about 6m and CPU usage goes to 0% about 30s later.

The CircleCI machine has 4CPU and 8G RAM.

troyraen:
Adding the Simbad restriction to Taurus Molecular Cloud isn't enough to bring it under the memory limit.

PR #670, opened 2026-04-15
moved spectral cubes files into spectroscopy directory

jkrick:
takes care of 1-2 from #651


toc.yml
```diff
@@ -13,6 +13,7 @@ project:
       file: spectroscopy/spectroscopy.md
       children:
         - file: spectroscopy/spectra_collector.md
+        - file: spectroscopy/emission_search_cubes.md
```
bsipocz: unfortunately there are content changes required for this notebook to be able to run in CI, so we can't yet add it to the website. There are a couple other PRs that are trying to deal with that step.

So, I would recommend removing this line from this PR; then the rest of it can be merged right now.
```suggestion
```

bsipocz (commented):
This can be merged right now if the TOC addition is removed

jkrick:
update to toc removed.  please merge when you are ready.

bsipocz (approved):
Thank you!

PR #673, opened 2026-05-13
Make lsdb calls more efficient and upgrade lsdb

### LINKED ISSUE #587: Update notebooks for lsdb v0.8 `crossmatch()` changes

### LINKED ISSUE #646: Make lsdb calls more efficient
Follow up to #420.

Following are comments from #608 

----

In the ZTF function, can you try moving the "filterid" query into the catalog read? It's an example of this from #420 that looks like low-hanging fruit for us:

> One thing I learned recently is that dask will do best if filter criteria is provided directly in the open_catalog() command.

I tried something similar when investigating #609 and it looked like a good improvement, though I didn't do a proper comparison. The code looks like this:

```python
# Remove the current `ztf_lc = lsdb.read_hats()` call.
for fid, band_name in band_map.items():
    # Remove `ztf_lc.query()`. Put that filter in `open_catalog()`.
    ztf_band = lsdb.open_catalog(
        "s3://ipac-irsa-ztf/contributed/dr23/lc/hats/",
        filters=[("filterid", "==", fid)],
        columns=[...],  # Same as before
    )

    # Now same as before
    matched = sample_lsdb.crossmatch(...)
    df = matched.compute()
```

_Originally posted by @troyraen in https://github.com/nasa-fornax/fornax-demo-notebooks/pull/608#pullrequestreview-3673244828_

----

I asked claude to look at the panstarrs function since that one is particularly painful, having to `join()` as well as `crossmatch()`, and also gives the large task graph warning. The following two pieces of feedback seem worth following up on. Maybe you checked them already?

My caveats: These responses seem plausible but I haven't checked them in much detail. Also, like you said earlier, the large task graph warning isn't a problem by itself and there's no need to address it if the only improvement is that the warning goes away. (I suspect that it's a real issue and not just an empty warning, but whether we can do anything about it, idk.)

----

> **For Speed and Memory: Materialize the Crossmatch First**
> The best approach for chaining complex operations like crossmatch() → join() is to save the intermediate crossmatch result to disk, then reload it for the join. This breaks the computation into two smaller graphs instead of one massive one.
> 
> ```python
> matched_objects = sample_lsdb.crossmatch(...)  # Same as we do now
> matched_objects.to_hats("/path/to/xmatch_hats")
> matched_objects = lsdb.open_catalog("/path/to/xmatch_hats")
> # Now join
> ```
> 
> Alternately, if you have enough RAM, use persist() instead of disk:
> 
> ```python
> from distributed import wait
> 
> matched_objects = sample_lsdb.crossmatch(...)  # Same as we do now
> matched_objects = client.persist(matched_objects)
> wait(matched_objects)  # block until crossmatch completes
> # Now join
> ```

Thinking about larger sample sizes, my initial thought is that the recommended approach (`to_hats()`) would be better. Should reduce memory usage and also allow not to start completely over if the server dies during the join.

----

> The "large graph" warning is caused by from_dataframe(). When you use lsdb.from_dataframe(), the entire pandas DataFrame gets embedded into the Dask task graph. This is serialized and sent to every worker, causing the 11+ MiB warning.
>
> Solutions:
> Option A: Save targets to HATS format first
> 
> ```python
> lsdb.from_dataframe(sample_df, ...).to_hats("/path/to/targets_sample_hats")
> sample_lsdb = lsdb.open_catalog("/path/to/targets_sample_hats")
> # Now crossmatch
> ```
> 
> Option B: Use client.scatter() for the DataFrame
> 
> ```python
> sample_df_future = client.scatter(sample_df, broadcast=True)
> sample_lsdb = lsdb.from_dataframe(sample_df_future, ...)
> # Now crossmatch
> ```

Again thinking about larger sample sizes, option A might be better. We could do that once and use it for both panstarrs and ztf.

_Originally posted by @troyraen in https://github.com/nasa-fornax/fornax-demo-notebooks/issues/608#issuecomment-3762612997_

jaladh-singhal:
Fixes #646, also fixes #587 

Regarding the lsdb upgrade, the warning about explicitly setting `suffix_method` is still there (even in v0.9). It is just cautioning because default will change from `all_columns` to `overlapping_columns`. #608 already explicitly defined `suffix_method="all_columns"` so we were safe to upgrade lsdb even then. I have set the same in `ztf_ps1_crossmatch.md` but I changed `light_curve_collector` functions to  use `overlapping_columns` and made relevant dataframe column name updates because I got the impression that's what #587 was suggesting.

## Benchmarking
### 6 May — lsdb calls updated                                                                                                         
                                      
  | Change | Before (s) | After (s) | Improvement |                                                                                     
  |---|---|---|---|
  | ZTF: `filterid` filter moved into `open_catalog()` | 173.96 | 136.11 | −21.8% |                                                      
  | PanSTARRS: open detection catalog with `PixelSearch` filter | 85 | 45 | −47.1% |
                                                                                                                                         
### 12 May — lsdb calls updated + lsdb upgraded to ≥0.8.1       
                                                                                                                                                                                                                
  | Change | Before (s) | After (s) | Improvement |                                                                                     
  |---|---|---|---|
  | ZTF: `filterid` filter moved into `open_catalog()` | 198 | 147 | −25.7% |                                                            
  | PanSTARRS: open detection catalog with `PixelSearch` filter  | 100 | 61 | −39.0% |

Note: Timings in "Before" column (executing the tutorial in fornax production) differ between the two tables (e.g. 173s vs 198s before) likely just due to environment/dask cluster variability on the day.

### DIFF
--- a/crossmatch/requirements_ztf_ps1_crossmatch.txt
+++ b/crossmatch/requirements_ztf_ps1_crossmatch.txt
@@ -1,11 +1,7 @@
 astropy
 pandas
 matplotlib
-# >=0.6.5 for ConeSearch import. https://github.com/nasa-fornax/fornax-demo-notebooks/issues/523
-# Upper limit <0.8 is for instant workaround of coming incompatibility https://github.com/nasa-fornax/fornax-demo-notebooks/issues/587
-lsdb>=0.6.5,<0.8
-# Indirect dependency of lsdb via hats. Remove once our minimum required version of lsdb/hats is 0.8.1
-cdshealpix<0.8
+lsdb>=0.8.1
 dask
 # Required by lsdb to read files from AWS S3, revise when lsdb[s3fs] is supported by required minimum version.
 s3fs
--- a/light_curves/requirements_light_curve_collector.txt
+++ b/light_curves/requirements_light_curve_collector.txt
@@ -14,11 +14,7 @@ astroquery>=0.4.8.dev0
 acstools
 lightkurve
 alerce
-# >=0.6.2 for improved s3 read speed.
-# Upper limit <0.8 is for instant workaround of coming incompatibility https://github.com/nasa-fornax/fornax-demo-notebooks/issues/587
-lsdb>=0.6.2,<0.8
-# Indirect dependency of lsdb via hats. Remove once our minimum required version of lsdb/hats is 0.8.1
-cdshealpix<0.8
+lsdb>=0.8.1
 # Required by lsdb to read files from AWS S3, revise when lsdb[s3fs] is supported by required minimum version.
 s3fs
 # Required by lsdb, a few versions are known to cause issues
--- a/crossmatch/ztf_ps1_crossmatch.md
+++ b/crossmatch/ztf_ps1_crossmatch.md
@@ -4,7 +4,7 @@ jupytext:
     extension: .md
     format_name: myst
     format_version: 0.13
-    jupytext_version: 1.17.2
+    jupytext_version: 1.19.1
 kernelspec:
   display_name: py-ztf_ps1_crossmatch
   language: python
@@ -170,7 +170,8 @@ ps1 = lsdb.open_catalog(ps1_path, columns=["objName","objID","raMean","decMean"]
 
 ```{code-cell} ipython3
 # Setting up the cross-match actually takes very little time
-ztf_x_ps1 = ztf_piece.crossmatch(ps1, radius_arcsec=1, n_neighbors=1, suffixes=("_ztf", "_ps1"))
+ztf_x_ps1 = ztf_piece.crossmatch(ps1, radius_arcsec=1, n_neighbors=1,
+                                 suffixes=("_ztf", "_ps1"), suffix_method='all_columns')
 ztf_x_ps1
 ```
 
--- a/light_curves/code_src/lsdb_utils.py
+++ b/light_curves/code_src/lsdb_utils.py
@@ -1,0 +1,31 @@
+import lsdb
+import pandas as pd
+
+
+def sample_table_to_lsdb(sample_table):
+    """Convert a sample astropy Table to an LSDB catalog.
+
+    Parameters
+    ----------
+    sample_table : astropy.table.Table
+        Table with columns: coord (SkyCoord), objectid (int), label (str).
+
+    Returns
+    -------
+    lsdb.Catalog
+        Spatially partitioned catalog ready for crossmatch or join.
+    """
+    # SkyCoord cannot be stored directly in a DataFrame; extract ra/dec explicitly
+    sample_df = pd.DataFrame({
+        'objectid': sample_table['objectid'],
+        'ra_deg': sample_table['coord'].ra.deg,
+        'dec_deg': sample_table['coord'].dec.deg,
+        'label': sample_table['label'],
+    })
+    return lsdb.from_dataframe(
+        sample_df,
+        ra_column="ra_deg",
+        dec_column="dec_deg",
+        margin_threshold=10,
+        drop_empty_siblings=True,
+    )
--- a/light_curves/code_src/panstarrs_functions.py
+++ b/light_curves/code_src/panstarrs_functions.py
@@ -4,6 +4,7 @@ import pandas as pd
 from dask.distributed import Client
 
 from data_structures import MultiIndexDFObject
+from lsdb_utils import sample_table_to_lsdb
 
 # panstarrs light curves from hats catalog in S3 using lsdb
 
@@ -56,22 +57,8 @@ def panstarrs_get_lightcurves(sample_table, *, radius=1):
                  "nStackDetections",  # some other data to use
                  ]
     )
-    # convert astropy table to pandas dataframe
-    # special care for the SkyCoords in the table
-    sample_df = pd.DataFrame({'objectid': sample_table['objectid'],
-                              'ra_deg': sample_table['coord'].ra.deg,
-                              'dec_deg': sample_table['coord'].dec.deg,
-                              'label': sample_table['label']})
-
-    # convert dataframe to hipscat
-    sample_lsdb = lsdb.from_dataframe(
-        sample_df,
-        ra_column="ra_deg",
-        dec_column="dec_deg",
-        margin_threshold=10,
-        # Optimize partition size
-        drop_empty_siblings=True
-    )
+    # convert astropy table to lsdb catalog
+    sample_lsdb = sample_table_to_lsdb(sample_table)
 
     # plan to cross match panstarrs object with my sample
     # only keep the best match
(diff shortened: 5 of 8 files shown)


bsipocz (commented):
I'm very very confused here: this is very much expected to fail tests with the fornax images as the previous and current requirements are incompatible with each other as the image should have been built from the previous list. So something is seriously off there.



OTOH, the circleCI job should be passing, I'm looking into what hidden incompatibilities we're running into.

bsipocz:
OK, the circleCI conflict is clear, we have a few upper pins for the light curve classifier notebook brought in by sktime, but maybe those can be dropped now. I'm opening a separate issue for that. 

```
# We need to manully add the upper pin both numpy and pandas as the resolver is somehow doesn't
# pick it up properly from sktime. Remove once sktime removes the upper limit.
numpy<2.3
pandas[parquet]<2.3
```

jaladh-singhal:
> OK, the circleCI conflict is clear, we have a few upper pins for the light curve classifier notebook brought in by sktime, but maybe those can be dropped now. I'm opening a separate issue for that.

I didn't realize the different dependency environments are merged together in the build docs. Is it not feasible to run each tutorial within its own environment for building docs (like we are doing for tests)?

crossmatch/requirements_ztf_ps1_crossmatch.txt
```diff
@@ -1,11 +1,7 @@
 astropy
 pandas
 matplotlib
-# >=0.6.5 for ConeSearch import. https://github.com/nasa-fornax/fornax-demo-notebooks/issues/523
-# Upper limit <0.8 is for instant workaround of coming incompatibility https://github.com/nasa-fornax/fornax-demo-notebooks/issues/587
-lsdb>=0.6.5,<0.8
-# Indirect dependency of lsdb via hats. Remove once our minimum required version of lsdb/hats is 0.8.1
-cdshealpix<0.8
+lsdb>=0.8.1
```
bsipocz: the notebook works without this version limit, so can we try if removing it would resolve the rendering job?
```suggestion
lsdb
```

light_curves/requirements_light_curve_collector.txt
```diff
@@ -14,11 +14,7 @@ astroquery>=0.4.8.dev0
 acstools
 lightkurve
 alerce
-# >=0.6.2 for improved s3 read speed.
-# Upper limit <0.8 is for instant workaround of coming incompatibility https://github.com/nasa-fornax/fornax-demo-notebooks/issues/587
-lsdb>=0.6.2,<0.8
-# Indirect dependency of lsdb via hats. Remove once our minimum required version of lsdb/hats is 0.8.1
-cdshealpix<0.8
+lsdb>=0.8.1
```
bsipocz: this notebook seems to be working with olders lsdb, too, so we can probably drop the version pin
```suggestion
lsdb
```

bsipocz (changes requested):
Is there any particular reason for lower pinning the lsdb version? Both notebooks seems to be running fine with lsdb 0.7.3, so I would just recommend dropping the pin.

An this finding actually resolves my concern of how/why the fornax image tests pass and if it was just a lucky fluke. Nope, the notebooks actually seem to work OK.

(if you really need it, I suppose we can patch it around just for the rendering jobs though)

jaladh-singhal:
> Is there any particular reason for lower pinning the lsdb version? Both notebooks seems to be running fine with lsdb 0.7.3, so I would just recommend dropping the pin.

I only added it because of the following comment on transitive dependency that you had:
```
# Indirect dependency of lsdb via hats. Remove once our minimum required version of lsdb/hats is 0.8.1
cdshealpix<0.8
```

Besides this, we get speed improvements (if any) from latest versions but if that's causing CI problems, I'm ok to remove it and see if things work.

jaladh-singhal:
Oh well, @bsipocz unpinning lsdb is failing the docs build now due to your original comment about cdshealpix.
These are the versions reported by CircleCI logs:
```
lsdb==0.7.3
hats==0.7.3
cdshealpix==0.8.1 
```

bsipocz:
Ok, let me patch it around in a different way then

crossmatch/requirements_ztf_ps1_crossmatch.txt
```diff
@@ -1,11 +1,7 @@
 astropy
 pandas
 matplotlib
-# >=0.6.5 for ConeSearch import. https://github.com/nasa-fornax/fornax-demo-notebooks/issues/523
-# Upper limit <0.8 is for instant workaround of coming incompatibility https://github.com/nasa-fornax/fornax-demo-notebooks/issues/587
-lsdb>=0.6.5,<0.8
-# Indirect dependency of lsdb via hats. Remove once our minimum required version of lsdb/hats is 0.8.1
-cdshealpix<0.8
+lsdb
```
troyraen: The crossmatch notebook still uses lsdb.ConeSearch, so I think we need to keep the lower pin.

light_curves/requirements_light_curve_collector.txt
```diff
@@ -14,11 +14,7 @@ astroquery>=0.4.8.dev0
 acstools
 lightkurve
 alerce
-# >=0.6.2 for improved s3 read speed.
-# Upper limit <0.8 is for instant workaround of coming incompatibility https://github.com/nasa-fornax/fornax-demo-notebooks/issues/587
-lsdb>=0.6.2,<0.8
-# Indirect dependency of lsdb via hats. Remove once our minimum required version of lsdb/hats is 0.8.1
-cdshealpix<0.8
+lsdb
```
troyraen: Looks like we need to keep `cdshealpix<0.8` if we're not going to pin lsdb.

troyraen (approved):
Thanks @jaladh-singhal! Nice to have this taken care of. I tested on Fornax large server and got similar timing results to those you reported.

I see that @bsipocz is working around dependencies for our environments but I think we still need to keep a couple of pins for end users who create their own environments based on these requirements files.

light_curves/requirements_light_curve_collector.txt
jaladh-singhal: Yes let's hope @bsipocz figures out a way to keep in `lsdb>=0.8.1` (my original change) - that way we won't need cdshealpix pin and it will also cover the ConeSearch lower bound requirement too.

bsipocz:
My workaround is ugly, but works. Let me know if you are OK with it.

bsipocz:
Failure is reported already in https://github.com/nasa-fornax/fornax-demo-notebooks/issues/666, and should not block merging here.

jaladh-singhal:
> My workaround is ugly, but works. Let me know if you are OK with it.

@bsipocz I'm ok with it! I see you have reverted the lsdb pin to >=0.8.1 so that looks right.

Can you approve it so that we can merge?

bsipocz (approved):
Approving Jaladh's commits and he did it with mine (but can't do that formally as GH doesn't let PR authors to do so).

PR #678, opened 2026-05-27
Handle light-curve-collector get sample failures

### LINKED ISSUE #628: BUG: AttributeError in scale_up notebook
I have not looked into the version details, or in fact any details beyond this very minimal traceback, but from what I see this is a weird one.

```
light_curves/scale_up.md
  - An exception occurred during code execution, halting further execution:




AttributeError                            Traceback (most recent call last)
Cell In[16], line 4
      1 # Skip archive calls that use internal parallelization.
      2 _archive_names = [name for name in archive_names if name.lower() not in ["panstarrs", "ztf"]]
Site has 1 error and 0 warning, stopping build.
----> 4 sample_table = helpers.scale_up.run(build="sample", **kwargs_dict)
      5 # sample_table is returned if you want to look at it but it is not used below
      7 with multiprocessing.Pool(processes=len(_archive_names)) as pool:
      8     # submit one job per archive

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/light_curves/code_src/helpers/scale_up.py:103, in run(build, **kwargs_dict)
    100 my_kwargs_dict = _construct_kwargs_dict(**kwargs_dict)
    102 if build == "sample":
--> 103     return _build_sample(**my_kwargs_dict)
    105 if build == "lightcurves":
    106     return _build_lightcurves(**my_kwargs_dict)

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/light_curves/code_src/helpers/scale_up.py:168, in _build_sample(get_samples, consolidate_nearby_objects, sample_file, overwrite_existing_sample, **_)
    165     get_sample_fnc(coords, labels, **kwargs)
    167 # create an astropy Table of objects
--> 168 sample_table = sample_selection.clean_sample(
    169     coords, labels, consolidate_nearby_objects=consolidate_nearby_objects
    170 )
    172 # save and return the Table
    173 sample_table.write(sample_file, format="ascii.ecsv", overwrite=True)

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/light_curves/code_src/sample_selection.py:692, in clean_sample(coords_list, labels_list, consolidate_nearby_objects, verbose)
    688     return sample_table['objectid', 'coord', 'label']
    690 # now join the table with itself within a defined radius.
    691 # We keep one set of original column names to avoid later need for renaming
--> 692 tjoin = join(sample_table, sample_table, keys='coord',
    693              join_funcs={'coord': join_skycoord(0.005 * u.deg)},
    694              uniq_col_name='{col_name}{table_name}', table_names=['', '_2'])
    696 # this join will return 4 entries for each redundant coordinate:
    697 # 1 for the match with itself and 1 for the match with the similar
    698 # enough coord target then the same thing again for the match, but all of
    699 # these 4 will have the same id in the new 'coords_id' column
    700 
    701 # keep only those entries in the resulting table which are unique
    702 uniqued_table = unique(tjoin, keys='coord_id')['coord_id', 'coord', 'label']

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/.tox/py312-buildhtml/lib/python3.12/site-packages/astropy/table/operations.py:443, in join(left, right, keys, join_type, keys_left, keys_right, keep_order, uniq_col_name, table_names, metadata_conflicts, join_funcs)
    440 # In case keep_order=True we need try/finally to ensure that the temporary column
    441 # is removed even if an exception is raised.
    442 try:
--> 443     out = _join(
    444         left,
    445         right,
    446         keys,
    447         join_type,
    448         uniq_col_name,
    449         table_names,
    450         metadata_conflicts,
    451         join_funcs,
    452         keys_left=keys_left,
    453         keys_right=keys_right,
    454     )
    455     if sort_table is not None:
    456         # Sort joined table to the original order and remove the temporary column.
    457         out.sort(sort_table_index_key)

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/.tox/py312-buildhtml/lib/python3.12/site-packages/astropy/table/operations.py:1239, in _join(left, right, keys, join_type, uniq_col_name, table_names, metadata_conflicts, join_funcs, keys_left, keys_right)
   1234     if not all(key in keys for key in join_funcs):
   1235         raise ValueError(
   1236             f"join_funcs keys {join_funcs.keys()} must be a "
   1237             f"subset of join keys {keys}"
   1238         )
-> 1239     left, right, keys = _apply_join_funcs(left, right, keys, join_funcs)
   1241 len_left, len_right = len(left), len(right)
   1243 if len_left == 0 or len_right == 0:

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/.tox/py312-buildhtml/lib/python3.12/site-packages/astropy/table/operations.py:1116, in _apply_join_funcs(left, right, keys, join_funcs)
   1114 right = right.copy(copy_data=False)
   1115 for key, join_func in join_funcs.items():
-> 1116     ids1, ids2 = join_func(left[key], right[key])
   1117     # Define a unique id_key name, and keep adding underscores until we have
   1118     # a name not yet present.
   1119     id_key = key + "_id"

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/.tox/py312-buildhtml/lib/python3.12/site-packages/astropy/table/operations.py:192, in join_skycoord.<locals>.join_func(sc1, sc2)
    190 def join_func(sc1, sc2):
    191     # Call the appropriate SkyCoord method to find pairs within distance
--> 192     idxs1, idxs2, d2d, d3d = distance_func(sc1, sc2, distance)
    194     # Now convert that into unique identifiers for each near-pair. This is
    195     # taken to be transitive, so that if points 1 and 2 are "near" and points
    196     # 1 and 3 are "near", then 1, 2, and 3 are all given the same identifier.
    197     # This identifier will then be used in the table join matching.
    198 
    199     # Identifiers for each column, initialized to all zero.
    200     ids1 = np.zeros(len(sc1), dtype=int)

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/.tox/py312-buildhtml/lib/python3.12/site-packages/astropy/coordinates/matching.py:382, in search_around_sky(coords1, coords2, seplimit, storekdtree)
    377     raise ValueError(msg)
    379 # we convert coord1 to match coord2's frame.  We do it this way
    380 # so that if the conversion does happen, the KD tree of coord2 at least gets
    381 # saved. (by convention, coord2 is the "catalog" if that makes sense)
--> 382 coords1 = coords1.transform_to(coords2)
    384 # strip out distance info
    385 urepr1 = coords1.data.represent_as(UnitSphericalRepresentation)

AttributeError: 'Column' object has no attribute 'transform_to'

py312-buildhtml: exit 1 (489.95 seconds) /home/runner/work/fornax-demo-notebooks/fornax-demo-notebooks> bash -c 'jupyter-book build --execute --html --strict -d' pid=2537
  py312-buildhtml: FAIL code 1 (575.95=setup[85.47]+cmd[0.36,0.00,0.00,0.02,0.13,0.01,489.95] seconds)
  evaluation failed :( (575.97 seconds)

```

troyraen:
- Make `clean_sample()` error immediately with an informative message if `coords_list` is empty. Closes #666.
- Make `Ned_query_refcode_exceptions()` retry failed queries. Closes #628.

### DIFF
--- a/light_curves/code_src/sample_selection.py
+++ b/light_curves/code_src/sample_selection.py
@@ -1,3 +1,4 @@
+from time import sleep
 import astropy.units as u
 from alerce.core import Alerce
 from astropy.coordinates import SkyCoord
@@ -527,7 +528,7 @@ def get_sdss_sample(coords, labels, *, num=10, zmin=0, zmax=10, randomize_z=Fals
         print('SDSS Quasar: ' + str(num))
 
 
-def Ned_query_refcode_exceptions(paper_link):
+def Ned_query_refcode_exceptions(paper_link, *, max_retries=2, retry_idx=0):
     """
     Call Ned.query_refcode() with basic network exception handling.
 
@@ -535,6 +536,10 @@ def Ned_query_refcode_exceptions(paper_link):
     ----------
     paper_link : str
         ADS/NED reference code or paper identifier.
+    max_retries: int
+        Maximum number of times to retry the call.
+    retry_idx: int
+        Number of times the call has already been retried.
 
     Returns
     -------
@@ -544,6 +549,11 @@ def Ned_query_refcode_exceptions(paper_link):
     try:
         return Ned.query_refcode(paper_link)
     except (ConnectionError, RequestException, TimeoutError) as e:
+        if retry_idx < max_retries:
+            sleep(10)
+            return Ned_query_refcode_exceptions(paper_link, max_retries=max_retries, retry_idx=retry_idx + 1)
+
+        # We've hit the maximum number of retries.
         warnings.warn(
             f"Encountered {type(e).__name__} for paper {paper_link}. skipping.",
             category=RuntimeWarning,
--- a/light_curves/code_src/sample_selection.py
+++ b/light_curves/code_src/sample_selection.py
@@ -528,7 +528,7 @@ def get_sdss_sample(coords, labels, *, num=10, zmin=0, zmax=10, randomize_z=Fals
         print('SDSS Quasar: ' + str(num))
 
 
-def Ned_query_refcode_exceptions(paper_link, *, max_retries=2, retry_idx=0):
+def Ned_query_refcode_exceptions(paper_link, *, max_retries=2):
     """
     Call Ned.query_refcode() with basic network exception handling.
 
@@ -538,28 +538,26 @@ def Ned_query_refcode_exceptions(paper_link, *, max_retries=2, retry_idx=0):
         ADS/NED reference code or paper identifier.
     max_retries: int
         Maximum number of times to retry the call.
-    retry_idx: int
-        Number of times the call has already been retried.
 
     Returns
     -------
     result : object or None
         Result from Ned.query_refcode, or None if a network error occurs.
     """
-    try:
-        return Ned.query_refcode(paper_link)
-    except (ConnectionError, RequestException, TimeoutError) as e:
-        if retry_idx < max_retries:
-            sleep(10)
-            return Ned_query_refcode_exceptions(paper_link, max_retries=max_retries, retry_idx=retry_idx + 1)
-
-        # We've hit the maximum number of retries.
-        warnings.warn(
-            f"Encountered {type(e).__name__} for paper {paper_link}. skipping.",
-            category=RuntimeWarning,
-            stacklevel=2,
-        )
-        return
+    for attempt in range(max_retries + 1):
+        try:
+            return Ned.query_refcode(paper_link)
+        except (ConnectionError, RequestException, TimeoutError) as e:
+            if attempt < max_retries:
+                # Sleep for a bit, then try again.
+                sleep(10)
+            else:
+                # We've hit the maximum number of retries.
+                warnings.warn(
+                    f"Encountered {type(e).__name__} for paper {paper_link}. skipping.",
+                    category=RuntimeWarning,
+                    stacklevel=2,
+                )
 
 
 def get_paper_sample(coords, labels, *, paper_link="2019A&A...627A..33D", label="Cicco19", verbose=1):
--- a/light_curves/code_src/sample_selection.py
+++ b/light_curves/code_src/sample_selection.py
@@ -677,6 +677,8 @@ def clean_sample(coords_list, labels_list, *, consolidate_nearby_objects=True, v
     The returned table has one row per retained object after optional
         consolidation.
     """
+    if not len(coords_list) > 0:
+        raise ValueError("'coords_list' is empty. Cannot clean sample.")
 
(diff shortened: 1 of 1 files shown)


light_curves/code_src/sample_selection.py
```diff
@@ -544,6 +549,11 @@ def Ned_query_refcode_exceptions(paper_link):
     try:
         return Ned.query_refcode(paper_link)
     except (ConnectionError, RequestException, TimeoutError) as e:
+        if retry_idx < max_retries:
+            sleep(10)
+            return Ned_query_refcode_exceptions(paper_link, max_retries=max_retries, retry_idx=retry_idx + 1)
```
bsipocz: This is a very weird solution, could we instead do a loop for the try except rather than do a full recursion on the function and then don't even expose the index as a parameter?

light_curves/code_src/sample_selection.py
```diff
@@ -677,6 +687,8 @@ def clean_sample(coords_list, labels_list, *, consolidate_nearby_objects=True, v
     The returned table has one row per retained object after optional
         consolidation.
     """
+    if not len(coords_list) > 0:
+        raise ValueError("'coords_list' is empty. Cannot clean sample.")
```
bsipocz: Do you want to propagate the ValueError here, or instead return an empty Table for `sample_table` with all the correct columns? That may work more gracefully later in the notebook code.

bsipocz (commented):
2 comments up for discussion.

light_curves/code_src/sample_selection.py
troyraen: Later cells in the notebook will fail. There's also no point in running the notebook with a sample size of 0. 

We could make `clean_sample()` return an empty table for folks who want to lift the code out and alter for their own use case if you'd like. But then I think we should update `validate_sample_table()` to raise an informative error if the table has 0 rows.

light_curves/code_src/sample_selection.py
troyraen: done

light_curves/code_src/sample_selection.py
bsipocz: OK, I think this can go in as is then. I would still think cleaning a zero input is still cleaning it and it should be failed in the validation step, but it's bike shedding and don't really matter that much to hold up the PR.

bsipocz (approved):

light_curves/code_src/sample_selection.py
troyraen: We can do that, it's not hard. I think it's reasonable to say "can't clean something that doesn't exist", but you're right that the validation step is a better place to fail.

Issue #700, opened 2026-07-15
Gaia call failing with HTTP 500


troyraen:
Gaia call is failing in light-curve-collector (and scale-up). I suspect we need to look into registering ("Registered users have an increased quota, please log in.").

Error message from our `gaia_get_lightcurves()` isn't very informative: `HTTPError: Error 500: null`.

<details><summary>HTTPError traceback</summary>
<p>

```python
500 Error 500:
null
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
Cell In[6], line 4
      1 gaiastarttime = time.time()
      2 
      3 # get Gaia light curves
----> 4 df_lc_gaia = gaia_get_lightcurves(sample_table, search_radius=1/3600, verbose=0)
      5 
      6 # add the resulting dataframe to all other archives
      7 df_lc.append(df_lc_gaia)

File [/opt/notebooks/fornax-demo-notebooks/light_curves/code_src/gaia_functions.py:58](https://console.fornax.sciencecloud.nasa.gov/opt/notebooks/fornax-demo-notebooks/light_curves/code_src/gaia_functions.py#line=57), in gaia_get_lightcurves(sample_table, search_radius, verbose)
     11 '''
     12 Creates a lightcurve Pandas MultiIndex object from Gaia data for a list of coordinates.
     13 This is the MAIN function.
   (...)     49 
     50 '''
     51 # This code is broken into two steps.  The first step, `gaia_retrieve_catalog` retrieves the
     52 # Gaia source ids for the positions of our sample. These come from the "Gaia DR3 source lite catalog".
     53 # However, that catalog only has a single photometry point per object.  To get the light curve
   (...)     56 
     57 # Retrieve Gaia table with Source IDs ==============
---> 58 gaia_table = gaia_retrieve_catalog(sample_table,
     59                                    search_radius=search_radius,
     60                                    verbose=verbose
     61                                    )
     62 # if none of the objects were found, there's nothing to load and the gaia_retrieve_epoch_photometry fnc
     63 # will raise an HTTPError. just return an empty dataframe instead of proceeding
     64 if len(gaia_table) == 0:

[...]

Issue #697, opened 2026-07-15
IceCube data appears to have moved


troyraen:
The IceCube call in light-curve-collector (and scale-up) is failing with a `URLError` (full traceback below). It's failing on [this line](https://github.com/nasa-fornax/fornax-demo-notebooks/blob/7fc2b639b0d8c04fcdb3284be3cef0cfc8813b3e/light_curves/code_src/icecube_functions.py#L212) trying to retrieve `url="http://icecube.wisc.edu/data-releases/20210126_PS-IC40-IC86_VII.zip"`. That url does not appear to exist. The function docstring points to https://icecube.wisc.edu/data-releases/2021/01/all-sky-point-source-icecube-data-years-2008-2018/, which points to https://dataverse.harvard.edu/dataset.xhtml?persistentId=doi:10.7910/DVN/VKL316 for the data.

Go to that last url, click "Access Dataset", and then "Original Format ZIP (390.2 MB)" -- I'm pretty sure that's what we need (modulo version 1.0 vs 2.0). The problem is, I can't figure out how to get that programmatically. Right-clicking and selecting "Copy Link Address" just gives me a link to the main page. Starting the download and grabbing the url from there returns "404 No such download job!" when I try to use it independently.

<details><summary>URLError traceback</summary>
<p>

```python
SSLCertVerificationError                  Traceback (most recent call last)
File [/opt/envs/base/lib/python3.12/urllib/request.py:1344](https://console.fornax.sciencecloud.nasa.gov/opt/envs/base/lib/python3.12/urllib/request.py#line=1343), in AbstractHTTPHandler.do_open(self, http_class, req, **http_conn_args)
   1343 try:
-> 1344     h.request(req.get_method(), req.selector, req.data, headers,
   1345               encode_chunked=req.has_header('Transfer-encoding'))
   1346 except OSError as err: # timeout error

File [/opt/envs/base/lib/python3.12/http/client.py:1358](https://console.fornax.sciencecloud.nasa.gov/opt/envs/base/lib/python3.12/http/client.py#line=1357), in HTTPConnection.request(self, method, url, body, headers, encode_chunked)
   1357 """Send a complete request to the server."""
-> 1358 self._send_request(method, url, body, headers, encode_chunked)

[...]

Issue #696, opened 2026-07-14
scale_up notebook failing consistently in the images CI


zoghbi-a:
https://github.com/nasa-fornax/fornax-images/actions/runs/29350279984/job/87144477754

zoghbi-a:
Also fails with the same error when tested manually.

troyraen:
I ran this twice on Fornax. First one ran without error, second one produced the error you report above. I suspect `multiprocessing` may be swallowing some error messages, so I ran it without `multiprocessing` and immediately got an HTTP 500 from Gaia. I'll investigate more, but I wonder if the 500 causes `multiprocessing` to exit early. If so, `df_lc` would only contain the results from the other archives that were fully processed before Gaia errored, so whether it triggers the error reported above (ie, whether <10 results total) just depends on how fast Gaia errors.

<details><summary>Gaia HTTP 500 traceback</summary>
<p>

```python
2026/07/14 20:51:19 UTC | [pid=868] Starting build=sample
Building object sample from: ['yang']
Changing Look AGN- Yang et al:  31
Object sample size, after duplicates removal: 30
Object sample saved to: /opt/notebooks/fornax-demo-notebooks/light_curves/output/lightcurves-demo-Yang-sample/object_sample.ecsv
2026/07/14 20:51:19 UTC
2026/07/14 20:51:19 UTC | [pid=868] Starting build=lightcurves, archive=gaia
500 Error 500:
null
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
Cell In[20], line 9
      5 # sample_table is returned if you want to look at it but it is not used below
      6 
      7 for archive in _archive_names:
      8     print(archive)
----> 9     helpers.scale_up.run(build="lightcurves", archive=archive, **kwargs_dict)

File [/opt/notebooks/fornax-demo-notebooks/light_curves/code_src/helpers/scale_up.py:106](https://console.fornax.sciencecloud.nasa.gov/opt/notebooks/fornax-demo-notebooks/light_curves/code_src/helpers/scale_up.py#line=105), in run(build, **kwargs_dict)
    103     return _build_sample(**my_kwargs_dict)
    105 if build == "lightcurves":
--> 106     return _build_lightcurves(**my_kwargs_dict)
    108 return _build_other(keyword=build, **my_kwargs_dict)

[...]

PR #705, opened 2026-07-15
Fix WISE band names in light-curve-collector

### LINKED ISSUE #704: WISE band names were only partially changed in light-curve-collector
#574 changed WISE band names in the code of fluxconversions.py and wise_functions.py and in the serial section of light_curve_collector.md, but didn't change them in docstrings of the code files or in the parallel section of light_curve_collector.md. This doesn't cause tests to fail because the parallel call doesn't raise an error in the main process, but WISE data are missing from the results and the plots.

troyraen:
Resolves #704

I expect Fornax image tests to fail due to #697 and/or #700.

### DIFF
--- a/light_curves/code_src/fluxconversions.py
+++ b/light_curves/code_src/fluxconversions.py
@@ -14,7 +14,7 @@ def convert_wise_flux_to_millijansky(flux_nanomaggy, *, band=None):
     flux_nanomaggy : float or iterable of floats
         Flux in nanomaggy.
     band : str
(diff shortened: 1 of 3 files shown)


troyraen:
Failures are #697 and #700, plus a HEASARC connection error which I'm guessing is transient because it worked for me on Fornax a few minutes after this test failed. (If it's actually not transient, we'll catch that with #696.)

troyraen:
@jkrick will you review this?

jkrick (approved):
I can confirm that the parallel section completes through wise

Issue #710, opened 2026-07-20
PanSTARRS raising ValueError in light-curve-collector


troyraen:
I know the PanSTARRS data was updated recently ([fornax announcement](https://discourse.fornax.sciencecloud.nasa.gov/t/new-fornax-features-hats-catalogs-and-notebooks/1083#p-3722-new-hats-catalogs-2)), so I'm guessing something about that is what's causing this.

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[12], line 5
      1 panstarrsstarttime = time.time()
      2 
      3 panstarrs_search_radius = 1.0 # search radius = 1 arcsec
      4 # get panstarrs light curves
----> 5 df_lc_panstarrs = panstarrs_get_lightcurves(sample_table, radius=panstarrs_search_radius)
      6 
      7 # add the resulting dataframe to all other archives
      8 df_lc.append(df_lc_panstarrs)

File [~/fornax/fornax-demo-notebooks/light_curves/code_src/panstarrs_functions.py:89](https://console.fornax.sciencecloud.nasa.gov/home/raen/fornax/fornax-demo-notebooks/light_curves/code_src/panstarrs_functions.py#line=88), in panstarrs_get_lightcurves(sample_table, radius)
     78 panstarrs_detect = lsdb.open_catalog(
     79     's3://stpubdata/panstarrs/ps1/public/hats/detection',
     80     search_filter=lsdb.PixelSearch(matched_objects.get_healpix_pixels()),
   (...)     85              ]
     86 )
     88 # plan to join that cross match with detections to get light-curves
---> 89 matched_lc = matched_objects.join(
     90     panstarrs_detect,
     91     left_on="objID",
     92     right_on="objID",
     93     output_catalog_name="yang_ps_lc",
     94     suffixes=["", ""],
     95     suffix_method="overlapping_columns",
     96     log_changes=False,
     97 )
     99 # Create default local cluster
    100 # here is where the actual work gets done
    101 # Use multiple workers with a single thread per worker for better performance on Fornax
    102 with Client(threads_per_worker=1, memory_limit=None):
    103     # compute the cross match with object table
    104     # and the join with the detections table

[...]

Issue #663, opened 2026-03-12
BUG: make CI pick up notebooks that has associated code change


bsipocz:
E.g. #660 hasn't been triggered the CI as the PR didn't change the notebook itself but the associated source code. This is a fornax unique issue, we don't face similar issues with our regex for the IRSA notebooks

bsipocz:
We should also make sure that it's not just a one-on-one matching for notebooks, but it could be multiple notebooks dependent on code snippets (e.g. the scale-up notebook heavily reuses the lightcurve materials)

see e.g. https://github.com/nasa-fornax/fornax-demo-notebooks/issues/704, even though we wouldn't have seen issues as we would not have picked up on the missing band without having a raised error

troyraen:
I don't think scale-up is a special case here, it calls source code the same way all the other notebooks do. We should check what each notebook actually imports because we have 4 `code_src` directories and 7 notebooks, so notebooks don't necessarily use every file in their `code_src` directory. (Maybe we're actually saying the same thing here? I'm not sure what you meant.)

PR #714, opened 2026-07-25
Switch to TAP query for IceCube data

### LINKED ISSUE #697: IceCube data appears to have moved

### LINKED ISSUE #717: Merge icecube_functions.py with heasarc_functions.py
#714 switches the icecube_functions.py call to query HEASARC TAP via astroquery. It's very similar to heasarc_functions.py and the two should be merged together. Things to watch out for:

- [ ] I'm not sure whether IceCube should just be added as another catalog for `heasarc_get_lightcurves()` to get or if it's different enough that it needs to remain a standalone function. This will impact both heasarc_functions.py and light_curve_collector.md.
- [ ] heasarc_functions implements chunking for large sample sizes. icecube_functions may need that too (pending #711).
- [ ] The WHERE clauses of the two queries are different. heasarc: `WHERE cat.error_radius < {max_error_radius}`; icecube: `WHERE 1=CONTAINS(POINT('ICRS', samp.ra, samp.dec), CIRCLE('ICRS', cat.ra, cat.dec, {max_search_radius}))`. ~~I haven't looked in detail but am guessing the heasarc one should be switched to use the icecube one.~~ Neither is wrong, they're just different (see my comment below)
- [ ] icecube_functions leaves the event energy in log10(GeV) while heasarc_functions just makes up a "flux". Might be better to make these handle that the same way since they're trying to solve the same problem (these events don't have fluxes but we want to plot them with the other light curves).

troyraen:
Closes #697 

Switches `icecube_get_lightcurves()` to use a TAP query via `astroquery.heasarc.Heasarc` instead of downloading a zip file.

I have the zip file, so was able to compare results from both methods. I verified that they return the same dataframe for the default sample, except for 1 row for `objectid` = 18 that has the same flux but a different time. This happens because:

- For each object, we sort by flux (descending) and then return only the top 3.
- The 3rd largest flux for `objectid` = 18 is 3.61, but there are three of those. So a different sorting w.r.t. time will return a different row.

```
flux | err | time         | objectid | label   | band
3.98 | 0.0 | 54663.120914 | 18       | Yang 18 | IceCube
3.87 | 0.0 | 57641.869582 | 18       | Yang 18 | IceCube
3.61 | 0.0 | 56782.741748 | 18       | Yang 18 | IceCube    <--- Returned by this branch
3.61 | 0.0 | 56669.239576 | 18       | Yang 18 | IceCube    <--- Returned by main branch
3.61 | 0.0 | 57386.723595 | 18       | Yang 18 | IceCube
```

So I think this is fine.

I expect testing to fail due to #700 and/or #710.


troyraen:
The Fornax image test didn't execute the notebook and I don't know why. I know light-curve-collector is skipped in the CircleCI job but I think it should be executed in the Fornax image test.

Update: It's because I didn't touch the notebook, just a code file, and #663 is still open.

light_curves/code_src/icecube_functions.py
```diff
@@ -116,111 +73,53 @@ def icecube_get_lightcurves(sample_table, *, icecube_select_topN=3, max_search_r
...
+    icecube_catalog = "icecubepsc"
+    query = f"""
+        SELECT samp.objectid, samp.label, cat.time, cat.event_energy
+        FROM {icecube_catalog} cat, tap_upload.sample_table samp
+        WHERE 1=CONTAINS(POINT('ICRS', samp.ra, samp.dec), CIRCLE('ICRS', cat.ra, cat.dec, {max_search_radius}))
+        AND DISTANCE(POINT('ICRS', samp.ra, samp.dec), POINT('ICRS', cat.ra, cat.dec)) < cat.error_radius
+    """
+
+    # Query HEASARC for IceCube events within our sample target areas.
+    # This will prepend the catalog name ("samp" or "cat") to the column names, eg, "time" -> "cat_time".
```
jkrick: the existing heasarc function does this with `select cat.time AS time`   Is this worth doing for consistency?

jkrick (changes requested):
I am getting a `TypeError: HeasarcClass.query_tap() got an unexpected keyword argument 'uploads'` and the cell fails.  I am running on the py-light_curve_collector kernel on Fornax.

Once working, please update the runtime in the main notebook, in case this decreased the runtime.  Also, I couldn't update it after the gaia change, so now is a good time to get a new estimate.  

While we are in this function, there are a few consistency issues with the "other" heasarc_function.  1) it uses chunking, to avoid problems when scaling. 2. It also does a 'select cat.time AS time` (mentioned in my comment below) instead of ice cube's version which just has the `cat` prepended. 

What do you think about putting this ice cube function inside heasarc_functions.py?  We don't do that for the other archives, ie., there is no irsa_functions.py, but both the existing heasarc_function and the ice_cube function do hit heasarc.  Or we rename the existing heasarc_functions to be something more specific to what it does, and not be named for the archive.  Doesn't have to be this PR, just a thought.

light_curves/code_src/icecube_functions.py
```diff
@@ -116,111 +73,53 @@ def icecube_get_lightcurves(sample_table, *, icecube_select_topN=3, max_search_r
     return (MultiIndexDFObject(data=filter_icecube_df))
 
 
-def icecube_get_catalog(path=DATA_PATH, verbose=False):
+def icecube_query_catalog(sample_table, *, max_search_radius):
```
bsipocz: this is fine here as we only use the functions in notebooks, but overall I think at some point we should converge on function name and threat them as public API -- only change them with deprecations.

For the timeline, I would say once we have them rendered in the documentation as API; which we don't yet do, all of these are practically hidden from the HTML pages.

bsipocz (commented):
I would need to fix up #663 to get an actual CI status on this. As for the code, it looks OK, my only comment is for the long term future and to start to think about the functions as public API.

troyraen:
> I am getting a TypeError: HeasarcClass.query_tap() got an unexpected keyword argument 'uploads' and the cell fails.

Ah, my update to the requirements file didn't make it in here yet...

light_curves/code_src/icecube_functions.py
troyraen: I don't think it matters much one way or the other. The column names get standardized before being returned to the notebook.

light_curves/code_src/icecube_functions.py
troyraen: I think of the functions the notebook calls as public and all others (like this one) as private. I think we talked at some point about prepending `_` to private functions but never implemented that as our standard practice.

light_curves/requirements_light_curve_collector.txt
```diff
@@ -10,7 +10,10 @@ matplotlib
 hpgeom
 astropy
 pyvo
-astroquery>=0.4.8
+# astroquery>=0.4.12 needed for table upload to Heasarc.query_tap.
+# .dev needed until v0.4.12 is released.
+# When .dev is removed, also remove --pre from pip-install cell of notebook.
+astroquery>=0.4.12.dev
```
troyraen, line 16: @bsipocz @jkrick, this is what's required for the updates as written. Do you think that's a problem? I'm hoping no but guessing yes. If yes, I suppose I should merge this into heasarc_functions.py now.

jkrick: I will defer to @bsipocz on that one.

bsipocz: no, it should not be a problem, this would pick up a compatible release from pypi (and we had this approach previously for other versions, too)

light_curves/code_src/icecube_functions.py
bsipocz: ok, that's fine then I think. We may want to do the imports a little bit differently to hint on this difference, but that is also goes beyond scope for this PR

troyraen:
Ok, this is ready to review again @jkrick. You'll need to run the pip-install cell on Fornax to update the environment. Once we get all these bugs fixed we should open an issue in fornax-images asking them to pull in the notebook/code updates and also update that kernel.

> What do you think about putting this ice cube function inside heasarc_functions.py? 

Too bad I didn't realize that's doing almost exactly the same thing. I do think we should combine them but I'll open a separate issue. Shouldn't be too hard but there are 2-3 things to look at and I don't have time right now.

troyraen:
Fornax image tests is failing because it doesn't pick up the environment update. That is expected since the point is to test using the current Fornax kernel. It would be nice to have a test that uses the current fornax kernel plus env updates made in the PR, but that's non-trivial and low priority atm.

bsipocz:
>  It would be nice to have a test that uses the current fornax kernel plus env updates made in the PR, but that's non-trivial and low priority atm.

Once moving the rendering job off of circleCI we can easily add back this notebook to be executed, that should solve this problem.

jkrick:
The parallel cell doesn't complete the ice cube search for me.  It appears as though the problem is that the IceCube TAP query hits the same HEASARC Xamin service as heasarc_get_lightcurves, and when the two run concurrently in the multiprocessing pool Xamin returns an empty/errored response ~50% of the time.  icecube_get_lightcurves  has no error_callback so it silently discards the failure, leaving no IceCube rows in parallel_df_lc.  Claude suggests moving one of the two functions out of the parallel call and adding an error_callback.

troyraen:
@jkrick can you elaborate on what you saw and how many times you tried it? I can't tell from your last comment whether you got an error response or an empty dataframe or where the 50% number came from. I ran the notebook end-to-end yesterday without error and there was icecube data in the figures. Trying again now but I haven't been able to load the sample because NED is timing out.

jkrick:
The symptom is lack of ice cube data in the final plot when using parallel_df_lc to make that plot.  I ran it intentionally 10 times to figure out what was going on, and on about 50% of those runs, the prallel_df_lc didn't have ice cube data in it.  The code doesn't give you an error response, but my suggestion is to change the code so that it will give an error when something like this happens, in addition to moving either ice cube or heasarc_get_lighcurves out of the parallel loop to avoid the problem.

troyraen:
Thanks.
I think I'd rather just do #717 instead of taking it out of the pool. The more calls we take out of there the less useful that whole parallel section becomes.

jkrick:
> I think I'd rather just do https://github.com/nasa-fornax/fornax-demo-notebooks/issues/717 instead of taking it out of the pool

Unless you are going to merge the functions today,  please take it out of the pool in this PR so that the bug gets fixed, while you figure out #717.  I don't think #717 will be so easy.  Early on I tried making 'uber' functions for the archives, and it was not so easy.  I think it is important to get the notebook functioning again.

troyraen:
I definitely agree that it's important to get the notebook functioning again.

troyraen:
I merged `icecube_get_lightcurves()` into `heasarc_get_lightcurves()`. This makes the function more complicated because it has to handle different kinds of arguments (see https://github.com/nasa-fornax/fornax-demo-notebooks/issues/717#issuecomment-5125729757). I tried to make the docstring really clear about how to call it. In order to be able to call these serially in parallel processing, I don't see any other option but to have one function that accepts arguments for all three HEASARC catalogs. I verified that this returns the same data as the main branch for all three catalogs, except the one row mentioned above. Let me know what you think. This would close #717. 

I also wrapped the `lightkurve` call in `tess_kepler_get_lightcurves()` in a context manager to silence this warning that gets repeated for every object in `sample_table`:

```
WARNING: AstropyDeprecationWarning: "objectname" was deprecated in version 0.4.12 and will be removed in a future version. Use argument "object_name" instead. [astroquery.utils.class_or_instance]
```

bsipocz:
> I also wrapped the lightkurve call in tess_kepler_get_lightcurves() in a context manager to silence this warning that gets repeated for every object in sample_table:

Thanks! Ideally lightkurve will fix that, but I don't see a a timeline for it so indeed it's better if we also wrap it in our code for now.

light_curves/scale_up.md
```diff
@@ -310,7 +311,7 @@ Specifically:
 
 There are many interesting features in the figure that the reader may want to look at in more detail.
 For example, other than ZTF's memory spike at the end, we see that the full run collecting multi-wavelength light curves for the SDSS 500k sample could be completed with about 2 CPU and 60G RAM.
-We also see that the IceCube call completes very quickly but requires a significant amount of memory.
+We also see that the IceCube call completed very quickly but required a significant amount of memory, which was the code behavior prior to an update in August 2026.
```
jkrick, line 319: I don't think it is necessary to reference prior behavior.  I would suggest just deleting this line.

light_curves/code_src/heasarc_functions.py
```diff
@@ -175,19 +169,162 @@ def heasarc_get_lightcurves(sample_table, *, catalog_error_radii={"FERMIGTRIG":
...
+        given.
+
+    Returns
+    -------
+    str or pandas.DataFrame
+        The ADQL query string, if `max_error_radius` is given, otherwise the light curve DataFrame
+        processed from `hresulttable`.
+    """
+    if max_error_radius is not None:
+        # SELECT cat.name, cat.ra, cat.dec, cat.error_radius, cat.time AS time,
```
jkrick:
```suggestion

```

light_curves/code_src/heasarc_functions.py
```diff
@@ -175,19 +169,162 @@ def heasarc_get_lightcurves(sample_table, *, catalog_error_radii={"FERMIGTRIG":
...
+    Returns
+    -------
+    str or pandas.DataFrame
+        The ADQL query string, if `max_search_radius` is given, otherwise the light curve DataFrame
+        processed from `hresulttable` and `select_topN`.
+    """
+    if max_search_radius is not None:
+        # ADQL to find IceCube events within max_search_radius of our sample targets.
+        # AND DISTANCE removes events with an error radius less than the distance between the event and our sample target.
+            # SELECT mt.objectid, mt.label, cat.time, cat.event_energy
```
jkrick:
```suggestion

```

light_curves/light_curve_collector.md
jkrick:
```suggestion
The function to retrieve HEASARC data accesses the HEASARC archive using an astroquery TAP query with a table upload.  This is the fastest way to access data from HEASARC catalogs at scale.
```

light_curves/light_curve_collector.md
```diff
@@ -369,7 +368,8 @@ This time series (time vs. neutrino energy) information is similar to photometry
 icecubestarttime = time.time()
 
 # get icecube data points
-df_lc_icecube = icecube_get_lightcurves(sample_table, icecube_select_topN=3)
+heasarc_catalogs = {"icecubepsc": {"select_topN": 3, "max_search_radius": 2.0}}
```
jkrick: a bit confusing because you already assigned heasarc_catalogs above.  If this stays here and gets done separate from section 2.1, I suggest renaming heasarc_catalogs here to icecube_catalog, then also in the line below.

jkrick (approved):
should we move section 3.3 Icecube into section 2.1 heasarc?  Fine to make this an issue to do later.  

Also, I just noticed that the scale_up notebook doesn't have a runtime section.

light_curves/light_curve_collector.md
troyraen: changed

light_curves/scale_up.md
troyraen: I think we should comment on this because the figure shows it but the behavior of the function will be quite different now.

troyraen (commented):
I added a runtime section to scale-up.

I think the IceCube section should stay separate from the other HEASARC section in light-curve-collector because it's not a NASA funded project. I added "HEASARC" to the IceCube section name to match what we're doing elsewhere.

troyraen:
The CircleCI job has some errors I didn't expect. I think they're all transient connection errors but it's a little hard to tell due to multiprocessing. I've run both these notebooks on Fornax a couple times since my last commit and everything went fine. So I'll merge this now and watch the deployment job to make sure it succeeds.

Issue #677, opened 2026-05-27
`sparclclient` fails to create a client


troyraen:
`sparclclient`, used by `DESIBOSS_get_spec()` function in spectra_collector, is failing. We’re using the latest version (1.3.0), which makes me wonder if this is an upstream issue.

MRE:

```python
from sparcl.client import SparclClient
client = SparclClient()
```

This is preventing all deployments. First seen on May 11 in: https://github.com/nasa-fornax/fornax-demo-notebooks/actions/runs/25658837449/job/75313921808

```
Exception                                 Traceback (most recent call last)
Cell In[14], line 2
      1 ## Get DESI and BOSS spectra with SPARCL
----> 2 df_spec_DESIBOSS = DESIBOSS_get_spec(sample_table, search_radius_arcsec=5)
      3 df_spec.append(df_spec_DESIBOSS)

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/spectroscopy/code_src/desi_functions.py:33, in DESIBOSS_get_spec(sample_table, search_radius_arcsec)
     14 """
     15 Retrieve DESI and BOSS spectra for a list of sources.
     16 Note, that we can also retrieve SDSS-DR16 spectra here, which
   (...)     29     The spectra returned from the archive.
     30 """
     32 # Set up client
---> 33 client = SparclClient()
     35 # Initialize multi-index object:
     36 df_spec = MultiIndexDFObject()

File ~/work/fornax-demo-notebooks/fornax-demo-notebooks/.tox/py312-buildhtml/lib/python3.12/site-packages/sparcl/client.py:246, in SparclClient.__init__(self, url, verbose, show_curl, connect_timeout, read_timeout, announcement)
    235 if (int(self.apiversion) - int(expected_api)) >= 1:
    236     msg = (
    237         f"The SPARCL Client you are running expects an older "
    238         f"version of the API services. "
   (...)    244         f"at {self.apiurl}."
    245     )
--> 246     raise Exception(msg)
    248 self.clientversion = client_version
    249 self.fields = Fields(self.apiurl)

Exception: The SPARCL Client you are running expects an older version of the API services. Please upgrade to the latest "sparclclient" using  "pip install --upgrade sparclclient". The Client you are using expected version 13.0 but got 14.0 from the SPARCL Server at https://astrosparcl.datalab.noirlab.edu/api.
```

troyraen:
[...]

PR #725, opened 2026-08-07
Bump to sparclclient>=1.3.0

troyraen:
#677 is back in https://github.com/nasa-fornax/fornax-images/issues/198. I now strongly suspect that the resolution is to pin sparcclient>=1.3.0, so that's what this PR does.

We thought we resolved this in our GHA job by bumping to python=3.13 but that's probably because it pulled in sparcclient=1.3.0 as a side effect rather than being due to the python upgrade itself. See https://github.com/nasa-fornax/fornax-demo-notebooks/issues/677#issuecomment-5210963836 for details.

### DIFF
--- a/spectroscopy/requirements_spectra_collector.txt
+++ b/spectroscopy/requirements_spectra_collector.txt
@@ -4,7 +4,8 @@
 numpy
 matplotlib
 pandas
-sparclclient>=1.2.8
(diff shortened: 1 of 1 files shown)


bsipocz (approved):
The python 3.13 solution worked as it wasn't running into unresolvable pandas requirement, so while this PR is fine, I don't expect this to be enough on its own and we still need to keep the newer python version.

PR #730, opened 2026-08-18
Switch light-curve-collector to use Dask Gateway

### LINKED ISSUE #691: Run multiple lsdb/dask calls in parallel
light-curve-collector parallel section currently uses python's `multiprocessing` to run most of the archive calls in parallel, but then has to run the ones that use lsdb/dask separately and serially. LINCC suggests trying the following pattern to run all of them in parallel:

```python
with Client(...) as client:
    cat1_future = client.compute(catalog1.to_dask_dataframe())[0]
    cat2_future = client.compute(catalog2.to_dask_dataframe())[0]
    df1 = cat1_future.result()
    df2 = cat2_future.result()
```

### LINKED ISSUE #729: Update light-curve-collector to use Dask Gateway
Replace `multiprocessing` with Dask Gateway. This should allow us to run all archive calls in parallel. Should also allow the parallel section to be used for very large samples, making scale-up obsolete.

troyraen:
Closes #729
Closes #691 

Leave this as a draft to avoid accidentally merging before the Dask Gateway feature is released on Fornax.

To Do: 

- [x] Switch parallel section to use dask gateway. (@troyraen)
- [x] Update archive calls that use lsdb/dask so they work both when run locally (in serial section) and when run on dask gateway (in parallel section). (@troyraen)
- [ ] Set dask gateway worker profile and configure worker environments. (@troyraen or @DeekshaMohanty)
- [ ] Test notebook end-to-end with default sample and make sure it works. (@troyraen or @DeekshaMohanty)
- [ ] Test parallel section with full 500k sample and make sure it works.
- [x] Remove scale-up notebook and helper code. (Using Dask Gateway in light-curve-collector makes it obsolete.) (@troyraen)
    - [ ] BLOCKED by decision on #734 
- [ ] Update narrative to explain the usage of dask gateway, remove references to scale-up, and make sure it includes all info users will need to run it with very large samples. Comment that parallel section won't work outside Fornax or for users without permission to use dask gateway. (@jkrick)
    - [ ] BLOCKED by decision on #735 
    - Note that [fornax-howtos#10](https://github.com/nasa-fornax/fornax-howtos/issues/10) will cover how to use dask gateway, so this notebook can point there rather than repeating that info.
    - Note that [fornax-howtos#12](https://github.com/nasa-fornax/fornax-howtos/issues/12) will cover scaling and troubleshooting with dask gateway, but unlikely to be done before this one merges. This notebook shouldn't cover that material in detail, but might need to say _something_ about it until fornax-howtos#12 is done. That should be more clear after we test the 500k sample.


light_curves/light_curve_collector.md
```diff
@@ -62,14 +60,41 @@ By the end of this tutorial, you will be able to:
 
 ## Imports
 
+:::{warning}
+For the parallel section, Python and several Dask-related libraries must be pinned to the exact versions installed on the Dask Gateway cluster.
+Use `requirements_light_curve_collector_dg.txt` to create a new environment and kernel with those versions.
```
DeekshaMohanty: I don’t see `requirements_light_curve_collector_dg.txt` in this branch. Did you maybe mean `requirements_light_curve_collector.txt`, or are we planning to rename the existing requirements file?

light_curves/light_curve_collector.md
```diff
@@ -62,14 +60,41 @@ By the end of this tutorial, you will be able to:
...
+First, you must copy the into a directory where it's the only requirements file because Fornax's `setup-pip-env` expects to find exactly one and this directory has several.
```
DeekshaMohanty: Noticed a small typo - 
```suggestion
First, you must copy the requirements file into a directory where it's the only requirements file because Fornax's `setup-pip-env` expects to find exactly one and this directory has several.
```

light_curves/light_curve_collector.md
```diff
@@ -62,14 +60,41 @@ By the end of this tutorial, you will be able to:
...
+
+```sh
+# Assuming your home directory doesn't already have a requirements file.
+cp requirements_light_curve_collector_dg.txt ~/.
+cd ~/
+# Create the env and kernel.
+# Remove --user to make them temporary so they don't persist across restarts.
+setup-pip-env --user --python=3.13.6
```
DeekshaMohanty: I found that `setup-pip-env` did not discover the underscore-named file: it returned 

`No requirements-*.txt files found`, 

until I copied it as `requirements-light-curve-collector.txt`, so I think it expects a hyphen after requirements? Should we add a separately named requirements file that matches the helper’s `requirements-*.txt` pattern, or update the instructions to copy the existing file under a hyphenated name?

light_curves/light_curve_collector.md
```diff
@@ -432,36 +453,63 @@ rsp_search_radius = 0.001
...
+    client.submit(gaia_get_lightcurves, sample_table, **gaia_kwargs),
+    client.submit(ztf_get_lightcurves, sample_table, radius=ztf_search_radius),
+    client.submit(panstarrs_get_lightcurves, sample_table, radius=panstarrs_search_radius),
+#    client.submit(rubin_get_lightcurves, sample_table, **rsp_kwargs),
+]
+for df_lc in client.gather(futures):
+    parallel_df_lc.append(df_lc)
+
+# shut down the cluster now that all archive calls have completed
+client.close()
```
DeekshaMohanty: Maybe we should move this to a different cell in the notebook, as it is in the Dask Gateway how-to notebook? Since `client.gather(futures)` can raise an exception, the current `client.close()` and `cluster.shutdown()` lines are never reached when that happens, which leaves the cluster running. A separate cleanup cell would give users something they can execute manually after a failure.

